PR #5, opened 2019-01-31
Prepare release script

### LINKED ISSUE #1: Add prepare_release script for CI pipeline
As discussed: We need a `prepare_release` script that runs adapts the `ControllerRegistration` example file (https://github.com/gardener/gardener-extensions/blob/master/controllers/os-coreos/example/controllerregistration.yaml#L13-L16).

See https://github.com/gardener/dashboard/blob/master/.ci/pipeline_definitions#L30-L32 for pipeline configuration.

PS: There is already the `controllers/os-coreos/hack/generate-examples` script - can we re-use it or adapt it somehow?

adracus:
Fixes #1 
Current problem is that the `prepare_release` script is always being run in the default concourse image - This is not feasible for us as this would require us to do the whole bootstrap process ourselves in the script again, leading to us having to maintain yet another 'setup'.

### DIFF
--- a/.ci/prepare_release
+++ b/.ci/prepare_release
@@ -1,0 +1,55 @@
+#!/usr/bin/env sh
+#
+# Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


.ci/prepare_release
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env sh
```
rfranzke: Do we need to add a license header?

VERSION
```diff
@@ -1 +1 @@
-0.1.0-dev
+v0.2.0-dev
```
rfranzke: I think the pipeline can not understand versions with `v`

PR #11, opened 2019-02-25
Implement common command line setup

adracus:
Since we will write more and more controllers, we need some common boundaries to setup command line parsing / value propagation. This PR addresses it by creating `Option`s that expose command line flags which can be consumed in `manager.Options`, `controller.Options` and a `rest.Config`.

**Release note**:

```improvement operator
NONE
```


rfranzke:
Can you check what has been reported by CI/CD?

rfranzke:
Also, the worker count default is now `1` (instead of `5`). Can we keep it with `5`?

PR #28, opened 2019-03-14
GCP infrastructure controller

### LINKED ISSUE #24: Implement `Infrastructure` controller for GCP provider
Similar to how we have implemented the `Infrastructure` extension resource controller for the [AWS provider](https://github.com/gardener/gardener-extensions/tree/master/controllers/provider-aws/pkg/infrastructure) let's please now do it for GCP.

Based on [the current implementation](https://github.com/gardener/gardener/blob/master/example/90-shoot-gcp.yaml#L13-L18) the `InfrastructureConfig` should look like this:

```yaml
apiVersion: gcp.provider.extensions.gardener.cloud/v1alpha1
kind: InfrastructureConfig
networks:
# vpc:
#   name: some-existing-vpc-name
  internal: 10.250.112.0/22
  workers: 10.250.0.0/19
```

Based on the [current implementation](https://github.com/gardener/gardener/blob/master/charts/seed-terraformer/charts/gcp-infra/templates/_main.tf#L121-L135) the `InfrastructureStatus` should look like this:

```yaml
---
apiVersion: gcp.provider.extensions.gardener.cloud/v1alpha1
kind: InfrastructureStatus
networks:
  vpc:
    name: network-name
  subnets:
  - purpose: nodes
    name: my-subnet
  - purpose: internal
    name: other-subnet
serviceAccountEmail: serviceaccount@email
```

The current infrastructure creation/deletion implementation can be found [here](https://github.com/gardener/gardener/blob/master/pkg/operation/cloudbotanist/gcpbotanist/infrastructure.go). Please try to change as little as possible (with every change the risk that we break something increases!) and just move the code over into the extensions infrastructure actuator.

adracus:
Implement the GCP infrastructure controller.

Fixes #24 

**Release note**:

```improvement operator
The new GCP infrastructure controller allows the reconciliation of GCP infrastructures.
```


controllers/provider-gcp/pkg/controller/infrastructure/actuator.go
```diff
@@ -0,0 +1,207 @@
...
+	tf *terraformer.Terraformer,
+	infra *extensionsv1alpha1.Infrastructure,
+	config *gcpv1alpha1.InfrastructureConfig,
+) error {
+	outputKeys := []string{
+		gcp.TerraformerOutputKeyVPCName,
+		gcp.TerraformerOutputKeySubnetNodes,
+		gcp.TerraformerOutputKeyServiceAccountEmail,
+	}
+	hasInternal := config.Networks.Internal != nil
```
rfranzke: newline above to improve readability

pkg/controller/cluster.go
```diff
@@ -98,7 +98,11 @@ func ShootFromCluster(cluster *extensionsv1alpha1.Cluster) (*gardenv1beta1.Shoot
 // ShootIsFailed returns whether the given shoot is marked as 'failed'.
 func ShootIsFailed(shoot *gardenv1beta1.Shoot) bool {
 	lastOperation := shoot.Status.LastOperation
-	return lastOperation != nil && lastOperation.State == gardencorev1alpha1.LastOperationStateFailed && shoot.Generation == shoot.Status.ObservedGeneration
+	if lastOperation == nil {
+		return false
+	}
+
+	return lastOperation.State == gardencorev1alpha1.LastOperationStateFailed
```
rfranzke: Why is `&& shoot.Generation == shoot.Status.ObservedGeneration` removed?

pkg/controller/infrastructure/options.go
```diff
@@ -21,7 +21,7 @@ import (
 const (
 	// IgnoreOperationAnnotationFlag is the name of the command line flag to specify whether the operation annotation
 	// is ignored or not.
-	IgnoreOperationAnnotationFlag = "infrastructure-ignore-operation-annotation"
+	IgnoreOperationAnnotationFlag = "ignore-operation-annotation"
```
rfranzke: Why this change? It's only relevant for `infrastructure` controller

adracus: Yes but this should not be statically encoded into the flag name but rather be done by using the `PrefixOption` operator - for instance, if somebody ever wants to run an infrastructure controller on its own, then you probably wouldn't want to have this 'prefix'.

rfranzke: OK, then let's use the `PrefixOption` operator.

controllers/provider-gcp/pkg/internal/imagevector/imagevector.go
```diff
@@ -0,0 +1,69 @@
...
+
+const (
+	// TerraformerImageName is the name of the Terraformer image.
+	TerraformerImageName = "terraformer"
+)
+
+var imageVector imagevector.ImageVector
+
+func init() {
+	box := packr.New("charts", "../../../charts")
```
rfranzke: Can you do the same for AWS?

controllers/provider-gcp/pkg/internal/infrastructure/infrastructure.go
```diff
@@ -0,0 +1,79 @@
...
+// CleanupKubernetesFirewalls lists all Kubernetes firewall rules and then deletes them one after another.
+//
+// If a deletion fails, this method returns immediately with the encountered error.
+func CleanupKubernetesFirewalls(ctx context.Context, client gcpclient.Interface, projectID, network string) error {
+	firewallNames, err := ListKubernetesFirewalls(ctx, client, projectID, network)
+	if err != nil {
+		return err
+	}
+
+	for _, firewall := range firewallNames {
```
rfranzke: Instead , call `DeleteFirewalls` here.

pkg/controller/cluster.go
rfranzke: Why? :)

rfranzke:
Also, can you please add an example infrastructure manifest (similar to https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-aws/example/Infrastructure.yaml)?!

controllers/provider-gcp/pkg/internal/infrastructure/terraform.go
```diff
@@ -0,0 +1,211 @@
...
+			"region":  infra.Spec.Region,
+			"project": account.ProjectID,
+		},
+		"create": map[string]interface{}{
+			"vpc": createVPC,
+		},
+		"vpc": map[string]interface{}{
+			"name": vpcName,
+		},
+		"clusterName": cluster.Shoot.Name,
```
rfranzke:
```suggestion
		"clusterName": infrastructure.Namespace,
```

rfranzke:
`TypeMeta` missing in `.spec.status.providerStatus`

controllers/provider-gcp/example/Infrastructure.yaml
```diff
@@ -0,0 +1,42 @@
...
+            services: 10.243.0.0/17
+    status:
+      lastOperation:
+        state: Succeeded
+
+---
+apiVersion: extensions.gardener.cloud/v1alpha1
+kind: Infrastructure
+metadata:
+  namespace: default
```
rfranzke: This example probably won't work as it would try to create a VPC with name `default` that already exists in GCP.
Also, could you add here an how an example `Secret` should look like (and for AWS as well, it's missing there) so that users know the needed keys?

controllers/provider-aws/example/Infrastructure.yaml
```diff
@@ -1,4 +1,19 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  namespace: shoot--foo--bar
+  name: cloudprovider
+type: Opaque
+data:
+# accessKeyID: base64(access-key-id)
+# secretAccessKey: base64(secret-access-key)
+---
```
rfranzke: Can we have `metadata.namespace=shoot--foo--bar` for `Cluster` and `Infrastructure` resource as well?

controllers/provider-gcp/example/Infrastructure.yaml
```diff
@@ -0,0 +1,57 @@
...
+  type: gcp
+  region: europe-west1
+  secretRef:
+    namespace: shoot--gcp--foo
+    name: core-gcp
+  providerConfig:
+    apiVersion: gcp.provider.extensions.gardener.cloud/v1alpha1
+    kind: InfrastructureConfig
+    networks:
+      worker: 10.242.0.0/19
```
rfranzke: Similar to https://github.com/gardener/gardener/blob/master/example/90-shoot-gcp.yaml#L17 we should have `internal` here as well

adracus: `internal` is not used, so why have it? In the `aws` example, it's also not a fully functioning shoot.

PR #22, opened 2019-03-11
Add no-op controlplane controller to the local provider

stoyanr:
**What this PR does / why we need it**:
Adds a no-op extension controller that handles ControlPlane resources to the local provider.

**Which issue(s) this PR fixes**:
Fixes #21.

**Special notes for your reviewer**:
This should now be mergeable.

**Release note**:

```improvement operator
Add no-op controlplane controller to the local provider
```

PR #31, opened 2019-03-26
Add functional controlplane controller to the AWS provider

stoyanr:
Adds a functional extension controller that handles ControlPlane resources to the AWS provider.

Change was already discussed with @rfranzke, recommended changes are done. Based on #22 which should be merged first.

**Release note**:

```improvement operator
Add functional controlplane controller to the AWS provider
```


Gopkg.toml
```diff
@@ -9,7 +9,7 @@ required = [
 
 [[override]]
   name = "github.com/gardener/gardener"
-  revision = "bca4ab1463af383ea083e81820f6ee2bf3005e10"
+  revision = "bc8a397c8bb17f5e1490c0ece324ae85884a8438"
```
rfranzke:
```suggestion
  revision = "1ebbe303f020e91d42e8689144bac3904cacb08b"
```

Gopkg.toml
```diff
@@ -23,6 +23,14 @@ required = [
   name = "k8s.io/apimachinery"
   version = "kubernetes-1.13.4"
 
+[[override]]
+  name = "k8s.io/apiserver"
+  version = "kubernetes-1.13.4"
+
+[[override]]
```
rfranzke: Why is this added? I don't see it used anywhere.

Gopkg.toml
```diff
@@ -31,6 +39,21 @@ required = [
   name = "k8s.io/code-generator"
   version = "kubernetes-1.13.4"
 
+[[override]]
```
rfranzke: Same - not used anywhere

controllers/provider-aws/charts/cloud-controller-manager/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
rfranzke: Can you move this chart into the `internal` folder?

controllers/provider-aws/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml
```diff
@@ -0,0 +1,130 @@
...
+        operator: Exists
+      containers:
+      - name: cloud-controller-manager
+        image: {{ index .Values.images "hyperkube" }}
+        imagePullPolicy: IfNotPresent
+        command:
+        - /hyperkube
+        - cloud-controller-manager
+        - --allocate-node-cidrs=true
+        - --cloud-provider={{ .Values.cloudProvider }}
```
rfranzke: We can "hard-code" this to `aws`, we don't need it to be configurable anymore

controllers/provider-aws/charts/cloud-provider-config/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
rfranzke: Can you move this chart into the `internal` folder?

controllers/provider-aws/charts/utils-templates/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
rfranzke: Can you move this chart into the `internal` folder?

controllers/provider-aws/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml
```diff
@@ -0,0 +1,130 @@
...
+  clusterIP: None
+  ports:
+  - name: metrics
+    port: {{ include "cloud-controller-manager.port" . }}
+    protocol: TCP
+  selector:
+    app: kubernetes
+    role: cloud-controller-manager
+---
+apiVersion: {{ include "deploymentversion" . }}
```
rfranzke:
```suggestion
apiVersion: apps/v1
```

(It's not needed anymore)

controllers/provider-aws/charts/utils-templates/templates/_versions.tpl
```diff
@@ -0,0 +1,83 @@
+{{- define "kubeletcomponentconfigversion" -}}
```
rfranzke: This file can be deleted I guess - it's currently not used. We can add it later (with only the content we need) when we require it.

controllers/provider-aws/charts/utils-tls-cipher-suites/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
rfranzke: Can you move this chart into the `internal` folder?

controllers/provider-aws/pkg/controlplane/actuator.go
```diff
@@ -0,0 +1,359 @@
...
+			Name:       gardencorev1alpha1.SecretNameCACluster,
+			CommonName: "kubernetes",
+			CertType:   secrets.CACert,
+		},
+	},
+	SecretConfigsFunc: func(cas map[string]*secrets.Certificate, clusterName string) []secrets.ConfigInterface {
+		return []secrets.ConfigInterface{
+			&secrets.ControlPlaneSecretConfig{
+				CertificateSecretConfig: &secrets.CertificateSecretConfig{
+					Name:         common.CloudControllerManagerDeploymentName,
```
rfranzke: This name should not be imported from the gardener `common` package anymore - instead, we should define our own constant here in the AWS controller somewhere.

controllers/provider-aws/pkg/controlplane/actuator.go
```diff
@@ -0,0 +1,359 @@
...
+					SigningCA:    cas[gardencorev1alpha1.SecretNameCACluster],
+				},
+				KubeConfigRequest: &secrets.KubeConfigRequest{
+					ClusterName:  clusterName,
+					APIServerURL: common.KubeAPIServerDeploymentName,
+				},
+			},
+			&secrets.ControlPlaneSecretConfig{
+				CertificateSecretConfig: &secrets.CertificateSecretConfig{
+					Name:       common.CloudControllerManagerServerName,
```
rfranzke: Same as above

controllers/provider-aws/pkg/controlplane/actuator.go
```diff
@@ -0,0 +1,359 @@
...
+	Path: filepath.Join(aws.ChartsPath, "cloud-provider-config"),
+	Objects: []*chart.Object{
+		{
+			Type: &corev1.ConfigMap{},
+			Name: common.CloudProviderConfigName,
+		},
+	},
+}
+
+const (
```
rfranzke: Already defined here: https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-aws/pkg/aws/types.go#L22-L25

controllers/provider-aws/pkg/controlplane/actuator.go
```diff
@@ -0,0 +1,359 @@
...
+					"name": "AWS_SECRET_ACCESS_KEY",
+					"valueFrom": map[string]interface{}{
+						"secretKeyRef": map[string]interface{}{
+							"key":  SecretAccessKey,
+							"name": common.CloudProviderSecretName,
+						},
+					},
+				},
+			},
+			// TODO Use the higher resource limits below only for the "shooted seed" case
```
rfranzke: This TODO can be removed, right?

controllers/provider-aws/pkg/controlplane/actuator.go
```diff
@@ -0,0 +1,359 @@
...
+		logger: log.Log.WithName("controlplane-controller"),
+	}
+}
+
+// actuator is an Actuator that acts upon and updates the status of ControlPlane resources.
+type actuator struct {
+	scheme       *runtime.Scheme
+	decoder      runtime.Decoder
+	config       *rest.Config
+	cs           kubernetes.Interface
```
rfranzke:
```suggestion
	clientset    kubernetes.Interface
```

controllers/provider-aws/pkg/controlplane/actuator.go
```diff
@@ -0,0 +1,359 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "could not apply configuration chart for controlplane '%s'", objectName(cp))
+	}
+
+	// Get cloudprovider secret and configmap from cluster
+	cpSecret, err := a.getSecret(cp.Namespace, common.CloudProviderSecretName)
+	if err != nil {
+		return err
+	}
+	senConfigMap, err := a.getConfigMap(cp.Namespace, common.CloudProviderConfigName)
```
rfranzke:
```suggestion
	awsConfigMap, err := a.getConfigMap(cp.Namespace, common.CloudProviderConfigName)
```

controllers/provider-aws/pkg/controlplane/actuator.go
```diff
@@ -0,0 +1,359 @@
...
+	// Delete secrets
+	a.logger.Info("Deleting secrets", "controlplane", objectName(cp))
+	if err := controlPlaneSecrets.Delete(a.cs, cp.Namespace); err != nil {
+		return errors.Wrapf(err, "could not delete secrets for controlplane '%s'", objectName(cp))
+	}
+
+	return nil
+}
+
+func (a *actuator) getSecret(namespace, name string) (*corev1.Secret, error) {
```
rfranzke: Hm, why do we need a wrapper function here? Also, I'd rather prefer to use the controller-runtime client directly instead of the clientset.

controllers/provider-aws/pkg/controlplane/actuator.go
```diff
@@ -0,0 +1,359 @@
...
+func (a *actuator) getConfigMap(namespace, name string) (*corev1.ConfigMap, error) {
+	cm, err := a.cs.CoreV1().ConfigMaps(namespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not get configmap '%s/%s'", namespace, name)
+	}
+	return cm, nil
+}
+
+func objectName(obj runtime.Object) string {
+	k, _ := client.ObjectKeyFromObject(obj) // nolint:errcheck,gosec
```
rfranzke: Is the metadata accessor https://github.com/gardener/gardener-extensions/blob/e884183001bedd48de979b6b21b3b03e53786b68/pkg/controller/predicate.go#L55-L60 an alternative?

Gopkg.toml
stoyanr: Because of dep. It's a transitive dependency. After I changed the gardener dependency to a newer commit, dep downloaded `master` instead of `kubernetes-1.13.4`, which didn't build. So I had to pin it to its correct version here.

Gopkg.toml
stoyanr: See above, I had the same issue here as well.

Gopkg.toml
timuthy: Can we wait for https://github.com/gardener/gardener/pull/863?

This PR removes the Tiller dependency which is in `1ebbe303f020e91d42e8689144bac3904cacb08b` and would avoid pining some dependencies.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: Thanks, it should actually be `cpConfigMap` to be consistent with the rest.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: The idea is to get a string containing a namespaced name "<namespace>/<name>" from the object. It can be done in multiple ways, the one used here is from my perspective the most straightforward one. If you prefer, I can use the metadata accessor of course.

Gopkg.toml
stoyanr: No need to wait from my perspective, dependencies could be unpinned in another PR as well. But as you decide.

controllers/provider-aws/charts/cloud-controller-manager/Chart.yaml
stoyanr: Done.

controllers/provider-aws/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml
stoyanr: Done.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: I thought we might actually want to properly take care of the "shooted seed" case some day. But you are right, since it's not clear how to do it this TODO is rather misplaced. Removed.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: Done, `gcs` renamed to `gardenerClientset` as well.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: Now renamed.

controllers/provider-aws/charts/cloud-provider-config/Chart.yaml
stoyanr: Done.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: Refactored as suggested and removed the wrapper functions.

pkg/controller/shoot.go
```diff
@@ -0,0 +1,62 @@
...
+
+package controller
+
+import (
+	gardencorev1alpha1 "github.com/gardener/gardener/pkg/apis/core/v1alpha1"
+	gardenv1beta1 "github.com/gardener/gardener/pkg/apis/garden/v1beta1"
+)
+
+// Shoot provides convenience methods for getting various Shoot properties.
+type Shoot struct {
```
adracus: Instead of defining another `Shoot` type, please write these functions here as dedicated helper functions taking a regular `Shoot` as argument.

pkg/controller/shoot.go
```diff
@@ -0,0 +1,62 @@
...
+
+// GetKubernetesVersion returns this shoot's Kubernetes version.
+func (s *Shoot) GetKubernetesVersion() string {
+	return s.Shoot.Spec.Kubernetes.Version
+}
+
+// GetPodNetwork returns this shoot's pod network CIDR.
+func (s *Shoot) GetPodNetwork() gardencorev1alpha1.CIDR {
+	cloud := s.Shoot.Spec.Cloud
+	if cloud.AWS != nil {
```
adracus: I'd recommend a switch here.

pkg/util/util.go
```diff
@@ -26,3 +34,41 @@ func ContextFromStopChannel(stopCh <-chan struct{}) context.Context {
 
 	return ctx
 }
+
+// DNSNamesForService returns the possible DNS names for a service with the given name and namespace
+func DNSNamesForService(name, namespace string) []string {
```
adracus: I don't see how all these new methods are general utils for _all_ our controllers - they only make sense for the Control Plane controller (now). I'd rather move them into a `controlplane` package as we did it for `infrastructure` etc.

controllers/provider-aws/pkg/controlplane/actuator.go
```diff
@@ -0,0 +1,343 @@
...
+
+// Reconcile reconciles the given controlplane and cluster, creating or updating the additional Shoot
+// control plane components as needed.
+func (a *actuator) Reconcile(
+	ctx context.Context,
+	cp *extensionsv1alpha1.ControlPlane,
+	cluster *extensionscontroller.Cluster,
+) error {
+	// Decode providerConfig
+	cpConfig := &apisaws.ControlPlaneConfig{}
```
adracus: This method is very big and complex - please split it into various methods.

controllers/provider-aws/charts/utils-templates/Chart.yaml
stoyanr: Done.

controllers/provider-aws/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml
stoyanr: Done.

controllers/provider-aws/charts/utils-templates/templates/_versions.tpl
stoyanr: Done.

controllers/provider-aws/charts/utils-tls-cipher-suites/Chart.yaml
stoyanr: Done.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: I think the best place is right here in `actuator.go` and it should not be exported because it's unlikely to be used somewhere else. Done.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: Done.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: Thanks, missed them at first. Done.

pkg/controller/shoot.go
stoyanr: Are you sure? I don't see how this helps or adds any clarity. The preferred style in Go (and other languages) for helpers that act on a common types would be to add them as methods. @rfranzke, your opinion?

pkg/controller/shoot.go
stoyanr: Switch on what exactly? I thought about a switch and looked at how similar stuff is done in Gardener, I was only able to find `nil` checks. It's also more reliable in my opinion. But I can of course write it as a switch if you point me to the right example.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: "Very big and complex" is very subjective in my opinion. Could you be somewhat more specific regarding what should be extracted as functions, and why?
I personally am not a big fan of extracting random pieces of code just to make methods short. A longer methods that has no branch complexity is clearer and easier to understand than the same method cut into scattered pieces that have to be pieced together again by the reader each time.

stoyanr:
> PR structure is good, however there is no test at all! Please split your functions into testable pieces and write meaningful tests for all of them.

Tests in this project are currently virtually non-existent. There is not a single test in `controller/*` or `pkg/controller`. Why apply a double standard to this PR compared to the existing code?
Regarding splitting my functions into testable pieces, see my comment regarding `actuator.Reconcile()`. All other functions are already properly split I believe.

pkg/util/util.go
stoyanr: I put them where we agreed with @rfranzke they should be. But I think you have a point, so I moved them to `controlplane`.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: One piece of code that could be extracted would be determining the `subnetID` and `zone` from the infrastructure status. Another one to consider would be computing the checksums, but I am somewhat reluctant about this one, I personally prefer to see the entire flow in one place since the steps depend on each other.

controllers/provider-aws/example/controlplane.yaml
```diff
@@ -0,0 +1,54 @@
...
+      cloud:
+        profile: aws
+        aws:
+          networks:
+            pods: 10.250.0.0/19
+      kubernetes:
+        version: 1.13.4
+      hibernation:
+        enabled: false
+    status:
```
mvladev: status should not be put in the resource.

stoyanr: This is the status of the `Shoot` embedded in `Cluster`, not the status of the `Cluster` itself. It's also exactly the same as in https://github.com/gardener/gardener-extensions/blob/3e20d823b9c50a974f835c7e94a7669d5800615c/controllers/provider-aws/example/Infrastructure.yaml#L16. It also could be important, e.g. if the Shoot is in a failed state the controller could (in theory) behave differently.

stoyanr:
@rfranzke @adracus @mvladev Thanks for your comments! Something rather important that I just thought of and didn't see discussed or commented yet: should the controller watch and react on changes to the resources it creates? E.g. if I go and delete or change the CCM deployment, should the controller put it back? A classical Kubernetes operator would do this, so I rather think our controller should do it as well. What do you think?

If this should be introduced, I would vote to put this behaviour in a different PR rather than making this one even more complex.

pkg/controller/shoot.go
stoyanr: Actually, there is one good reason for not defining a `Shoot` type and it's that you wouldn't need to create an instance in order to work with the shoot, which makes the client code slightly less complex. So reworked as suggested.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: Extracted a few methods out of it now, basically everything that I think would make sense to extract. Hopefully it's less "very big and complex" now.

Gopkg.toml
timuthy: Okay, I just figured out that we cannot refer to a Gardner version which supports Kubernetes 1.14 because the Controller-Runtime is not 1.14 compliant yet. So let's stick to `1ebbe303f020e91d42e8689144bac3904cacb08b`

stoyanr: Ok, already done.

controllers/provider-aws/pkg/controlplane/actuator.go
```diff
@@ -0,0 +1,363 @@
...
+DisableSecurityGroupIngress=true
+KubernetesClusterTag=%q
+KubernetesClusterID=%q
+Zone=%q
+`,
+		vpcID, subnetID, clusterID, clusterID, zone)
+}
+
+func objectName(obj runtime.Object) string {
+	k, _ := client.ObjectKeyFromObject(obj)
```
zanetworker: Why are you ignoring the error here?

controllers/provider-aws/pkg/controlplane/actuator.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	// Collect additional configuration chart values
+	values := map[string]interface{}{
+		"cloudProviderConfig": getCloudProviderConfig(infraStatus.VPC.ID, subnetID, zone, cp.Namespace),
+	}
+
+	// Apply config chart
+	a.logger.Info("Applying configuration chart", "controlplane", objectName(cp), "chart", configChart.Name, "values", values)
+	err = configChart.Apply(ctx, a.gardenerClientset, a.chartApplier, cp.Namespace, cluster.Shoot, nil, nil, values)
```
zanetworker:
```suggestion
	if err := configChart.Apply(ctx, a.gardenerClientset, a.chartApplier, cp.Namespace, cluster.Shoot, nil, nil, values); err != nil {
	return errors.Wrapf(err, "could not apply configuration chart for controlplane '%s'", objectName(cp))
	}
```

controllers/provider-aws/pkg/controlplane/actuator.go
```diff
@@ -0,0 +1,363 @@
...
+
+	// Collect additional CCM chart values
+	values = make(map[string]interface{})
+	if cpConfig.CloudControllerManager != nil {
+		values["featureGates"] = cpConfig.CloudControllerManager.FeatureGates
+	}
+
+	// Apply CCM chart
+	a.logger.Info("Applying CCM chart", "controlplane", objectName(cp), "chart", ccmChart.Name, "values", values)
+	err = ccmChart.Apply(ctx, a.gardenerClientset, a.chartApplier, cp.Namespace, cluster.Shoot, imagevector.ImageVector, checksums, values)
```
zanetworker:
```suggestion
	if err := ccmChart.Apply(ctx, a.gardenerClientset, a.chartApplier, cp.Namespace, cluster.Shoot, imagevector.ImageVector, checksums, values); err != nil {
			return errors.Wrapf(err, "could not apply CCM chart for controlplane '%s'", objectName(cp))
	}
```

pkg/controller/shoot.go
adracus: If you have lots of `if` / `else` blocks, you can just switch on 'nothing' and write your `if`s as `case`s:
```go
switch {
  case a:
  case b:
}
```

adracus:
@stoyanr virtually no tests? I'm also right now filing a PR (#28) for the GCP infrastructure controller where I write tests as far as possible.
Regarding the tests in the `controller` package: Yes, there might not be that much, but if there is something testable, we tested it. If you're now writing new code and something is easily testable (which it definitely is, looking at your code) I'd definitely recommend not just throwing code in but writing actual tests for it. If something is already not that good, we should agree on not making it worse.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: Because it's a private method and I am quite sure an error can never occur. Also, in case it does occur, k would be an empty key that would be returned as "/". Since this method is only used in log entries, that would be not ideal, but fine. Returning the error from this method would complicate the client code too much with no benefit whatsoever.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: Why?

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: Why?

stoyanr:
> @stoyanr virtually no tests? I'm also right now filing a PR (#28) for the GCP infrastructure controller where I write tests as far as possible.
Your PR is not there so far, I didn't know about it.

> Regarding the tests in the `controller` package: Yes, there might not be that much, but if there is something testable, we tested it. If you're now writing new code and something is easily testable (which it definitely is, looking at your code) I'd definitely recommend not just throwing code in but writing actual tests for it. If something is already not that good, we should agree on not making it worse.
The only controller tests I was able to find are in `controllers/os-coreos-alicloud/pkg/coreos-alicloud/internal`, and that really only covers the template generator part of the actuator. That's about 5% test coverage or so. I don't think I am making anything worse, sorry.
Which parts of my code you find easily testable? The problem that I see is that the current design of controllers doesn't lend itself towards being easily testable, and that's why there are no tests in the first place. This would have to be changed in order to write tests for them, including changing some parts that are in Gardener. I have some ideas and I have a meeting with @rfranzke  tomorrow to discuss them. But I still think that this PR can be merged as it is. I will definitely look into tests next, among other things.

controllers/provider-aws/pkg/controlplane/actuator.go
zanetworker: Consistency across your code (at some points you are using a one-line if at others you separate), and fewer lines.

controllers/provider-aws/pkg/controlplane/actuator.go
zanetworker: Consistency across your code (at some points you are using a one-line if at others you separate), and fewer lines.

controllers/provider-aws/pkg/controlplane/actuator.go
zanetworker: there are benefits, you don't risk unexpected behavior that are made harder to debug due to lack of err. Also as described in the documentation, the error is returned to handle this case:  

> is returned when an object implements the List style interfaces but not the Object style interfaces.

So it does not cause alot of trouble IMHO to not ignore the error.

controllers/provider-aws/pkg/controlplane/actuator.go
zanetworker: For consistency across your code, and for shorter lines of code.

controllers/provider-aws/pkg/controlplane/actuator.go
zanetworker: For consistency across your code, and for shorter lines of code.

pkg/controller/shoot.go
stoyanr: Yes, you can do this, but it's by no means a widely used style. It's not present in any of the common Go coding style guidelines I know of. Do you have a coding guideline for Gardener that actually encourages such a style? If not, I would recommend to write one as you can't expect  contributors to know that they should write their code like this.

stoyanr: Reworked as asked.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: Done.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: Done.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: OK, so I added 3 more lines of code and the behaviour is exactly the same result as before. Hopefully the benefits are worth it.

controllers/provider-aws/charts/internal/cloud-controller-manager/values.yaml
```diff
@@ -0,0 +1,21 @@
+cloudProvider: aws
```
rfranzke: This can be removed now.

controllers/provider-aws/charts/internal/utils-templates/templates/_resources.tpl
```diff
@@ -0,0 +1,12 @@
+{{/*
```
rfranzke: Not used, can be removed.

controllers/provider-aws/charts/internal/utils-templates/README.md
```diff
@@ -0,0 +1,25 @@
+# Important
```
rfranzke: Not used, can be removed.

controllers/provider-aws/charts/internal/cloud-controller-manager/charts/utils-templates
```diff
@@ -0,0 +1 @@
+../../utils-templates
```
rfranzke: Not used, can be removed.

controllers/provider-aws/example/controlplane.yaml
```diff
@@ -0,0 +1,54 @@
...
+    kind: CloudProfile
+  seed:
+    apiVersion: garden.sapcloud.io/v1beta1
+    kind: Seed
+  shoot:
+    apiVersion: garden.sapcloud.io/v1beta1
+    kind: Shoot
+    spec:
+      cloud:
+        profile: aws
```
rfranzke: Is this needed in the example?

pkg/controller/controlplane/utils.go
```diff
@@ -0,0 +1,62 @@
...
+	"encoding/json"
+	"fmt"
+
+	gardenv1beta1 "github.com/gardener/gardener/pkg/apis/garden/v1beta1"
+	"github.com/gardener/gardener/pkg/utils"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// DNSNamesForService returns the possible DNS names for a service with the given name and namespace
+func DNSNamesForService(name, namespace string) []string {
```
rfranzke: All the functions here can be tested nicely, I guess that's what @adracus meant. :) Can we have a few tests for these utilities, please?

pkg/controller/shoot.go
```diff
@@ -0,0 +1,54 @@
...
+
+package controller
+
+import (
+	gardencorev1alpha1 "github.com/gardener/gardener/pkg/apis/core/v1alpha1"
+	gardenv1beta1 "github.com/gardener/gardener/pkg/apis/garden/v1beta1"
+)
+
+// GetKubernetesVersion returns the Kubernetes version of the given Shoot.
+func GetKubernetesVersion(shoot *gardenv1beta1.Shoot) string {
```
rfranzke: TBH, I would remove this function. It doesn't bring any benefit other than it's a little bit shorter. Access `shoot.Spec.Kubernetes.Version` directly.

pkg/controller/shoot.go
```diff
@@ -0,0 +1,54 @@
...
+		return *cloud.Alicloud.Networks.K8SNetworks.Pods
+	case cloud.Local != nil:
+		return *cloud.Local.Networks.K8SNetworks.Pods
+	default:
+		return ""
+	}
+}
+
+// GetReplicas returns the woken up replicas of the given Shoot.
+func GetReplicas(shoot *gardenv1beta1.Shoot, wokenUp int) int {
```
rfranzke: This could also be tested shortly.

pkg/controller/shoot.go
stoyanr: I believe it's still useful, as it abstracts away details of the `Shoot` structure. If they change, there would be a single place to fix.

rfranzke: Hmm not sure, you still provide `*garden1beta1.Shoot` as an input, so I cannot throw `*gardencorev1alpha1.Shoot` into the function. Thus, when having a new shoot I would have to adapt this code anyways, or?

pkg/controller/shoot.go
stoyanr: Done.

pkg/controller/shoot.go
stoyanr: OK, done.

controllers/provider-aws/example/controlplane.yaml
stoyanr: No, removed.

controllers/provider-aws/charts/internal/cloud-controller-manager/charts/utils-templates
stoyanr: Done.

controllers/provider-aws/charts/internal/utils-templates/README.md
stoyanr: Done.

controllers/provider-aws/charts/internal/cloud-controller-manager/values.yaml
stoyanr: Done.

controllers/provider-aws/charts/internal/utils-templates/templates/_resources.tpl
stoyanr: Done.

pkg/controller/controlplane/utils.go
stoyanr: Done.

stoyanr:
@rfranzke @adracus I believe all comments are addressed now and there are adequate tests for all utility functions.

PR #32, opened 2019-03-27
WIP: Implementation JeOS Operating System Configuration Controller

pablochacin:
This commit introduces an operating system configuration controller
for SUSE JeOS. The controller is based in alicloud's coreos
controller. The common code is moved to a generic os-common controller
which implement the bulk of the logic. JeOS controller implementation
is limited to provide a template for rendering the cloud-init
configuration.

Fixes #

TODO:
[x] Test ad Document generation of controller examples from os-common charts

**Release note**:
```noteworthy operator
Support for SUSE JeOS nodes via an OpertingSystemConfiguration Controller
```


controllers/os-common/charts/templates/rbac.yaml
```diff
@@ -0,0 +1,67 @@
...
+  labels:
+    app.kubernetes.io/name: gardener-extension-os-{{ .Values.OperatingSystem.Name }}
+    helm.sh/chart: gardener-extension-os-{{ .Values.OperatingSystem.Name }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+rules:
+- apiGroups:
+  - extensions.gardener.cloud
+  resources:
+  - operatingsystemconfigs
+  - operatingsystemconfigs/status
```
mvladev: I'll move this in a separate rule
```suggestion
- apiGroups:
  - extensions.gardener.cloud
  resources:
  - operatingsystemconfigs/status
  verbs:
  - patch
  - update
```

as `Status` subresources only support `patch` and `update`.

controllers/os-common/charts/templates/templates
```diff
@@ -0,0 +1 @@
+templates
```
mvladev: wrong link?

controllers/os-common/pkg/actuator/actuator.go
```diff
@@ -0,0 +1,66 @@
...
+// NewActuator creates a new actuator with the given logger.
+func NewActuator(osName string, generator generator.Generator) operatingsystemconfig.Actuator {
+	return &actuator{
+		logger:    log.Log.WithName(osName + "-operatingsystemconfig-actuator"),
+		osName:    osName,
+		generator: generator,
+	}
+}
+
+func (a *actuator) InjectScheme(scheme *runtime.Scheme) error {
```
mvladev: is `scheme` used for anything?

pablochacin: Most of this code is copied from Alicloud's controller. I think @adracus can comment on this better.

controllers/os-common/charts/templates/templates
pablochacin: FIxed

controllers/os-common/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: apps/v1
```
rfranzke: See above, don't do that, please.

extensions.yaml
```diff
@@ -4,6 +4,9 @@ extensions:
 - name: os-coreos
   gitHubRepo: https://github.com/gardener/gardener-extensions
   path: controllers/os-coreos
+- name: os-suse-jeos
+  gitHubRepo: file:///home/pchacin/workspace/go/src/github.com/gardener/gardener-extensions
```
rfranzke: fix

controllers/os-common/README.md
```diff
@@ -0,0 +1,82 @@
+# [Gardener Extension for OS Configurations](https://gardener.cloud)
```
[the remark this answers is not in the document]
pablochacin: The JeOS controller has its own README. Can you explain what this comment refers to?

controllers/os-common/charts/templates/deployment.yaml
pablochacin: Can you please be more specific what change you suggest?  It is having common templates?

controllers/os-common/README.md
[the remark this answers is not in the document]
rfranzke: OK, good, thanks. Then just remove the whole os-common directory

controllers/os-common/charts/templates/deployment.yaml
rfranzke: >  In the future we plan to move out the extensions into dedicated repositories, so already having most of the files needed in place will make that easier.

--> Everything that we share (and that's already too much) will complicate this process. Let's try to not share as much as possible so that the move-out gets easier.

pablochacin: I understand sharing templates is problematic when splitting into multiple repositories. I will move the templates.

pablochacin: With respect of the os-common. The idea is to have a common implementation that can be used for other controllers, including the existing alicloud controller. This can be considered a library, much in the same way that the share code in `gardener-extensions/pkg/controller'.  Having a separated repository for each controller will not be more complicated by the fact of using such common code.

rfranzke: If you want to share code between OS controllers then https://github.com/gardener/gardener-extensions/tree/master/pkg/controller/operatingsystemconfig is the correct place (probably a subpackage there). Everything in the `controllers` folder is to be moved out in a dedicated repo, thus it should only contain real controllers and not libraries.

controllers/os-common/charts/templates/rbac.yaml
pablochacin: As you said, having these extra verbs are not needed, but neither do they harm and adding one additional rule add some verbosity.

pkg/controller/operatingsystemconfig/os-common/README.md
```diff
@@ -0,0 +1,82 @@
+# [Gardener Extension for OS Configurations](https://gardener.cloud)
```
rfranzke: File should be removed (the content is misleading, no actual controller is implement here, only library functions)

pkg/controller/operatingsystemconfig/os-common/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: apps/v1
```
rfranzke: remove whole chart - no actual controller developed here.

pkg/controller/operatingsystemconfig/os-common/example/operatingsystemconfig.yaml
```diff
@@ -0,0 +1,44 @@
+---
```
rfranzke: remove, same reason as above, entire example folder should go away

controllers/os-suse-jeos/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: gardener-extension-os-{{- .Values.OperatingSystem.Name }}
```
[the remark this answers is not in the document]
pablochacin: > I would not make the name configurable here - otherwise I could name it `ubuntu` although it's only doing `jeos`.
> 
> replace all `{{- .Values.OperatingSystem.Name }} ` with `suse-jeos` and remove it from the `values.yaml`

The rationale was to allow easy copy/paste of the templates from os-common, without needing manual editing. I don't see it makes any harm.

controllers/os-suse-jeos/example/controller-registration-local.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: core.gardener.cloud/v1alpha1
```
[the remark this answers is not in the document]
pablochacin: This is the one required for registering the controller locally. The helm chart is "dummy" but allows the controller to be properly "installed" by gardener.

pkg/controller/operatingsystemconfig/os-common/README.md
pablochacin: The entire file? it provide information about how to use the library.

pkg/controller/operatingsystemconfig/os-common/charts/templates/deployment.yaml
pablochacin: Are templates. It is explained in the README.

pkg/controller/operatingsystemconfig/os-common/example/operatingsystemconfig.yaml
pablochacin: I don't understand the logic of removing anything that makes the library useful and reusable: README, examples, ...

controllers/os-suse-jeos/charts/templates/deployment.yaml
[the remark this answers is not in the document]
rfranzke: The problem is that the user of this Helm chart can overwrite this during deployment, and then we can get into the situation (name is `ubuntu` but actual controller only operates on `jeos`).

controllers/os-suse-jeos/example/controller-registration-local.yaml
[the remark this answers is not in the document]
rfranzke: OK, I got confused because you provide an image tag. Can you then remove

```yaml
      values:
        image:
          tag: 0.1.0-dev
```

It seems not needed, right?

pkg/controller/operatingsystemconfig/os-common/README.md
```diff
@@ -0,0 +1,82 @@
...
+Please find more information regarding the extensibility concepts and a detailed proposal [here](https://github.com/gardener/gardener/blob/master/docs/proposals/01-extensibility.md).
+
+----
+
+## How to use os-common in a new operating system configuration controller
+
+When implemening a controller for a specific operating system, it is necessary to provide:
+* A command line application for launching the controller
+* A template for translating the `cloud-config` to the format requried by the operating system.
+* Alternatevely, a new generator can also be provided, in case the transformations required by
```
rfranzke:
```suggestion
* Alternatively, a new generator can also be provided, in case the transformations required by
```

pkg/controller/operatingsystemconfig/os-common/README.md
rfranzke: OK, agreed, didn't see that. 👍

pkg/controller/operatingsystemconfig/os-common/README.md
```diff
@@ -0,0 +1,82 @@
...
+      namespace: default
+  command: <machine configuration command>
+  units:
+  - docker-monitor.service
+  - kubelet-monitor.service
+  - kubelet.service
+```
+The secret has one data key `cloud_config` that stores the generation.
+
+An example for a `ControllerRegistration` resource that can be used to register this controller to Gardener can be found [here](example/controller-registration.yaml).
```
rfranzke: Remove this part, there cannot be an example registration for a library.

pkg/controller/operatingsystemconfig/os-common/charts/templates/deployment.yaml
rfranzke: Nobody will maintain this. It will run out of date. Let's not have this Helm chart here. If somebody wants to develop a new OS controller he/she can easily copy the Helm chart of another existing controller which is actually working and up-to-date.

pkg/controller/operatingsystemconfig/os-common/example/operatingsystemconfig.yaml
rfranzke: Logic: Double maintenance effort. Let's only maintain examples for concrete controllers, not for any libraries (examples would anyway not work without a concrete controller implementation).

controllers/os-suse-jeos/charts/templates/deployment.yaml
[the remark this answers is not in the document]
pablochacin: Changed.

controllers/os-suse-jeos/example/controller-registration-local.yaml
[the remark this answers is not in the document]
pablochacin: Fixed.

pkg/controller/operatingsystemconfig/os-common/charts/templates/deployment.yaml
pablochacin: Removed

pkg/controller/operatingsystemconfig/os-common/actuator/add.go
```diff
@@ -0,0 +1,41 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// Default options
+var options = controller.Options{}
+
+// AddWithOptions adds a controller with the given Options to the given manager.
+// The opts.Reconciler is being set with a newly instantiated actuator.
+func AddWithOptions(mgr manager.Manager, os string, generator generator.Generator, opts controller.Options) error {
```
adracus: This has to be renamed to `AddToManagerWithOptions`

pkg/controller/operatingsystemconfig/os-common/actuator/add.go
```diff
@@ -0,0 +1,41 @@
...
+	return operatingsystemconfig.Add(mgr, operatingsystemconfig.AddArgs{
+		Actuator:          NewActuator(os, generator),
+		Type:              os,
+		ControllerOptions: opts,
+	})
+}
+
+// Add adds a controller with the default Options.
+func Add(mgr manager.Manager, os string, generator generator.Generator) error {
```
adracus: Rename to `AddToManager`

pkg/controller/operatingsystemconfig/os-common/actuator/actuator_reconcile.go
```diff
@@ -0,0 +1,109 @@
...
+func (a *actuator) reconcile(ctx context.Context, config *extensionsv1alpha1.OperatingSystemConfig) ([]byte, *string, []string, error) {
+	cloudConfig, cmd, err := a.cloudConfigFromOperatingSystemConfig(ctx, config)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("Could not generate cloud config: %v", err)
+	}
+
+	return []byte(cloudConfig), cmd, operatingSystemConfigUnitNames(config), nil
+}
+
+func (a *actuator) cloudConfigFromOperatingSystemConfig(ctx context.Context, config *extensionsv1alpha1.OperatingSystemConfig) ([]byte, *string, error) {
```
adracus: Please don't create a 'common' actuator but rather expose these methods as utility functions, e.g. in this case, create a function
```go
func CloudConfigFromOperatingSystemConfig(ctx context.Context, c client.Client, config *extensionsv1alpha1.OperatingSystemConfig) ([]byte,*string, error)
```

pkg/controller/operatingsystemconfig/os-common/actuator/actuator_reconcile.go
```diff
@@ -0,0 +1,109 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func (a *actuator) reconcile(ctx context.Context, config *extensionsv1alpha1.OperatingSystemConfig) ([]byte, *string, []string, error) {
+	cloudConfig, cmd, err := a.cloudConfigFromOperatingSystemConfig(ctx, config)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("Could not generate cloud config: %v", err)
```
adracus: All errors should start with lower case

pkg/controller/operatingsystemconfig/os-common/app/app.go
```diff
@@ -0,0 +1,77 @@
...
+				controllercmd.LogErrAndExit(err, "Could not instantiate manager")
+			}
+
+			if err := extcontroller.AddToScheme(mgr.GetScheme()); err != nil {
+				controllercmd.LogErrAndExit(err, "Could not update manager scheme")
+			}
+
+			ctrlOpts.Completed().Apply(options)
+
+			if err := actuator.Add(mgr, osName, generator); err != nil {
```
adracus: An actuator should not be '`Add`ed' to a manager, only a controller.

pkg/controller/operatingsystemconfig/os-common/generator/generator.go
```diff
@@ -0,0 +1,49 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package generator
+
+// Generator renders an OperatingSystemConfig into a
+// representation suitable for an specific OS
+type Generator interface {
+	Generate(*OperatingSystemConfig) ([]byte, *string, error)
```
adracus: As of now, the result type `([]byte, *string, error)` is not really descriptive - maybe add named result types?
Also, I think that the `Generate` call should not return the `cmd *string` as this is anyways a 'constant'. The `cmd` should be anyways available for the caller.

pablochacin: The command is fixed for an operating system controller, but not to the library, I just moved the logic from the actuator to the generator, so that each os can implement a different logic.

pablochacin: +1 with named result. types.

pkg/controller/operatingsystemconfig/os-common/app/app.go
pablochacin: Agree. In the refactoring this function ended in the actuator package. Moved to a new 'oscommon' package,

pkg/controller/operatingsystemconfig/oscommon/add.go
```diff
@@ -0,0 +1,42 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// Default options
+var options = controller.Options{}
+
+// AddWithOptions adds a controller with the given Options to the given manager.
+// The opts.Reconciler is being set with a newly instantiated actuator.
+func AddManagerWithOptions(mgr manager.Manager, os string, generator generator.Generator, opts controller.Options) error {
```
adracus: `AddToManagerWithOptions`

pkg/controller/operatingsystemconfig/oscommon/add.go
```diff
@@ -0,0 +1,42 @@
...
+	return operatingsystemconfig.Add(mgr, operatingsystemconfig.AddArgs{
+		Actuator:          actuator.NewActuator(os, generator),
+		Type:              os,
+		ControllerOptions: opts,
+	})
+}
+
+// Add adds a controller with the default Options.
+func AddManager(mgr manager.Manager, os string, generator generator.Generator) error {
```
adracus: `AddToManager`

pkg/controller/operatingsystemconfig/oscommon/actuator/actuator_reconcile.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return a.generator.Generate(&generator.OperatingSystemConfig{
+		Bootstrap: config.Spec.Purpose == extensionsv1alpha1.OperatingSystemConfigPurposeProvision,
+		Files:     files,
+		Units:     units,
+		Path:	   config.Spec.ReloadConfigFilePath,
+	})
+}
+
+func (a *Actuator) DataForFileContent(ctx context.Context, namespace string, content *extensionsv1alpha1.FileContent) ([]byte, error) {
```
adracus: Instead of creating a 'common' actuator, can you please rather declare this as a top level function, i.e.
```go
DataForFileContent(ctx context.Context, c client.Client, content *extensionsv1alpha1.FileContent) ([]byte, error)
```
Then, an implementor can just call these functions.

pkg/controller/operatingsystemconfig/oscommon/add.go
pablochacin: Sorry. Fixed.

pkg/controller/operatingsystemconfig/oscommon/actuator/actuator_reconcile.go
pablochacin: Done

pablochacin:
@adracus @rfranzke I think I have addressed all the issues on your reviews.

pablochacin:
@vlerenc  as commented during the community meeting on Friday April 5th, the general availability of the image will depend on having a setup for testing the image, possibly using the test-machinery framework used in the CI pipeline.

PR #39, opened 2019-04-02
Refactor AWS controlplane controller for testability and add tests

stoyanr:
Refactors the AWS controller for ControlPlane resources for testability and adds tests for this controller.

The approach taken has been already discussed with @rfranzke and @adracus, all recommendations have been taken into account. Based on #31 which should be merged first.

**Release note**:

```improvement operator
Refactor controlplane controllers add tests
```


rfranzke:
Now that #31 is merged the first commit of this PR can be dropped, right?

stoyanr:
> Now that #31 is merged the first commit of this PR can be dropped, right?

/cc @rfranzke Done, any chance to get a review?

controllers/provider-aws/pkg/controlplane/internal/helper.go
```diff
@@ -0,0 +1,235 @@
...
+)
+
+// Object names
+const (
+	cloudControllerManagerDeploymentName = "cloud-controller-manager"
+	cloudControllerManagerServerName     = "cloud-controller-manager-server"
+)
+
+// Helper is a control plane actuator helper type.
+type Helper struct {
```
adracus: Don't create a `struct` - Please create functions for each one of them without this `Helper` struct.
This means e.g. for `Secrets`, split up as follows:
`Secrets`:
* `DeploySecrets`
* `DeleteSecrets`

controllers/provider-aws/pkg/controlplane/internal/helper_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+func encode(obj runtime.Object) []byte {
+	data, _ := json.Marshal(obj)
+	return data
+}
+
+func decode(data []byte, _ *schema.GroupVersionKind, obj runtime.Object) (runtime.Object, *schema.GroupVersionKind, error) {
+	var err error
+	switch x := obj.(type) {
+	case *apisaws.ControlPlaneConfig:
```
adracus: Why this switch? You're anyways always doing the same.

controllers/provider-aws/pkg/imagevector/imagevector.go
```diff
@@ -23,11 +23,9 @@ import (
...
 )
 
-// ImageVector is the image vector that contains al the needed images
-var ImageVector imagevector.ImageVector
-
-func init() {
-	var err error
-	ImageVector, err = imagevector.ReadFile(filepath.Join(aws.ChartsPath, "images.yaml"))
+// GetImageVector returns the image vector that contains al the needed images.
+func GetImageVector() imagevector.ImageVector {
```
adracus: `GetImageVector()` -> `ImageVector()`

controllers/provider-aws/pkg/imagevector/imagevector.go
```diff
@@ -23,11 +23,9 @@ import (
...
+	imageVector, err := imagevector.ReadFile(filepath.Join(aws.ChartsPath, "images.yaml"))
```
adracus: Please use `packr` as in the other controllers.

controllers/provider-aws/pkg/infrastructure/add.go
```diff
@@ -41,7 +41,7 @@ var (
 
 func init() {
 	addToManagerBuilder.Register(Add)
-	terraformerImage, err := imagevector.ImageVector.FindImage(aws.TerraformerImageName, "", "")
+	terraformerImage, err := imagevector.GetImageVector().FindImage(aws.TerraformerImageName, "", "")
```
adracus: Write a method for this as in the other controllers (i.e. `TerraformerImage()`).

pkg/controller/controlplane/utils_test.go
```diff
@@ -0,0 +1,122 @@
...
+				"test-secret":     "8bafb35ff1ac60275d62e1cbd495aceb511fb354f74a20f7d06ecb48b3a68432",
+				"test-config":     "08a7bc7fe8f59b055f173145e211760a83f02cf89635cef26ebb351378635606",
+				"existing-secret": "cc5a179ccb1ea04bbf07089102398ca852fc6ac4e5c076ab4451e191102b240f",
+				"existing-config": "66e322807a642afda73ea4696ff94362a3b402fcdbe4c5676543478c63f26bb9",
+			}))
+		})
+	})
+})
+
+func getSecret(namespace, name string, data map[string][]byte) *corev1.Secret {
```
adracus: `getSecret` -> `mkSecret`

pkg/controller/controlplane/utils_test.go
```diff
@@ -0,0 +1,122 @@
...
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Data: data,
+	}
+}
+
+func getConfigMap(namespace, name string, data map[string]string) *corev1.ConfigMap {
```
adracus: `getConfigMap` -> `mkConfigMap`

pkg/controller/controlplane/utils.go
```diff
@@ -33,8 +36,42 @@ func DNSNamesForService(name, namespace string) []string {
 	}
 }
 
-// MergeSecretMaps merges the 2 given secret maps.
-func MergeSecretMaps(a, b map[string]*corev1.Secret) map[string]*corev1.Secret {
+// ComputeChecksums computes and returns the checksums for the given secrets and configmaps,
+// as well as the secrets and configmaps with the given names that are fetched from the cluster.
+func ComputeChecksums(
```
adracus: I'd rather still have the old `Merge...` methods and the old `ComputeChecksums` method. This one has too many responsibilities at once. Also, `Get`ting secrets and configmaps one by one by name is not really well performing and hides away the fact that one should rather apply smarter labels to all relevant secrets in order to fetch them in one request.
So my suggestion:
* Keep the old methods
* Apply better labels to relevant secrets
* Maybe create a method `AddSecretsToSecretsMap(secrets map[string]*corev1.Secret, newSecrets []*corev1.Secret)` (same for `ConfigMap`) that simplify building up the secrets and configmap map.

controllers/provider-aws/pkg/controlplane/internal/helper_test.go
```diff
@@ -0,0 +1,227 @@
...
+		ctrl *gomock.Controller
+
+		cp = &extensionsv1alpha1.ControlPlane{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "control-plane",
+				Namespace: namespace,
+			},
+			Spec: extensionsv1alpha1.ControlPlaneSpec{
+				ProviderConfig: &runtime.RawExtension{
+					Raw: encode(&apisaws.ControlPlaneConfig{
```
adracus: Do you really have to use the `Raw` values here? Can't those `ComputeValues` functions take the unmarshalled values and operate on these? This would also fit the SRP.

controllers/provider-aws/pkg/controlplane/internal/helper.go
```diff
@@ -0,0 +1,235 @@
...
+	return map[string]interface{}{
+		"cloudProviderConfig": getCloudProviderConfig(infraStatus.VPC.ID, subnetID, zone, cp.Namespace),
+	}, nil
+}
+
+// getControlPlaneChartValues collects and returns the control plane chart values.
+func getControlPlaneChartValues(decoder runtime.Decoder, cp *extensionsv1alpha1.ControlPlane, cluster *extensionscontroller.Cluster, checksums map[string]string) (map[string]interface{}, error) {
+	// Decode providerConfig
+	cpConfig := &apisaws.ControlPlaneConfig{}
+	if _, _, err := decoder.Decode(cp.Spec.ProviderConfig.Raw, nil, cpConfig); err != nil {
```
adracus: Write a `ControlPlaneConfigFromProviderConfig` method using a 'constant' scheme in a separate file just as we did it for the gcp infrastructure controller. Then this function only has a single responsibility.

controllers/provider-aws/pkg/controlplane/internal/helper.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	// Collect config chart values
+	return map[string]interface{}{
+		"cloudProviderConfig": getCloudProviderConfig(infraStatus.VPC.ID, subnetID, zone, cp.Namespace),
+	}, nil
+}
+
+// getControlPlaneChartValues collects and returns the control plane chart values.
+func getControlPlaneChartValues(decoder runtime.Decoder, cp *extensionsv1alpha1.ControlPlane, cluster *extensionscontroller.Cluster, checksums map[string]string) (map[string]interface{}, error) {
```
adracus: `getControlPlaneChartValues` -> `ComputeControlPlaneChartValues` in an `internal` package + add tests

controllers/provider-aws/pkg/controlplane/internal/helper.go
```diff
@@ -0,0 +1,235 @@
...
+			{Type: &corev1.Service{}, Name: "cloud-controller-manager"},
+			{Type: &appsv1.Deployment{}, Name: "cloud-controller-manager"},
+		},
+	}
+)
+
+// getConfigChartValues collects and returns the configuration chart values.
+func getConfigChartValues(decoder runtime.Decoder, cp *extensionsv1alpha1.ControlPlane, cluster *extensionscontroller.Cluster) (map[string]interface{}, error) {
+	// Decode providerConfig
+	cpConfig := &apisaws.ControlPlaneConfig{}
```
adracus: Same as below, write a decoding function with a constant scheme and make this method operate on the decoded variant.

controllers/provider-aws/pkg/controlplane/internal/helper.go
```diff
@@ -0,0 +1,235 @@
...
+		Images: []string{common.HyperkubeImageName},
+		Objects: []*chart.Object{
+			{Type: &corev1.Service{}, Name: "cloud-controller-manager"},
+			{Type: &appsv1.Deployment{}, Name: "cloud-controller-manager"},
+		},
+	}
+)
+
+// getConfigChartValues collects and returns the configuration chart values.
+func getConfigChartValues(decoder runtime.Decoder, cp *extensionsv1alpha1.ControlPlane, cluster *extensionscontroller.Cluster) (map[string]interface{}, error) {
```
adracus: `getConfigChartValues` -> `ComputeConfigChartValues` in an `internal` package + add tests.

controllers/provider-aws/pkg/controlplane/actuator.go
```diff
@@ -16,153 +16,37 @@ package controlplane
...
-	},
-	Objects: []*chart.Object{
-		{Type: &corev1.Service{}, Name: "cloud-controller-manager"},
-		{Type: &appsv1.Deployment{}, Name: "cloud-controller-manager"},
-	},
-}
-
 // NewActuator creates a new Actuator that acts upon and updates the status of ControlPlane resources.
-func NewActuator() controlplane.Actuator {
+func NewActuator(helper *internal.Helper) controlplane.Actuator {
```
adracus: Once splitting up the `Helper` is done, you don't need to embed this anymore.

adracus (changes requested):
Some changes, also you'll need to fix the merge conflicts.

controllers/provider-aws/pkg/controlplane/internal/helper.go
stoyanr: The `Helper` is a way to inject stuff in the `Actuator` that can be replaced by a mock implementation for testability.  If I replace it with functions, what do I inject so that it's still testable?

controllers/provider-aws/pkg/controlplane/internal/helper_test.go
stoyanr: The switch is a way to convert `obj` into a concrete type, so that `Unmarshal` works correctly. I know it looks strange, but it's the only way to write it - you need different cases for `x` to change its type, even though the 2 cases look the same. You can try it out if you have doubts.

controllers/provider-aws/pkg/imagevector/imagevector.go
stoyanr: This is not my code - it was like this before. I only changed the global variable to a function because reading a file from a hardcode path in an `init` function fails during unit tests (they are not executed at the root and the file is not found)

controllers/provider-aws/pkg/infrastructure/add.go
stoyanr: Again, this is not part of my change, I only changed it because unit tests were failing otherwise. Perhaps it makes sense, but I think it doesn't belong in this PR.

pkg/controller/controlplane/utils_test.go
stoyanr: Why? `mkSomething` is somewhat awkward naming in my opinion. It's not Go stylish, and I don't think it's used in either `gardener` or `gardener-extensions`. In general, if you suggest naming changes like this, could you point why do you think they should be made?

pkg/controller/controlplane/utils_test.go
stoyanr: See above.

pkg/controller/controlplane/utils.go
stoyanr: * The new method is exactly what `Actuators` would need to use in order to correctly build their checksums. I do believe a single method is very useful, otherwise its code will be copied for every provider. 
* In practice, it only reads 1 secret (cloudprovider) and 1 config map (cloud-provider-config) from the cluster, so performance considerations are not relevant. Adding labels etc. would only complicate code and would be a "premature optimization" at best. 
* The old methods are still there, I could just make them public if you prefer - but that would be increasing API surface prematurely.

controllers/provider-aws/pkg/controlplane/internal/helper_test.go
stoyanr: The `Raw` values are what the code under test decodes. Sure I could put there whatever and setup the decoder mocks to return concrete values, but I find encoding and decoding from JSON slightly more elegant, since the decoder mock setup in this case is always the same and very straightforward:

`decoder.EXPECT().Decode(cp.Spec.ProviderConfig.Raw, nil, &apisaws.ControlPlaneConfig{}).DoAndReturn(decode)`

Not sure what `ComputeValues` functions you refer to. If you mean `getConfigChartValues`, etc., these functions receive `ControlPlane` and decode it themselves, since the decoding itself is provider specific. I don't believe SRP is violated anywhere.

controllers/provider-aws/pkg/controlplane/internal/helper.go
stoyanr: Please see my answer to your first comment. What do I inject into `Actuator` in this case?

controllers/provider-aws/pkg/controlplane/internal/helper.go
stoyanr: See my comment above.

controllers/provider-aws/pkg/controlplane/actuator.go
stoyanr: See my comment above.

controllers/provider-aws/pkg/controlplane/internal/helper_test.go
```diff
@@ -0,0 +1,227 @@
...
+			// Call ControlPlaneChartValuesFunc and check the results
+			values, err := h.ControlPlaneChartValuesFunc(decoder, cp, cluster, checksums)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(values).To(Equal(expectedValues))
+		})
+	})
+
+})
+
+func encode(obj runtime.Object) []byte {
```
adracus: Error case missing - I know this is test code, but test code should be treated as production code as well. How about directly calling `Expect` in this function and calling it `mustEncode`.

controllers/provider-aws/pkg/controlplane/internal/helper_test.go
adracus: As said, this switch definitely has no value. You could just directly call `json.Unmarshal`.

controllers/provider-aws/pkg/imagevector/imagevector.go
adracus: You started to adapt the code thus you have to change it as well.

controllers/provider-aws/pkg/infrastructure/add.go
adracus: Again, you renamed it from `ImageVector` to `GetImageVector` - This has to be changed by you.

pkg/controller/controlplane/utils.go
adracus: The new method is what exactly your `controlplane` actuators need to do + it is heavily overloaded and non-performant. Testing all the edge cases of this single function is way too heavy duty. Also, if we split it up now, we can re-use this function easily in the future.

And even if it only reads one secret, if we in the future decide to add more of them, then we suddenly get performance issues due to the multiple `Get` requests - why build in performance pits if we can easily avoid them??

pkg/controller/controlplane/utils_test.go
adracus: This is Go and not Java. We are a community Go project, thus we should also strive to have code similar to the community for everyone to easily start working on it. `get` always indicates a more or less 'remote' call to retrieve something from somewhere. In this case, you're just building a `Secret`, thus `get` is a bad name.

pkg/controller/controlplane/utils_test.go
adracus: See above.

controllers/provider-aws/pkg/controlplane/internal/helper.go
adracus: I currently don't see any comment above. Still, as said, this should be properly tested, most probably in an `internal` package to avoid exposure.

adracus (changes requested):

controllers/provider-aws/pkg/controlplane/internal/helper.go
stoyanr: Injection isn't Java-only, it's the most meaningful way in any language, including Go, to make your code testable. At Uber, where I spent some time before coming back to SAP, injection is pervasive in Go code, they even have a special framework for this. You can check my talk at a Go meetup https://www.youtube.com/watch?v=LDGKQY8WJEM and the resources mentioned there.
Using this approach all Go code I wrote at Uber was covered by tests 80% or more. The SEN project I am currently working on uses the same approach and has a similar level of test coverage (without the framework, it's not that important but the approach itself is). 
I don't think that the suggested approach is the best one. I can do whatever you ask but we are missing the chance to do better.

controllers/provider-aws/pkg/imagevector/imagevector.go
stoyanr: I **have to** change it? No, I don't think so.

controllers/provider-aws/pkg/infrastructure/add.go
stoyanr: I don't think so - this is only an adaptation to make any tests work.

controllers/provider-aws/pkg/imagevector/imagevector.go
adracus: Ok, you should :wink:

controllers/provider-aws/pkg/controlplane/internal/helper.go
adracus: Yes, however having functions as seperated allows using them easily in other code without having to instantiate your actuator / 'pollute' one's struct with the embedded one.

stoyanr:
Abandoning - tired of this, sorry.

PR #44, opened 2019-04-08
Disable update engine service for AlicloudCoreOS

jia-jerry:
Disable update engine service for AlicloudCoreOS


rfranzke:
Thanks. FYI: This will trigger a rolling update of all Alicloud worker nodes of all clusters as soon as this is rolled out (because the bootstrap config changes).

PR #47, opened 2019-04-11
Add certificate service extension controller

timuthy:
This PR adds an implementation of the Certificate Service Extension. It manages components like `Cert-Manager` and `Cert-Broker` which provide certificate services for Shoot clusters.

It's mainly comprised of these controllers:
- Lifecycle-Controller: manages Cert-Manager instance in the Seed.
- Certificate-Service-Controller: manages Cert-Broker instances in the Seed + RBAC in Shoots.

Fixes parts of https://github.com/gardener/gardener/issues/730

**Release note**:

```improvement operator
Gardener Extensions got a new extension controller which manages certificate service components.
```


Makefile
```diff
@@ -21,6 +21,7 @@ LD_FLAGS                    := "-w -X github.com/gardener/gardener-extensions/pk
 VERIFY                      := true
 LEADER_ELECTION             := false
 IGNORE_OPERATION_ANNOTATION := false
+CERTIFICATE_SERVICE_CONFIG  := ./controllers/extension-certificate-service/example/service-config.yaml
```
rfranzke: I would call the file just "config.yaml" (when I first read I was confused what "service config" might be, only the `CERTIFICATE_SERVICE_CONFIG` gave me a hint ;-)).

controllers/extension-certificate-service/README.md
```diff
@@ -0,0 +1,105 @@
+# [Gardener Extension for certificate services](https://gardener.cloud)
+
+[![Go Report Card](https://goreportcard.com/badge/github.com/gardener/gardener-extensions/controllers/extension-certificate-controller)](https://goreportcard.com/report/github.com/gardener/gardener-extensions/controllers/extension-certificate-controller)
```
rfranzke:
```suggestion
[![Go Report Card](https://goreportcard.com/badge/github.com/gardener/gardener-extensions/controllers/extension-certificate-service)](https://goreportcard.com/report/github.com/gardener/gardener-extensions/controllers/extension-certificate-service)
```

controllers/extension-certificate-service/charts/cert-broker-rbac/templates/rbac.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: gardener.cloud:system:cert-broker
```
rfranzke: In gardener, this is called `garden.sapcloud.io:system:cert-broker`. Should we keep the name for now?

controllers/extension-certificate-service/charts/extension-certificate-service/templates/cert-config.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1	
+kind: Secret	
+metadata:	
+  name: extension-certificate-service
+  namespace: {{ .Release.Namespace }}	
+data:	
+  config.yaml: {{ include "certfonfig" . | b64enc | trim }}
```
rfranzke:
```suggestion
  config.yaml: {{ include "certconfig" . | b64enc | trim }}
```

controllers/extension-certificate-service/charts/extension-certificate-service/templates/_helpers.tpl
```diff
@@ -0,0 +1,51 @@
+{{- define "certfonfig" -}}
```
rfranzke:
```suggestion
{{- define "certconfig" -}}
```

controllers/extension-certificate-service/charts/extension-certificate-service/templates/_helpers.tpl
```diff
@@ -0,0 +1,51 @@
...
+---
+apiVersion: certificateservice.extensions.gardener.cloud/v1alpha1
+kind: CertificateServiceConfiguration
```
rfranzke:
```suggestion
kind: Configuration
```

because you already are in the `certificateservice` api group

controllers/extension-certificate-service/charts/utils-templates/templates/_resources.tpl
```diff
@@ -0,0 +1,12 @@
+{{/*
```
rfranzke: Do you use that? Otherwise, rm this file

controllers/extension-certificate-service/example/crd-extension.yaml
```diff
@@ -0,0 +1,51 @@
...
+    served: true
+    storage: true
+  version: v1alpha1
+  scope: Namespaced
+  names:
+    plural: extensions
+    singular: extension
+    kind: Extension
+    shortNames:
+    - gener
```
rfranzke:
```suggestion
    - ext
```

controllers/extension-certificate-service/example/crd-extension.yaml
```diff
@@ -0,0 +1,51 @@
...
+  additionalPrinterColumns:
```
rfranzke: Let's also add the creation timestamp here

controllers/extension-certificate-service/pkg/controller/certservice/actuator.go
```diff
@@ -0,0 +1,335 @@
...
+	client client.Client
+	config *rest.Config
+
+	certServiceConfig configv1alpha1.CertificateServiceConfiguration
+
+	logger logr.Logger
+}
+
+// CreateOrUpdate the Extension resource.
+func (a *actuator) CreateOrUpdate(ctx context.Context, ex *extensionsv1alpha1.Extension) error {
```
rfranzke: Why not

```suggestion
func (a *actuator) Reconcile(ctx context.Context, ex *extensionsv1alpha1.Extension) error {
```

controllers/extension-certificate-service/pkg/controller/certservice/actuator.go
```diff
@@ -0,0 +1,335 @@
...
+	namespace := ex.GetNamespace()
+	cluster, err := controller.GetCluster(ctx, a.client, namespace)
+	if err != nil {
+		return err
+	}
+
+	if err := a.createCertBroker(ctx, cluster.Shoot, namespace); err != nil {
```
rfranzke: Shouldn't this be the last step (first creating all requirements (kubeconfig, rbac)), and when everything worked we finally deploy the component?

controllers/extension-certificate-service/pkg/controller/certservice/actuator.go
```diff
@@ -0,0 +1,335 @@
...
+	if err != nil {
+		return fmt.Errorf("Error creating Kubeconfig to delete RBAC manifests: %v", err)
+	}
+
+	if err := a.deleteRBAC(ctx, kubecfg); err != nil {
+		return err
+	}
+
+	if err := a.client.Delete(ctx, secret); err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("error cleaning up secret for rbac manager: %v", err)
```
rfranzke:
```suggestion
		return fmt.Errorf("error cleaning up secret for cert-broker: %v", err)
```

controllers/extension-certificate-service/pkg/controller/certservice/actuator.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+// Delete the Extension resource.
+func (a *actuator) Delete(ctx context.Context, ex *extensionsv1alpha1.Extension) error {
+	namespace := ex.GetNamespace()
+	if err := a.deleteCertBroker(ctx, namespace); err != nil {
+		return err
+	}
+
+	secret, err := a.createKubeconfigForRBAC(ctx, namespace)
```
rfranzke: For the deletion you only need the name/namespace, so maybe add this information as a `const` variable somewhere that you can access here to delete the secret. Otherwise you would unnecessarily generate a certificate each and every time.

controllers/extension-certificate-service/pkg/controller/certservice/actuator.go
```diff
@@ -0,0 +1,335 @@
...
+	)
+}
+
+func (a *actuator) deleteCertBroker(ctx context.Context, namespace string) error {
+	chartRenderer, err := chartrenderer.NewForConfig(a.config)
+	if err != nil {
+		return err
+	}
+	configValues := make(map[string]interface{})
+	release, err := chartRenderer.Render(
```
rfranzke: Why not using [already existing utilities](https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-aws/pkg/controller/controlplane/actuator.go#L101-L152) instead of re-implementing some parts here?

controllers/extension-certificate-service/pkg/controller/certservice/actuator.go
```diff
@@ -0,0 +1,335 @@
...
+
+		if err := targetClient.Delete(ctx, &obj); err != nil && !apierrors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (a *actuator) createKubeconfigForRBAC(ctx context.Context, namespace string) (*corev1.Secret, error) {
```
rfranzke: As discussed: You can use the `gardener` secret/kubeconfig to do these things (https://github.com/gardener/gardener/blob/master/pkg/apis/core/v1alpha1/types_constants.go#L38)

pkg/controller/extension/actuator.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"context"
+
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+)
+
+// Actuator acts upon Extension resources.
+type Actuator interface {
+	// CreateOrUpdate the Extension resource.
+	CreateOrUpdate(ctx context.Context, ex *extensionsv1alpha1.Extension) error
```
rfranzke:
```suggestion
	Reconcile(ctx context.Context, ex *extensionsv1alpha1.Extension) error
```

pkg/util/shoot.go
```diff
@@ -0,0 +1,107 @@
...
+		secret.Annotations[caChecksumAnnotation] = computedChecksum
+		return nil
+	})
+}
+
+// KubeAPIServerServiceDNS returns a domain name which can be used to contact
+// the Kube-Apiserver deployment of a Shoot within the Seed cluster.
+// e.g. kube-apiserver.shoot--project--prod.svc.cluster.local.
+func kubeAPIServerServiceDNS(namespace string) string {
+	return fmt.Sprintf("%s.%s.%s", common.KubeAPIServerDeploymentName, namespace, "svc.cluster.local.")
```
rfranzke: You don't know that the internal domain is `cluster.local`. Just use `kube-apiserver.<ns>`

rfranzke (changes requested):

controllers/extension-certificate-service/pkg/controller/certservice/actuator.go
timuthy: You are absolutely right! Thank you

controllers/extension-certificate-service/pkg/controller/certservice/actuator.go
timuthy: Since we can use the `gardener` secret, we can skip the deletion anyway.

controllers/extension-certificate-service/pkg/controller/certservice/actuator.go
timuthy: From my point of view the referenced utility isn't quite useful for the use case here.
It needs to know the `Objects    []*Object` you want to remove (see [Delete implementation](https://github.com/gardener/gardener-extensions/blob/master/vendor/github.com/gardener/gardener/pkg/utils/chart/chart.go#L117)), i.e. we could even use `client.Delete` directly.

OTOH, the coding below is very generic and can deal with future changes in the chart (e.g. manifest is added / removed).

rfranzke: > OTOH, the coding below is very generic and can deal with future changes in the chart (e.g. manifest is added / removed).

I guess the coding below does not work in all cases. As an example, let's take a Helm chart that only deploys certain resources if a special flag in its `values.yaml` is set (we have a lot of these charts in the gardener repo itself). Here, you are again rendering the chart, however, with the default values (`configValues=nil`). That means that you could miss such manifests and they would remain orphaned in the system. The only way to prevent this is to have the EXACT same values for the deletion that you used during the time you deployed the chart, and these values you are not having.

> It needs to know the Objects []*Object you want to remove (see Delete implementation), i.e. we could even use client.Delete directly.

The reason why it needs to know the objects is what I just described above.
In this case you know exactly what is part of the `cert-broker` chart (maybe there are no such manifests that are only deployed if a certain field is set). However, the chances are high that if such a thing is introduced this deletion here does not work anymore.

I suggest to either use the utility to ease the cleaning/deletion process, or to simply call `client.Delete` as you suggested. The way it is currently implemented tries to solve it generically but this does not work in all cases. The next argument would be that you know the `cert-broker` chart and no such thing exists, but then it's again easier/cleaner to simply call `client.Delete`.

timuthy:
@adracus @rfranzke thanks for your reviews. I've addressed your comments, PTAL.

controllers/extension-certificate-service/pkg/controller/certservice/add.go
```diff
@@ -0,0 +1,85 @@
...
+	return AddToManagerWithOptions(mgr, ControllerOptions, ServiceConfig)
+}
+
+// AddToManagerWithOptions adds a controller with the given Options to the given manager.
+// The opts.Reconciler is being set with a newly instantiated actuator.
+func AddToManagerWithOptions(mgr manager.Manager, opts controller.Options, config controllerconfig.Config) error {
+	var cl = mgr.GetClient()
+
+	// watchBuilder determines which resources should be watched and the reconciler to be used.
+	var watchBuilder controllerutil.WatcherBuilder
```
adracus: you could also directly instantiate your `watchBuilder` with the two functions, so you don't have to call `Register` twice.

controllers/extension-certificate-service/pkg/controller/certservice/internal/mapper_test.go
```diff
@@ -0,0 +1,253 @@
...
+					gomock.AssignableToTypeOf(&extensionsv1alpha1.ExtensionList{}),
+				).
+				DoAndReturn(func(_ context.Context, _ *client.ListOptions, actual *extensionsv1alpha1.ExtensionList) error {
+					*actual = geList
+					return nil
+				})
+
+			result := requestFunc(object)
+
+			Expect(result).To(HaveLen(1))
```
adracus: Consider the `ConsistsOf` matcher.

controllers/extension-certificate-service/pkg/controller/certservice/internal/mapper_test.go
```diff
@@ -0,0 +1,253 @@
...
+			Expect(result[0].Name).To(Equal(resourceName))
```
adracus: Same as above.

controllers/extension-certificate-service/pkg/controller/lifecycle/actuator.go
```diff
@@ -0,0 +1,126 @@
...
+	a.client = client
+	return nil
+}
+
+const certManagerName = "cert-manager"
+
+var certManagerChartPath = filepath.Join(utils.ChartsPath, certManagerName)
+
+func (a *actuator) DeployCertManager(ctx context.Context, config configv1alpha1.Configuration) error {
+	applier, err := kubernetes.NewChartApplierForConfig(a.config)
```
adracus: You can instantiate the chart applier only once during `InjectConfig` - saves you error handling + instantiation time.

controllers/extension-certificate-service/pkg/controller/lifecycle/add.go
```diff
@@ -0,0 +1,39 @@
...
+var (
+	// ServiceConfig contains configuration for the certificate service.
+	ServiceConfig controllerconfig.Config
+)
+
+// AddToManager adds a Certificate Service Lifecycle controller to the given Controller Manager.
+func AddToManager(mgr manager.Manager) error {
+	reconciler := NewReconciler(NewActuator(), ControllerName, ServiceConfig.Configuration)
+
+	if err := mgr.Add(reconciler); err != nil {
```
adracus: Just return the `err` immediately.

pkg/controller/utils.go
```diff
@@ -260,3 +261,11 @@ func exponentialBackoff(ctx context.Context, backoff wait.Backoff, condition wai
 
 	return wait.ErrWaitTimeout
 }
+
+// WatcherBuilder holds various functions which add watch controls to the passed Controller.
+type WatcherBuilder []func(controller.Controller) error
```
adracus: I'd rather name this `WatchBuilder`.
Also, can you provide a `NewWatchBuilder` function analogous to what is there for the `SchemeBuilder` / `AddToManagerBuilder`?
Finally, I'd also recommend implementing an `AddToController` function on the `WatcherBuilder` to ease adding all the watches to a controller using your `WatchBuilder`.

adracus (changes requested):
Some minor changes, otherwise looks really good!

controllers/extension-certificate-service/README.md
```diff
@@ -0,0 +1,103 @@
...
+## Components
+Certificate services for Shoot clusters are provided by the composition of two components:
+
+- [Cert-Manager](https://github.com/jetstack/cert-manager)
+- [Cert-Broker](https://github.com/gardener/cert-broker)
+
+## Configuration
+Example configuration which is reconciled by this extension controller:
+```yaml
+apiVersion: certificateservice.extensions.gardener.cloud/v1alpha1
```
rfranzke: WDYT about naming it `certificate-service.extensions.gardener.cloud/v1alpha1` for better readability?

controllers/extension-certificate-service/charts/cert-broker-rbac/templates/rbac.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: garden.sapcloud.io:system:cert-broker
```
rfranzke: Hm, I know that I previously suggested to rename it back to `garden.sapcloud.io`, but thinking again about it: When this extension is introduced in Gardener then this old role will be deleted anyways, because it's currently managed by the kube-addon-manager. So, we can also choose a new name here, right? WDYT about `extensions.gardener.cloud:extension-certificate-service:cert-broker`? (E.g., the AWS provider will create a ClusterRole for MCM named `extensions.gardener.cloud:provider-aws:machine-controller-manager`).
(Same for all other `garden.sapcloud.io` names.)

controllers/extension-certificate-service/charts/cert-broker-rbac/templates/rbac.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
rfranzke: Do you want to move these charts into a subfolder (like the others), e.g. `charts/internal/cert-broker`?

controllers/extension-certificate-service/charts/cert-broker/values.yaml
```diff
@@ -0,0 +1,31 @@
+# Default values for cert-broker.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicas: 1
+
+images:
+  cert-broker: eu.gcr.io/gardener-project/gardener/cert-broker:0.2.0
```
rfranzke: I'd not put a concrete image here to avoid confusion where the actually used tag is coming from (should come from image vector), similar to https://github.com/gardener/gardener/blob/master/charts/seed-controlplane/charts/cert-broker/values.yaml#L8.

controllers/extension-certificate-service/charts/extension-certificate-service/templates/rbac.yaml
```diff
@@ -0,0 +1,107 @@
...
+  - events
+  verbs:
+  - create
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: gardener-extension-generic-certificate-service
+  labels:
+    app.kubernetes.io/name: gardener-extension-generic-certificate-service
```
rfranzke: Remove the `generic-` infix everywhere

controllers/extension-certificate-service/pkg/apis/config/v1alpha1/validation.go
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: By convention, validation is a separate package. Move it to `pkg/config/validation/validation.go` and only act on the internal version (don't validate on a versioned API).

controllers/extension-certificate-service/pkg/apis/config/v1alpha1/validation_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"testing"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	//. "github.com/onsi/ginkgo/extensions/table"
```
rfranzke: ?

controllers/extension-certificate-service/pkg/controller/config/config.go
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controller
+
+import "github.com/gardener/gardener-extensions/controllers/extension-certificate-service/pkg/apis/config/v1alpha1"
+
+// Config contains configuration for the certificate service.
+type Config struct {
```
rfranzke: Why do you have this?

pkg/controller/extension/reconciler.go
```diff
@@ -0,0 +1,248 @@
...
+	r.logger.Info("Extension resource deletion successful, removing finalizer.", "extension", ex.Name, "namespace", ex.Namespace)
+	if err := controllerutil.DeleteFinalizer(ctx, r.client, r.finalizerName, ex); err != nil {
+		r.logger.Error(err, "Error removing finalizer from Extension resource", "extension", ex.Name, "namespace", ex.Namespace)
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{}, nil
+}
+
+func (r *reconciler) updateStatusProcessing(ctx context.Context, ex *extensionsv1alpha1.Extension, lastOperationType gardencorev1alpha1.LastOperationType, description string) error {
+	ex.Status.LastOperation = extensionscontroller.LastOperation(lastOperationType, gardencorev1alpha1.LastOperationStateProcessing, 1, description)
```
rfranzke: I guess we want to use `TryUpdateStatus` here as well for all the status updates. The actuator might perform long-running operations.

controllers/extension-certificate-service/cmd/main.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/gardener/gardener-extensions/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+func main() {
+	log.SetLogger(log.ZapLogger(false))
+
+	cmd := app.NewServiceControllerCommand(controller.SetupSignalHandlerContext())
+	if err := cmd.Execute(); err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
```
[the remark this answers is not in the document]
timuthy: Okay, I also changed this for the other controllers.

controllers/extension-certificate-service/pkg/controller/certservice/actuator.go
```diff
@@ -0,0 +1,316 @@
...
+		&rbacv1.RoleBinding{
+			ObjectMeta: meta,
+		},
+		&rbacv1.ClusterRoleBinding{
+			ObjectMeta: meta,
+		},
+	}
+
+	a.logger.Info("Component is being deleted", "component", "cert-broker", "namespace", namespace)
+	for _, obj := range objects {
```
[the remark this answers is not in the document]
timuthy: Agreed. Let's keep it for now and use https://github.com/gardener/gardener/blob/65c4f5f9603b2d84dc6cc697f6f041f981dee072/pkg/utils/flow/taskfn.go#L153
as soon as we bumped to a new Gardener version, so we don't have to duplicate the logic here.

controllers/extension-certificate-service/pkg/controller/config/config.go
timuthy: It's needed to stick to the controller construction process we have on the top: https://github.com/gardener/gardener-extensions/blob/ea9fca24d90bd0f6f6c5917755c55780b39a6780/controllers/extension-certificate-service/cmd/app/app.go#L69

`Complete()` -> `Completed()` -> `Apply()` pattern.

Concretely, to `Apply` the parsed configuration from the command line to the configuration used by the reconcilers.

adracus (dismissed):
\lgtm as soon as the suggestions of @rfranzke have been applied.

controllers/extension-certificate-service/README.md
mvladev: If I recall correctly there was a bug in code generator and having a dash in the last DNS label causes an error

controllers/extension-certificate-service/pkg/apis/config/v1alpha1/types.go
```diff
@@ -0,0 +1,127 @@
...
+	ResourceNamespace string          `json:"resourceNamespace"`
+	ACME              ACME            `json:"acme"`
+	Providers         DNSProviders    `json:"providers"`
+}
+
+// ACME holds information about the ACME issuer used for the certificate service.
+type ACME struct {
+	Email      string  `json:"email"`
+	Server     string  `json:"server"`
+	PrivateKey *string `json:"privateKey,omitempty"`
```
mvladev: missing `// +optional` tag

controllers/extension-certificate-service/pkg/controller/certservice/actuator.go
```diff
@@ -0,0 +1,316 @@
...
+package certservice
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"k8s.io/apimachinery/pkg/runtime"
+
```
mvladev: there are multiple empty lines in imports

controllers/extension-certificate-service/charts/cert-broker-rbac/templates/rbac.yaml
timuthy: Sounds good!

controllers/extension-certificate-service/README.md
timuthy: I tried it out, the generation also works with the dash.

controllers/extension-certificate-service/pkg/controller/config/config.go
rfranzke: OK, thank you!

controllers/extension-certificate-service/pkg/apis/config/v1alpha1/validation.go
timuthy: Okay, I haven't added an internal version yet because I thought it'd be too much overhead for an internal configuration resource.

Anyway, I can add an internal version, validate on this but would also suggest to let the reconcilers/actuators rather work with the internal version (defaulted, validated) than the external `v1alpha1`. WDYT?

rfranzke: Usually, you also have internal versions for your configuration, e.g.
- https://github.com/gardener/gardener/blob/master/pkg/controllermanager/apis/config/types.go
- https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/config/types.go
- https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/apis/config/types.go
- ...

This yields the benefit that the whole API machinery can be used with this resource as well. 

> internal configuration resource.
End-users rely and configure it, so it's also external (that's why you have the `v1alpha1` version in the first place).

> but would also suggest to let the reconcilers/actuators rather work with the internal version (defaulted, validated)

Yes, definitely, this makes sense. 👍

timuthy:
Thanks for your review guys, PTAL.

controllers/extension-certificate-service/pkg/controller/certservice/actuator.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	a.logger.Info("Component is being deleted", "component", "cert-broker", "namespace", namespace)
+	for _, obj := range objects {
+		if err := a.client.Delete(ctx, obj); err != nil && !apierrors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	secret, err := a.createKubeconfigForCertManager(ctx, namespace)
```
rfranzke: Why not exporting the name of this kubeconfig as a constant somewhere and simply add another `&corev1.Secret` object to your list of objects above?

pkg/util/shoot.go
```diff
@@ -0,0 +1,121 @@
...
+		},
+	}
+
+	controlPlane, err := config.GenerateControlPlane()
+	if err != nil {
+		return nil, fmt.Errorf("error creating kubeconfig: %v", err)
+	}
+
+	return &secret, kubernetes.CreateOrUpdate(ctx, client, &secret, func() error {
+		secret.ObjectMeta.Name = certificateConfig.Name
```
rfranzke: Define the objectmeta in line 58 (this does not never change)

controllers/extension-certificate-service/pkg/controller/certservice/internal/mapper.go
```diff
@@ -0,0 +1,78 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	controllerutil "github.com/gardener/gardener-extensions/pkg/controller"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+)
+
+// ObjectNameToExtensionTypeMapper returns a `ToRequestsFunc` that gets the Extension resource of type certificate-service
+// in the namespace that is named after the incoming object's name.
+func ObjectNameToExtensionTypeMapper(cl client.Client, extensionType string) handler.ToRequestsFunc {
```
rfranzke: Can you move these two functions here outside of the cert-service? It is helpful for other extensions as well, right?

controllers/extension-certificate-service/pkg/controller/certservice/actuator.go
```diff
@@ -0,0 +1,306 @@
...
+	a.logger.Info("Manifests are being applied")
+	if err := applier.ApplyChart(ctx, rbacChartPath, "", "", nil, nil); err != nil {
+		return fmt.Errorf("error applying manifests: %v", err)
+	}
+
+	return nil
+}
+
+func (a *actuator) deleteRBAC(ctx context.Context, config *rest.Config) error {
+	renderer, err := chartrenderer.NewForConfig(config)
```
rfranzke: I still like the approach you chose in `deleteCertBroker` much more... :-)

pkg/controller/extension/reconciler.go
```diff
@@ -0,0 +1,256 @@
...
+		result reconcile.Result
+		err    error
+	)
+
+	if ex.DeletionTimestamp != nil {
+		return r.delete(r.ctx, ex)
+	}
+
+	result, err = r.reconcile(r.ctx, ex)
+	if err == nil {
```
rfranzke: Can you first handle the `err != nil` case?

controllers/extension-certificate-service/pkg/apis/config/v1alpha1/register.go
```diff
@@ -0,0 +1,58 @@
...
+
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// GroupName is the group name use in this package
+const GroupName = "certificate-service.extensions.gardener.cloud"
```
rfranzke:
```suggestion
const GroupName = "config.certificate-service.extensions.gardener.cloud"
```

by convention this group is contains `config.`:
- https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/apis/config/register.go#L25
- https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/config/register.go#L25
- https://github.com/gardener/gardener/blob/master/pkg/controllermanager/apis/config/register.go#L23

timuthy: Okay, but to follow the convention shouldn't it rather be `certificate-service.extensions.config.gardener.cloud` ?

rfranzke: Not every component has a config API. The scope of the cert-service extension is `certificate-service.extensions.gardener.cloud`, so I guess the cleanest way is to add `config` to the beginning (I'm doing it similarly with the worker controller: `config.aws.provider.extensions.gardener.cloud`).

timuthy: As discussed, we're going to go with `certificate-service.extensions.config.gardener.cloud` since we already use `config.gardener.cloud` for the `Gardener-Controller-Manager` and it is aligned with the config structure of k8s.io.

controllers/extension-certificate-service/example/config.yaml
```diff
@@ -0,0 +1,34 @@
+---
```
rfranzke: Can you please also add an example `Extension` resource?

rfranzke (approved):
/lgtm

PR #63, opened 2019-04-23
Add kubelet config to AWS controlplane webhook

stoyanr:
Adds additional logic for mutating kubelet configuration in `OperatingSystemConfig` resources to the AWS controlplane webhook.

Reusable types and functions are added to `pkg/webhook/controlplane` so that they could be used by other controlplane webhooks as well.

**Release note**:

```improvement operator
Add kubelet config to AWS controlplane webhook
```


controllers/provider-aws/pkg/webhook/controlplane/mutator.go
```diff
@@ -156,3 +168,82 @@ func ensureVolumes(ps *corev1.PodSpec) {
...
+	if opt := controlplane.UnitOptionWithSectionAndName(opts, "Service", "ExecStart"); opt != nil {
+		command := controlplane.DeserializeCommandLine(opt.Value)
+		command = ensureKubeletCommandLineArgs(command)
+		opt.Value = controlplane.SerializeCommandLine(command, 1, " \\\n    ")
+	}
+}
+
+func ensureKubeletCommandLineArgs(command []string) []string {
+	command = controlplane.EnsureStringWithPrefix(command, "--cloud-provider=", "aws")
+	command = controlplane.EnsureNoStringWithPrefix(command, "--provider-id=")
```
rfranzke: This is not needed I guess - Gardener won't add this flag.

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
```diff
@@ -156,3 +168,82 @@ func ensureVolumes(ps *corev1.PodSpec) {
...
+	command = controlplane.EnsureNoStringWithPrefix(command, "--enable-controller-attach-detach=")
```
rfranzke: This is not needed I guess - Gardener won't add this flag.

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
```diff
@@ -156,3 +168,82 @@ func ensureVolumes(ps *corev1.PodSpec) {
...
+	command = controlplane.EnsureNoStringWithPrefix(command, "--cloud-config=")
```
rfranzke: This is not needed I guess - Gardener won't add this flag.

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
```diff
@@ -156,3 +168,82 @@ func ensureVolumes(ps *corev1.PodSpec) {
...
+
+	return command
+}
+
+func ensureKubeletConfiguration(kubeletConfig *kubeletconfigv1beta1.KubeletConfiguration) {
+	// Make sure CSI-related feature gates are not enabled
```
rfranzke: I get why you added that, but I would remove this part. If the user enables the feature gates via the Shoot manifest then it's a configuration issue anyways and we can't help (probably other components would start failing as well). Generally, there are so many options that we allow users to configure (feature gates, runtime config, admission plugins, oidc, ...) and we cannot prevent them from harming themselves (it's simply too much).
Adding this here can maybe help with this one problem, however, it also suggests to try to do the same for other flags/configuration, etc., which is hard to keep up and maintain.

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
stoyanr: This is consistent with the behavior of this webhook with the 2 deployments. Here, as well as there, it makes sure that any potentially harmful flags / configuration bits are removed. It makes no assumptions what Gardener would do or would not do - wasn't this what we agreed?

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
stoyanr: See above, this one is related to CSI.

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
stoyanr: See above.

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
stoyanr: See above. All these things are related to CSI. This is consistent with the behavior of `ensureKubeAPIServerCommandLineArgs` which makes sure that `"--enable-admission-plugins=` contains `PersistentVolumeLabel` and `--disable-admission-plugins=` doesn't contain it. These are also related to CSI. Here, as well as there, no assumptions are made what Gardener would do or would not do, as we agreed. One difference here is that if these feature gates end up enabled, this is likely not going to hurt (they would just get unused if a CSI plugin is not present).

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
rfranzke: OK, I don't recall whether we talked about exactly this. It's definitely safer to have this logic. I was just worried that it would be hard to maintain with new Kubernetes versions, etc., because there are so many options that cannot be used in conjunction. We can keep it as it is now, and potentially revisit this in the future if we see that the complexity is increased. WDYT?

pkg/webhook/controlplane/kubeletconfigcodec.go
```diff
@@ -0,0 +1,120 @@
...
+	// Encode encodes the given *kubeletconfigv1beta1.KubeletConfiguration into a *extensionsv1alpha1.FileContentInline.
+	Encode(*kubeletconfigv1beta1.KubeletConfiguration, *extensionsv1alpha1.FileContentInline) error
+	// Decode decodes a *kubeletconfigv1beta1.KubeletConfiguration from the given *extensionsv1alpha1.FileContentInline.
+	Decode(*extensionsv1alpha1.FileContentInline, *kubeletconfigv1beta1.KubeletConfiguration) error
+}
+
+// NewKubeletConfigCodec creates an returns a new KubeletConfigCodec.
+func NewKubeletConfigCodec() KubeletConfigCodec {
+	// Create scheme
+	scheme := runtime.NewScheme()
```
adracus: Instantiate this `scheme` only once since it's always the same (e.g. in an `init` function using a `SchemeBuilder`, then you can also ensure using `utilruntime.Must` that the scheme construction worked).

pkg/webhook/controlplane/kubeletconfigcodec.go
```diff
@@ -0,0 +1,120 @@
...
+
+type kubeletConfigCodec struct {
+	codec runtime.Codec
+}
+
+// Encode encodes the given *kubeletconfigv1beta1.KubeletConfiguration into a *extensionsv1alpha1.FileContentInline.
+func (c *kubeletConfigCodec) Encode(kubeletConfig *kubeletconfigv1beta1.KubeletConfiguration, fci *extensionsv1alpha1.FileContentInline) error {
+	// Encode kubelet configuration to YAML
+	var buffer bytes.Buffer
+	if err := c.codec.Encode(kubeletConfig, &buffer); err != nil {
```
adracus: there's `runtime.Encode` for encoding to `[]byte`.

pkg/webhook/controlplane/kubeletconfigcodec.go
```diff
@@ -0,0 +1,120 @@
...
+			return errors.Wrap(err, "could not encode kubelet config file content data")
+		}
+	}
+
+	fci.Data = string(data)
+	return nil
+}
+
+// Decode decodes a *kubeletconfigv1beta1.KubeletConfiguration from the given *extensionsv1alpha1.FileContentInline.
+func (c *kubeletConfigCodec) Decode(fci *extensionsv1alpha1.FileContentInline, kubeletConfig *kubeletconfigv1beta1.KubeletConfiguration) error {
```
adracus: Can we factor out the part that decodes the `FileContentInline` into a method like `DecodeFileContentInline`? Then we could also use the same in the `coreos-alicloud` controller.

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
stoyanr: Actually, I do think that without any assumptions about Gardener behavior the complexity of webhooks may explode. I think we should document what assumptions can be made in a README somewhere. The README should state what Gardener **will** do, what it **will not** do, and what it **may or may not** do (it could be safer not to make definite promises about some things). Such a README could play the role of a more explicit contract between Gardener and webhooks. This of course means that Gardener should try to abide by it in the future, breaking the contract only if absolutely necessary.

pkg/webhook/controlplane/utils.go
```diff
@@ -18,9 +18,33 @@ import (
...
+	"github.com/coreos/go-systemd/unit"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 )
 
+var cmdLineSpace = [256]uint8{'\t': 1, '\n': 1, '\v': 1, '\f': 1, '\r': 1, ' ': 1, '\\': 1}
+
+// DeserializeCommandLine deserializes the given string to a slice of command line elements by splitting it
+// on white space and the "\" character.
+func DeserializeCommandLine(s string) []string {
```
adracus: Why not split by a regex? For whitespace you could use `\s`, which makes the `var` above unnecessary.

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
stoyanr: Removed as initially suggested.

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
stoyanr: Removed as initially suggested.

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
stoyanr: Removed as initially suggested.

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
stoyanr: Keeping these even though they should probably be removed, so that the webhook code could serve as an example of modifying the `KubeletConfiguration` YAML file. Added a TODO to perhaps remove them later.

pkg/webhook/controlplane/kubeletconfigcodec.go
stoyanr: Done.

pkg/webhook/controlplane/kubeletconfigcodec.go
stoyanr: Using `runtime.Encode` and `runtime.Decode` now and changed the interface to return results as return values instead of modifying parameters.

controllers/provider-aws/pkg/webhook/controlplane/mutator.go
rfranzke: Agreed 👍

pkg/webhook/controlplane/kubeletconfigcodec.go
stoyanr: Factored out `FileContentInlineCodec` with `Encode` and `Decode`, still in `pkg/webhook/controlplane`. Should I move it somewhere else, e.g. `pkg/controller`?

pkg/webhook/controlplane/utils.go
stoyanr: Good idea, done.

PR #64, opened 2019-04-24
Create GCP infrastructure tests

adracus:
Create GCP infrastructure actuator tests.

**Release note**:

```improvement operator
NONE
```


stoyanr:
The proposed design captures the 4 main dependencies of the `Actuator` type into 4 global variables `NewGCPClientFromServiceAccount`, `NewTerraformer`, `CleanupKubernetesCloudArtifacts`, and `TerraformerDefaultInitializer` that are initialized with various functions. These global variables are then replaced in the test by mock implementations of interfaces defined in `pkg/internal/mock/infrastructure`  using a special reflection-based `testutil.WithVars` tool.

Besides the `Actuator` tests, there are also tests for individual functions in `infrastructure.go` and `terraform.go`. Note however that the `Actuator` tests also test the functions in `terraform.go` (easy to see if you run only these tests with coverage), since these functions are not replaced by the mechanism described above, so tests in `terraform_test.go` are somewhat redundant.

I would call this approach "dependency replacement", since dependencies are not really injected, but they are nevertheless externalized in global variables to make them replaceable for testing. I see the following drawbacks of this approach:

* Global variables, and global state in general, are considered "bad" not just in Go, but in most programming languages. See my talk https://www.youtube.com/watch?v=LDGKQY8WJEM and especially the posts by influential members of the Go community mentioned in it, [A theory of modern Go](https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html) and [Go, without package scoped variables](https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables).
* The entire implementation is in the `internal/infrastructure` containing a few dozen exported variables, functions, and types. This complicates the package structure, exposes a large API surface (still visible to all code in the GCP controller), and makes future evolution somewhat problematic. It would be much better if all these details could stay internal to a package, and if there would be just a single package.
* Replacing dependencies for testing is only possible using an awkward reflection-based mechanism. Replacing dependencies for other purposes would be even more problematic.

I think that the above drawbacks could be addressed by refactoring the design to use dependency injection. What needs to be done is to capture the 4 dependencies in interfaces (in fact this is already done, see `pkg/internal/mock/infrastructure`, but could be done better), and then inject them explicitly into the `Actuator`.

- `NewGCPClientFromServiceAccount`: define an interface `GCPClientFactory.NewFromServiceAccount`, the function `gcpclient.NewFromServiceAccount` already provides an implementation (functions can implement interfaces in Go via an additional `Func` type, e.g. `GCPClientFactoryFunc`).
- `NewTerraformer`: define an interface `TerraformerFactory.New`, the function `internal.NewTerraformer` already provides an implementation.
- `CleanupKubernetesCloudArtifacts`: define an interface `KubernetesCloudArtifactsCleaner.Cleanup`, the function `cleanupKubernetesCloudArtifacts` already provides an implementation. This funciton should be exported of course (and perhaps moved to the `pkg/internal` package), but none of the other functions in `infrastructure.go` need to be, as the entire `Cleanup` functionality can be tested quite adequately without exporting individual functions.
- `TerraformerDefaultInitializer`: define an interface `TerraformerInitalizerFactory.New`, the function `terraformer.DefaultInitializer` already provides an implementation.

Example for `GCPClientFactory`:

```
type GCPClientFactory interface {
	NewFromServiceAccount(context.Context, []byte) (gcpclient.Interface, error)
}

type GCPClientFactoryFunc func(context.Context, []byte) (gcpclient.Interface, error)

func (f GCPClientFactoryFunc) NewFromServiceAccount(ctx context.Context, serviceAccount []byte) (gcpclient.Interface, error) {
	return f(ctx, serviceAccount)
}
```

The `Actuator` type and constructor then would look as follows:

```
type actuator struct {
	gcpClientFactory GCPClientFactory
	...
}

func NewActuator(gcpClientFactory GCPClientFactory, ...) infrastructure.Actuator {
	return &actuator{
		gcpClientFactory: gcpClientFactory,
		...
	}
}
```

The actual instantiation of `Actuator` would look as follows:

`NewActuator(GCPClientFactoryFunc(gcpclient.NewFromServiceAccount), ...)`

In the resulting design:
* There would be no global variables, so no "magic". All dependencies would be explicit and the resulting code would be easier to understand and reason about.
* There would be no need of the `pkg/internal/infrastructure` package at all. All code from this package could simply stay in `pkg/controller/infrastructure`. This package would contain very few exported types (the interface types mentioned above), methods (`Reconcile`, `Delete`), and functions (`NewActuator` and `CleanupKubernetesCloudArtifacts`, if it remains in this package). Everything else would remain unexported, and therefore easy to refactor and change as needed in the future.
* In tests, the mocks of the above 4 interfaces would be injected directly, without using any special tools. Injecting a different implementation for other purposes is straightforward as well.

PR #69, opened 2019-04-30
Add GCP controlplane controller

stoyanr:
Adds GCP extension controller for ControlPlane resources, using the existing AWS controller as a model.

* The AWS controller code was copied from #50, not the version in the master branch, since I regard the small changes to the design of the AWS controller done there as important.
* There is a lot of code duplication between the 2 controlplane controllers (>90%). As discussed, I think this should be eliminated, but it could also be done later in a new PR (definitely before adding a third controlplane controller).
* Tests are not added yet. I would prefer to add them after the code duplication has been eliminated (perhaps in the same PR), since otherwise there will be even more duplication and they would have to be heavily refactored as well.

**Release note**:

```improvement operator
Add GCP controlplane controller
```


controllers/provider-gcp/pkg/apis/gcp/types_controlplane.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ControlPlaneConfig contains configuration settings for the control plane.
+type ControlPlaneConfig struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Zone is the GCP zone
+	Zone string `json:"zone,omitempty"`
```
rfranzke:
```suggestion
	Zone string `json:"zone"`
```

controllers/provider-gcp/pkg/apis/gcp/types_controlplane.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	gardenv1beta1 "github.com/gardener/gardener/pkg/apis/garden/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ControlPlaneConfig contains configuration settings for the control plane.
+type ControlPlaneConfig struct {
+	metav1.TypeMeta `json:",inline"`
```
rfranzke: no `json` tags needed in the internal API

controllers/provider-gcp/pkg/apis/gcp/v1alpha1/types_controlplane.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ControlPlaneConfig contains configuration settings for the control plane.
+type ControlPlaneConfig struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Zone is the GCP zone
+	Zone string `json:"zone,omitempty"`
```
rfranzke:
```suggestion
	Zone string `json:"zone"`
```

controllers/provider-gcp/pkg/apis/gcp/types_controlplane.go
stoyanr: Done

controllers/provider-gcp/pkg/apis/gcp/types_controlplane.go
stoyanr: Done

PR #77, opened 2019-05-08
Disable controllers and webhooks via command line flags

stoyanr:
Adds possibility to disable controllers and webhooks via command line flags

* By default, all controllers and webhooks are enabled, in `Makefile` as well as in provider charts. Should we disable some of them for the time being, e.g. everything "controlplane"?

**Release note**:

```improvement operator
Disable controllers and webhooks via command line flags
```


pkg/controller/utils.go
```diff
@@ -90,26 +90,31 @@ func SetupSignalHandlerContext() context.Context {
...
 
 // AddToManagerBuilder aggregates various AddToManager functions.
-type AddToManagerBuilder []func(manager.Manager) error
+type AddToManagerBuilder map[string]func(manager.Manager) error
 
 // NewAddToManagerBuilder creates a new AddToManagerBuilder and registers the given functions.
-func NewAddToManagerBuilder(funcs ...func(manager.Manager) error) AddToManagerBuilder {
-	var builder AddToManagerBuilder
-	builder.Register(funcs...)
+func NewAddToManagerBuilder(params ...interface{}) AddToManagerBuilder {
```
adracus: Don't modify the `AddToManagerBuilder`: Its responsibility is to accumulate `func(manager.Manager) error` to make it possible to call them in sequence. If you don't want that a function is added, you should just not register it / have several `AddToManager` functions.

pkg/controller/cmd/options.go
```diff
@@ -138,6 +143,8 @@ type ManagerOptions struct {
 	LeaderElectionID string
 	// LeaderElectionNamespace is the namespace to do leader election in.
 	LeaderElectionNamespace string
+	// Disable is a comma-separated list of controllers and webhooks that should be disabled.
+	Disable string
```
adracus: `ManagerOptions` reflect all the parameters that are related to building up a `Manager`. `Disable` doesn't correspond to one and thus I'd recommend having this as a separate flag that is a `bool`.

controllers/provider-openstack/cmd/gardener-extension-provider-openstack/app/app.go
```diff
@@ -82,7 +82,7 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 			infraCtrlOpts.Completed().Apply(&openstackinfrastructure.DefaultAddOptions.Controller)
 			infraReconcileOpts.Completed().Apply(&openstackinfrastructure.DefaultAddOptions.IgnoreOperationAnnotation)
 
-			if err := openstackcontroller.AddToManager(mgr); err != nil {
+			if err := openstackcontroller.AddToManager(mgr, mgrOpts.Completed().Disabled); err != nil {
```
adracus: If you switch the `Disabled` to a standalone `bool` flag you could instead just say
```go
if !disabled {
  openstackcontroller.AddToManager(mgr)
}
```

pkg/webhook/utils.go
```diff
@@ -52,32 +52,37 @@ const (
...
+	for i := 0; i < len(params)-1; i += 2 {
+		name, f := params[i].(string), params[i+1].(func(manager.Manager) (webhook.Webhook, error))
+		a[name] = f
+	}
 }
 
 // AddToManager creates a webhook server adds all webhooks created by the AddToManager-functions of this builder to it.
 // It exits on the first error and returns it.
-func (a *AddToManagerBuilder) AddToManager(mgr manager.Manager, cfg *cmd.WebhookServerConfig) error {
+func (a *AddToManagerBuilder) AddToManager(mgr manager.Manager, cfg *cmd.WebhookServerConfig, disabled map[string]bool) error {
```
adracus: Same as for the 'regular' `AddToManagerBuilder`, I'd not consider this the responsibility of the `AddToManagerBuilder` and move the logic to disable something outside.

pkg/controller/utils.go
stoyanr: The point is that I wan't to modify the functions that get called via a command-line flag containing a comma-separated list of things to be disabled. So I turn the builder into a map and pass the keys to be excluded later. Do you see a better way to do this? If yes, could you please explain?

pkg/controller/cmd/options.go
stoyanr: I can put it in a separate struct but it needs to be a string, since it's "a comma-separated list of controllers and webhooks that should be disabled"
The idea is to use `--disable=controlplane-controller,controlplane-webhook` in the command line.

controllers/provider-openstack/cmd/gardener-extension-provider-openstack/app/app.go
stoyanr: `disabled` should be a map since there are multiple things to disable. I could use something like:

```go
if !disabled[controlplane.ControllerName] {
  awscontrolplane.AddToManager(mgr)
}
```

and so on for all other controllers (and in all `app.go` files), but I don't really like this code - the `AddToManagerBuilder` already provides a useful abstraction that can be used for this, as I did it.

pkg/webhook/utils.go
stoyanr: See above, please explain how exactly you think this could be done (and why it's better). IMO, `AddToManagerBuilder` is exactly the right abstraction to modify - it allows collecting all functions and then switching on / off which of them should be added based on a set of keys, exactly what I need.

pkg/controller/cmd/options.go
adracus: Then I'd recommend having another `Option` that takes this 'list' and parses it into individual boolean flags, i.e.
```go
type ControllerSwitches struct {
  DisableControlplaneController bool
  DisableControlplaneWebhook bool
}

type ControllerSwitchOptions struct {
  Disable string
  completed *ControllerSwitches
}
```
The `ControllerSwitchOptions` would then have the `AddFlags` as all the other options and can do meaningful parsing and once `Completed` return the parsed `ControllerSwitches` that can then be used to decide whether to add something to the `Manager` or not.

controllers/provider-openstack/cmd/gardener-extension-provider-openstack/app/app.go
adracus: I think with the `AddToManagerBuilder` its single responsibility is to accumulate `AddToManager` functions - if it now has this behavior of this 'string map' it's gonna be tightly coupled to command line parsing which it shouldn't since it knows nothing about that.

pkg/webhook/utils.go
adracus: Same comment as above, an `AddToManagerBuilder` should not be involved with some logic like that - it doesn't even know about what each function does and thus cannot make decisions on what to add and what not - the right entity being able to decide what goes in is a level higher.

pkg/controller/utils.go
adracus: I'd suggest the way I described in the comment below - have some dedicated entity that knows what each of these strings means and parse them into a struct and also error if some of these don't exist (see comment with `ControllerSwitchOptions`).

pkg/controller/cmd/options.go
stoyanr: This approach has the disadvantage that adding a new controller means adding a new switch, something that's unnecessary and should be avoided IMO. Furthermore, if it can be done this way, this doesn't help with the "builder" discussion, see other comments.

pkg/controller/utils.go
stoyanr: The `AddToManagerBuilder` is such a dedicated entity IMO. If I use a different entity, I would have to delete the `AddToManagerBuilder` since it won't be used any more. In this case, I don't see why not using the builder itself?

controllers/provider-openstack/cmd/gardener-extension-provider-openstack/app/app.go
stoyanr: The single responsibility that you see is preserved, just functions are now named which allows filtering. I don't see any tight coupling - the names to filter on are just passed, independently of where they are obtained (command line, env variable, anything else), there is no package level dependency, etc. We have a similar mechanism in SEN and no command line flags whatsoever.

pkg/webhook/utils.go
stoyanr: I agree, and it isn't - the names to filter on are just passed from outside. The builder itself doesn't do any decisions at all.

pkg/controller/utils.go
stoyanr: More importantly, I still don't understand what exactly do you suggest. Could you perhaps give an example? We need a mechanism to filter which `AddToManager` functions are added to the manager, and which are not. To avoid repeated ifs for every controller, I'd argue we need a map of names to bools somewhere. Error handling is not hindered, we could just as easily validate on names without having a number of named boolean variables (which will have to be changed every time a new controller is added / removed). Do you have a concrete design in mind? If yes, could you please describe it so that it's understood? If not, may I suggest not to reject a design that is perhaps good enough?

pkg/webhook/utils.go
adracus: I would tend to disagree - in the new code, the builder does make decisions in my opinion:
https://github.com/gardener/gardener-extensions/pull/77/files/a3ddf255d0354ff81901eb09159581935bc3ff40#diff-ff368195ba700fad1507e2a15dfb8b09R79
Here you can see that the code and also the behavior of the builder is changed and it now decides on a map with which it actually does not have any relation.

controllers/provider-openstack/cmd/gardener-extension-provider-openstack/app/app.go
adracus: Let's maybe look at the signature: Before, the responsibility was clear: It just accumulated all the functions and made it possible to do one `AddToManager` for all of these functions.
Now, it decides on a mapping and names which are not related to anything inside this code - Why should a string to boolean map matter for the `AddToManagerBuilder` whether it should add a controller or not? The only entity that knows to what a name is related is the one that does the flag parsing and thus it should be the one deciding which function to pass into an `AddToManagerBuilder`. In the end, you can probably reuse the `AddToManagerBuilder` but there is another entity that knows how to make the links. I'll provide a quick comment how I mean that.

pkg/controller/cmd/options.go
adracus: It's true that this has the disadvantage of adding more named variables, but the `ControllerSwitches` map could be improved to be initialized with a `map[string]func(manager.Manager) error` that maps from the controller's name to the `AddToManager` function of that particular controller. Then it can do proper flag parsing and even validate when there is a name in that list that does not exist (and fail early) and then build up an `AddToManager` function with the required controllers.

adracus:
What I'd recommend is cleanly separating the command line parsing from the `AddTo<X>Builder`s. It's also important that we fail early if somebody specifies a controller that doesn't exist (same goes e.g. for Kubernetes' API server if you configure unknown `AdmissionPolicy`).

We could still stick to the
```go
type ControllerSwitches struct {
  DisabledControllers []string
  nameToAddToManager map[string]func(manager.Manager) error

  addToManagerBuilder AddToManagerBuilder
}

func (c *ControllerSwitches) AddFlags(fs *pflag.FlagSet) {
  fs.StringSlice(&c.DisabledControllers, ...)
}
```
Then, during its `Complete`, it can validate that all of the given controller names are valid and build the `AddToManagerFunction` correctly:
```go
func (c *ControllerSwitches) Complete() error {
  disabled := sets.NewString()
  for _, disabledName := range c.DisabledControllers {
    if _, ok := c.nameToAddToManager[disabledName]; !ok {
      return fmt.Errorf("cannot disable unknown controller %q", disabledName)
    }
    disabled.Insert(disabledName)
  }
  disabled := sets.NewString(c.DisabledControllers)
  for name, addToManager := range c.nameToAddToManager {
    if !disabled.Has(name) {
      c.addToManagerBuilder.Register(addToManager)
    }
  }
  return nil
}
```

pkg/controller/utils.go
adracus: I just posted a comment that gives an example that decouples these concerns. The 'map of bools' is there as well (in a slightly changed form).

pkg/webhook/utils.go
stoyanr: The builder filters out functions based on a filter passed from the outside. The builder is passive (it just uses the filter) and agnostic to where this filter comes from. Honestly, I don't see a problem with this approach.

stoyanr:
> What I'd recommend is cleanly separating the command line parsing from the `AddTo<X>Builder`s. 

It's already the case.

> It's also important that we fail early if somebody specifies a controller that doesn't exist (same goes e.g. for Kubernetes' API server if you configure unknown `AdmissionPolicy`).

Fully agreed, but this doesn't justify a total redesign.

> ```go
>   nameToAddToManager map[string]func(manager.Manager) error
> ```

Here is the important thing: how do you initialize the map above? Right now, when adding a controller, you add an `AddToManager` function in the corresponding `controllers.go`. This is preserved in design proposed in the PR (additionally you add a name). With the approach suggested above however, you have to maintain a map between controller names and functions somewhere in a third place, and furthermore it's a package that deals with command line parsing. I don't agree that this is a good idea or that it decreases coupling - on the contrary, it increases coupling between command line parsing and controllers.

How about if tomorrow we decide to use an env variable instead of a command line flag? We would need to refactor this entire thing.

Same goes for webhooks. Also, controllers and webhooks would need 2 different maps (due to different function signatures), and 2 different `Complete()` implementations, quite a lot of code for something that can be solved quite elegantly by just making the functions accumulated by `AddToManagerBuilder` filterable.

> ```go 
> func (c *ControllerSwitches) AddFlags(fs *pflag.FlagSet) {
>   fs.StringSlice(&c.DisabledControllers, ...)
> }
> ```

Making it a string slice is a good idea.

I understand what you mean finally but I am still puzzled with respect to the advantages that you see.

pkg/webhook/utils.go
adracus: Problem is that the filter does not mean anything to the `AddToManagerBuilder`. It doesn't have any relation to the names, so why put some domain specific filtering logic in there? Also, as mentioned, with this approach it's possible to specify unknown controller names without causing an error although the tool should indeed error and fail early showing the user a misconfiguration.

controllers/provider-openstack/cmd/gardener-extension-provider-openstack/app/app.go
stoyanr: The responsibility is the same and it's as clear as before. Functions are now named so that they can be filtered. The exact origin of the filter is not important (and not being depended upon). It can come from the command line, or from another source. *The only entity that knows to what a name is related is the one that does the flag parsing* is true right now only incidentally because we are using command line flags for configuration in preference to other alternatives. "Filterability" of functions is not accidental, however, it's a "core feature" that we would need even if we don't use command line flags at all.

adracus:
The same question is 'How do you initialize your binding between controller names and functions?'. Previously you moved the logic to the builder although, as mentioned, it doesn't have any notion of what such a name may mean. In the command line building, we can easily setup the names as we wish and decouple the command line parsing from building up the correct controller.

> Fully agreed, but this doesn't justify a total redesign.
It's not a redesign, we're still using the same builder code to build up the `AddToManager` functions.
It's just an indirection on top that decides what goes in and what not which is coupled to some command line flags.

We *have* to validate the names right at the parsing time and thus this is the only time we can know what name relates to which function. This is why it's completely ok to initialize the map in the command line code we have, e.g. by having a constructor building the `ControllerSwitches` where you are able to pass in the map or by exposing a function on them where you can register the mapping of name to `AddToManager` function.

> I understand what you mean finally but I am still puzzled with respect to the advantages that you see.
We have several important advantages:
* The logic of the builders doesn't become couple with names it doesn't have any relation to
* We have one place where we define the names and the binding to the corresponding controller 
* We fail early if there is an unknown name
* We can chose to rebind the controller names without altering our core code This closely relates to your comment
 > on the contrary, it increases coupling between command line parsing and controllers.
-> Being able to rebind names without changing code at other places shows that we indeed don't have coupling ongoing.

controllers/provider-openstack/cmd/gardener-extension-provider-openstack/app/app.go
adracus: The names and the functions are coupled at a code level. As mentioned, we cannot rename functions anymore without touching core code. That should not happen.
And filterability is definitely one thing you'd like to put on top. The controller used to work perfectly before, you want to add a concern so you add a layer of indirection on top.

stoyanr:
See my comment above. The design in the PR is motivated by the idea that "filterability" should be regarded as a core feature of the mechanism that accumulates the functions to be added to the manager (`AddToManagerBuilder`), while command line flags are just an "accidental" way of configuring the filter. 

> * The logic of the builders doesn't become couple with names it doesn't have any relation to

As I stated above, I don't think that adding names to functions increases coupling in any way. It adds a *feature* - filterability, which I think is important and quite independent of command line flags or other configuration mechanisms.

> * We have one place where we define the names and the binding to the corresponding controller

I actually think it's a disadvantage - having it distributed in `controllers.go` is easier to understand and maintain and less "coupled" (no single place that has a dependency on every controller and webhook). It's not accidental that it was distributed like this in the first place, and I'd rather leave it as it is.

> * We fail early if there is an unknown name

I agree that this is important, but can be achieved in a number of different ways. The redesign that you are proposing is not mandatory for this.

> * We can chose to rebind the controller names without altering our core code

If by "rebind" you mean change, well, yes, but again I fail to see the strong benefit. I'd prefer to name a controller when it's added and (for clarity and consistency) to use the name to refer to the same controller everywhere - logging, configuration, command line. I would need a very strong reason **not** to do it, and it's hard for me to come up with such a reason. 

> Being able to rebind names without changing code at other places shows that we indeed don't have coupling ongoing.

Coupling has little to do with names (strings you use in code) and much more to do with dependencies between packages. With the approach you are suggesting, we'll end up with a single place depending on everything that has to be changed every time a new controller or webhook is added / removed. This place could be the `cmd` package (bad, as it's peripheral to everything) or the `app` package (better, but still not ideal).

As I see that this discussion, just as some previous discussions between you and me, tends to go nowhere, could I suggest a compromise? While I don't want to suggest that the proposed design is the best, I believe it's "good enough" in that has no serious technical drawbacks or fundamental issues to be resolved. Could we, while acknowledging our different opinions, just decide to keep it? I will address some of the points that you make above:

* Making sure parsing of command line flags fails with an unknown name
* Moving `Disable` to its own struct and making it a `StringSlice`
* Using `set` instead of plain `map[string]bool`
* Anything I missed?

adracus:
I see that the coding you've done works but I also see this part of the code as an integral building block of our code that many of our existing controllers already use. Also, when mitigating the mentioned issues in the future I think some of the points here might arise, so I'd also like to keep signature changes minimal. With the current `AddToManagerBuilder` and the `Option` 'framework' I'd say that we already have well-proven building blocks in our code that allow implementing this with minimal changes to existing code.

@rfranzke and me sat together and tried to come up with a solution that addresses your issues (duplication, doing things over again) and my issues (separating the parsing and failing early).
You can find the draft here: https://github.com/adracus/gardener-extensions/tree/feature.controller-switches

The core design is having the `ControllerSwitches` as an `Option`. You then define your mapping from name to function only once in the 'root controller package' per provider and can also generate the 'take-it-all' `AddToManagerFunc` from this mapping. Then you can use this mapping in a new `cmd` package per provider to build up the `ControllerSwitches` per provider and then use this and the created `AddToManager` function `app/main.go`.

After doing this, the same can also be done for Webhooks.

So, to sum up, I'd still suggest to:
* Implement the switch options: https://github.com/adracus/gardener-extensions/blob/feature.controller-switches/pkg/controller/cmd/options.go#L300
* Create the mapping from name to function once in the provider package: https://github.com/adracus/gardener-extensions/blob/feature.controller-switches/controllers/provider-aws/pkg/controller/controller.go#L29
* Use the generated `AddToManager` function in `app/main.go`: https://github.com/adracus/gardener-extensions/blob/feature.controller-switches/controllers/provider-aws/cmd/gardener-extension-provider-aws/app/app.go#L104

stoyanr:
I looked at your changes. I am barely able to understand this `ControllerSwitchOptions` thing. I have to squint and ponder to understand how it works. I still don't believe that all this complication is an improvement over what was initially meant to be a relatively simple and straightforward change.

Not to mention that it all has to be duplicated for the webhooks as well (as the `AddToManager` signature is different there).

But ok, this is what you decided, stopping arguing at this point.

In this case, I don't think you need me here. You have most of it in your branch, you could just complete and merge it. It has very little to do with what's in this PR anyway.

rfranzke:
As discussed, this PR will be closed in favour of https://github.com/adracus/gardener-extensions/tree/feature.controller-switches.
@adracus can you please finish the work and file a new PR?

PR #81, opened 2019-05-09
Enhance Infrastructure controller with custom Watch controls

svetlinas:
Enhances Infrastructure controller's `AddArgs` to include `WatchBuilder` argument which contains additional Watch controllers defined and passed by a provider.

### DIFF
--- a/pkg/controller/infrastructure/controller.go
+++ b/pkg/controller/infrastructure/controller.go
@@ -46,10 +46,8 @@ type AddArgs struct {
 	// Predicates are the predicates to use.
 	// If unset, GenerationChangedPredicate will be used.
 	Predicates []predicate.Predicate
-	// WantsClusterWatch specifies whether the infrastructure actuator requests to have
-	// a watch on `Cluster` objects (which will result into mapping and automatic requeuing
-	// of `Infrastructure` objects).
-	WantsClusterWatch bool
+	// WatchBuilder defines additional watches on controllers that should be set up.
+	WatchBuilder extensionscontroller.WatchBuilder
 }
 
 // DefaultPredicates returns the default predicates for an infrastructure reconciler.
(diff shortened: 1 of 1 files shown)


pkg/controller/infrastructure/controller.go
```diff
@@ -77,24 +79,31 @@ func DefaultPredicates(client client.Client, typeName string, ignoreOperationAnn
...
+	if err := ctrl.Watch(&source.Kind{Type: &extensionsv1alpha1.Infrastructure{}}, &handler.EnqueueRequestForObject{}, args.Predicates...); err != nil {
 		return err
 	}
-	if err := ctrl.Watch(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: SecretToInfrastructureMapper(mgr.GetClient(), predicates)}); err != nil {
+	if err := ctrl.Watch(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: SecretToInfrastructureMapper(mgr.GetClient(), args.Predicates)}); err != nil {
 		return err
 	}
-	if clusterWatch {
-		if err := ctrl.Watch(&source.Kind{Type: &extensionsv1alpha1.Cluster{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: ClusterToInfrastructureMapper(mgr.GetClient(), predicates)}); err != nil {
+	if args.WantsClusterWatch {
```
rfranzke: Now that we can enable adding any arbitrary watch we probably can remove this `WantsClusterWatch`. Any actuator that wants to have this can simply use the `WatchBuilder` to register this. WDYT?

svetlinas: You are absolutely right. I was still wondering if this argument has not been used by some providers.
However, I cannot see any provider from gardener-extensions that uses it. In this case, I will update the PR.

pkg/controller/infrastructure/controller.go
```diff
@@ -77,24 +75,26 @@ func DefaultPredicates(client client.Client, typeName string, ignoreOperationAnn
...
 	}
-	if err := ctrl.Watch(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: SecretToInfrastructureMapper(mgr.GetClient(), predicates)}); err != nil {
+	if err := ctrl.Watch(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: SecretToInfrastructureMapper(mgr.GetClient(), args.Predicates)}); err != nil {
 		return err
 	}
-	if clusterWatch {
-		if err := ctrl.Watch(&source.Kind{Type: &extensionsv1alpha1.Cluster{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: ClusterToInfrastructureMapper(mgr.GetClient(), predicates)}); err != nil {
+
+	// Add additional watches to the controller besides the standard one.
+	for _, f := range args.WatchBuilder {
```
adracus: We don't need to iterate here, we can just call `args.WatchBuilder.AddToController` - this does the error handling etc. automatically for us

svetlinas: Correct. Thanks. I will update the PR

adracus:
Ahh linter caught an error:
```pkg/controller/infrastructure/controller.go:96:35: Error return value of `args.WatchBuilder.AddToController` is not checked (errcheck)
	args.WatchBuilder.AddToController(ctrl)```

PR #83, opened 2019-05-10
Refactor controlplane webhooks to eliminate code duplication

stoyanr:
Refactors the AWS and GCP controlplane mutators into a generic mutator and provider-specific "ensurers" in order to eliminate code duplication.

The proposed approach is similar to #76.

A new `Ensurer` interface is introduced in `pkg/webhook/controlplane` with the following methods:

```go
type Ensurer interface {
    EnsureKubeAPIServerService(context.Context, *corev1.Service) error
    EnsureKubeAPIServerDeployment(context.Context, *appsv1.Deployment) error
    EnsureKubeControllerManagerDeployment(context.Context, *appsv1.Deployment) error
    EnsureKubeSchedulerDeployment(context.Context, *appsv1.Deployment) error
    EnsureKubeletServiceUnitOptions(context.Context, []*unit.UnitOption) ([]*unit.UnitOption, error)
    EnsureKubeletConfiguration(context.Context, *kubeletconfigv1beta1.KubeletConfiguration) error
}
```

In addition, a generic `Mutator` is introduced in `pkg/webhook/controlplane` into which an `Ensurer` is injected. The existing provider-specific `Mutators` are converted to `Ensurers`. The `Mutator` interface is kept and can be implemented directly if a mutator needs to mutate something else than the objects mentioned above.

Note that the current performance / behavior of webhooks is not be affected by this change, since which webhooks are called for which object types is configured completely independently from the `Mutator` implementation.

**Release note**:

```improvement operator
Refactor controlplane webhooks to eliminate code duplication
```


stoyanr:
> Could you move the `genericmutator` into a separate package in `pkg/webhook/controlplane` (and the `genericactuator` as well)?

Did it for `genericmutator`. For `genericactuator`, I will do this as part of #91 to avoid merge conflicts.

PR #85, opened 2019-05-14
Implement alicloud infrastructure controller

### LINKED ISSUE #26: Implement `Infrastructure` controller for Alicloud provider
Similar to how we have implemented the `Infrastructure` extension resource controller for the [AWS provider](https://github.com/gardener/gardener-extensions/tree/master/controllers/provider-aws/pkg/infrastructure) let's please now do it for GCP.

Based on [the current implementation](https://github.com/gardener/gardener/blob/master/example/90-shoot-alicloud.yaml#L13-L18) the `InfrastructureConfig` should look like this:

```yaml
apiVersion: alicloud.provider.extensions.gardener.cloud/v1alpha1
kind: InfrastructureConfig
networks:
  vpc:
    cidr: 10.250.0.0/16
#   id: some-id
  zones:
  - name: zone-1a
    workers: 10.250.0.0/19
```

Based on the [current implementation](https://github.com/gardener/gardener/blob/master/charts/seed-terraformer/charts/alicloud-infra/templates/_main.tf#L102-L112) the `InfrastructureStatus` should look like this:

```yaml
---
apiVersion: alicloud.provider.extensions.gardener.cloud/v1alpha1
kind: InfrastructureStatus
keyName: ssh-key-name
networks:
  vpc:
    id: vpc-id
    vswitches:
    - purpose: nodes
      id: vswitch-id
      zone: zone-1a
    securityGroups:
    - purpose: nodes
      id: sec-group-id
```

The current infrastructure creation/deletion implementation can be found [here](https://github.com/gardener/gardener/blob/master/pkg/operation/cloudbotanist/alicloudbotanist/infrastructure.go). Please try to change as little as possible (with every change the risk that we break something increases!) and just move the code over into the extensions infrastructure actuator.

adracus:
Implementation of the infrastructure controller for Alicloud.
This PR also comes with some interfaces that allow mocking external dependencies such as the `Terraformer` and the creation of it.

Fixes #26 

**Release note**:

```improvement operator
There is now an implementation of the infrastructure resource for Alicloud.
```


controllers/provider-alicloud/pkg/controller/infrastructure/terraform.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	"github.com/gardener/gardener-extensions/controllers/provider-alicloud/pkg/apis/alicloud/v1alpha1"
+	"github.com/gardener/gardener-extensions/pkg/controller"
+	extensionsterraformer "github.com/gardener/gardener-extensions/pkg/gardener/terraformer"
+	gardencorev1alpha1 "github.com/gardener/gardener/pkg/apis/core/v1alpha1"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ComputeCreateVPCInitializerValues computes the InitializerValues to create a new VPC.
```
stoyanr: IMO all functions in this file should be unexported. As far as I understand, they are currently only exported in order to be tested. However, you are already testing them as part of the actuator implementation, so testing them independently is redundant.

Alternatively, if you would rather prefer keeping them and the tests for them, you could implement them as the methods of a type (or types), e.g. `valuesProvider` for the fist 3, and inject this type via an interface into the actuator.

In summary, I believe with dependency injection you should aim at the following 3 cases:

* Unexported functions that are used to provide internal structure for your package. They are not tested independently and could be refactored into a different set of functions at any moment without any tests being broken.
* Explicitly declared dependencies via interfaces that are always mocked away in tests.
* Inexplicit dependencies to utility functions / types that are intended for reuse by different packages. These are typically "simple" enough to justify direct usage instead of injection via an interface.

As a result, there should not be exported utility functions / types in the package that are used solely by this package.

pkg/gardener/chartrenderer/chartrenderer.go
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+package chartrenderer
+
+import (
+	"github.com/gardener/gardener/pkg/chartrenderer"
+	"k8s.io/client-go/rest"
+)
+
+type factory struct{}
```
stoyanr: Instead of an empty struct, I would create a function type satisfying the Factory interface:

```go
type FactoryFunc func(*rest.Config) (chartrenderer.Interface, error)

func (f FactoryFunc) NewForConfig(config *rest.Config) (chartrenderer.Interface, error) {
	return f(config)
}
```

And then use the `chartrenderer.NewForConfig` function as the implementation.

pkg/gardener/terraformer/terraformer.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	gardenerterraformer "github.com/gardener/gardener/pkg/operation/terraformer"
+	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type terraformer struct {
```
stoyanr: This type and its methods are not needed. `*gardenerterraformer.Terraformer` already implements `Interface`

pkg/gardener/terraformer/terraformer.go
```diff
@@ -0,0 +1,93 @@
...
+func (t *terraformer) GetStateOutputVariables(variables ...string) (map[string]string, error) {
+	return t.tf.GetStateOutputVariables(variables...)
+}
+
+// ConfigExists implements Terraformer.
+func (t *terraformer) ConfigExists() (bool, error) {
+	return t.tf.ConfigExists()
+}
+
+type initializer struct {
```
stoyanr: This type is probably not needed as well. Whatever type `gardenerterraformer.DefaultInitializer` returns already implements `Initializer`, either directly or through an adapter `Func` type.

pkg/gardener/terraformer/terraformer.go
```diff
@@ -0,0 +1,93 @@
...
+	initialize func(config *gardenerterraformer.InitializerConfig) error
+}
+
+// Initialize implements Initializer.
+func (i *initializer) Initialize(config *gardenerterraformer.InitializerConfig) error {
+	return i.initialize(config)
+}
+
+type factory struct{}
```
stoyanr: Since this interface has several methods, here using a `struct` is justified. However, at least the implementation could be simplified a lot by removing the `terraformer` and `initializer` types.

controllers/provider-alicloud/pkg/alicloud/client/types.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// Credentials are the credentials to access the Alicloud API.
+type Credentials struct {
+	AccessKeyID     string
+	AccessKeySecret string
+}
+
+// VPC is the interface to the Alicloud VPC service.
+type VPC interface {
```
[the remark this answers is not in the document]
adracus: The package is already called `client`, that's why I chose to call it `VPC`.

controllers/provider-alicloud/pkg/alicloud/client/types.go
```diff
@@ -0,0 +1,39 @@
...
+	// DescribeVPCs describes the VPCs for the request.
+	DescribeVPCs(req *alicloudvpc.DescribeVpcsRequest) (*alicloudvpc.DescribeVpcsResponse, error)
+	// DescribeNATGateways describes the NAT gateways for the request.
+	DescribeNATGateways(req *alicloudvpc.DescribeNatGatewaysRequest) (*alicloudvpc.DescribeNatGatewaysResponse, error)
+}
+
+// Factory is the factory to instantiate Alicloud clients.
+type Factory interface {
+	// NewVPCClientWithAccessKey creates a new VPC client from the given credentials and region.
+	NewVPCClientWithAccessKey(region, accessKeyID, accessKeySecret string) (VPC, error)
```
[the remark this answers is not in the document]
adracus: That's not possible as far as I can see: If we have the factory interface like
```go
type Factory interface {
    NewVPCClientWithAccessKey(region, accessKeyID, accessKeySecret string) (VPC, error)
}
```
then this function is not compatible with that interface (go doesn't do automatic casts like that).

controllers/provider-alicloud/pkg/alicloud/client/client.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, fmt.Errorf("secret %s/%s has no access key secret at data.%s", secret.Namespace, secret.Name, AccessKeySecretField)
+	}
+
+	return &Credentials{
+		AccessKeyID:     string(accessKeyID),
+		AccessKeySecret: string(accessKeySecret),
+	}, nil
+}
+
+type factory struct{}
```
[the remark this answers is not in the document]
adracus: What is the benefit of using a `func`? The good thing about the `struct` is that we don't have to change anything if we choose to add further constructors / functions in the future.

pkg/gardener/terraformer/terraformer.go
adracus: Not exactly: `Interface` expects the terraformer `Interface` as return type for some functions. Go won't do automatic casts if your implementation returns the concrete type, thus the interface is required.

controllers/provider-alicloud/pkg/alicloud/client/client.go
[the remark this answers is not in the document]
adracus: What's the benefit of exposing an additional `FactoryFunc` type? The empty struct as of now can be easily extended with further factory functions and we don't expose anything additionally.

pkg/gardener/chartrenderer/chartrenderer.go
adracus: Same question as for the client factory function: Why expose an additional type that is not that easily extendable? If we never show how the factory is implemented, why is it important to have this function type?

pkg/gardener/terraformer/terraformer.go
adracus: As mentioned, `terraformer` cannot be removed, and for `Initializer` you'll still at least need the function adapter type. I've changed from the `initializer` struct to an `initializerFunc`, but I guess you probably can't do more about it.

controllers/provider-alicloud/pkg/alicloud/client/types.go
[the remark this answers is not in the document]
stoyanr: The `Factory` interface should be changed as well, see above.

controllers/provider-alicloud/pkg/alicloud/client/client.go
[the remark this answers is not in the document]
stoyanr: The benefit is that you can provide something existing as an implementation, so you write less code. The `FactoryFunc` is just an adapter between the function and the interface it implements. In future versions of Go this type might not be needed, for now it's needed to satisfy the type system. It's quite trivial and I don't see any harm in exposing it.

pkg/gardener/chartrenderer/chartrenderer.go
stoyanr: See above. Although it's relatively minor, IMO an adapter type captures the intention to implement the interface via a specific function (or a set of function) better by making it more explicit. It can always be refactored if it should be extended at some point.

controllers/provider-alicloud/pkg/alicloud/client/types.go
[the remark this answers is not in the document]
stoyanr: OK, but then let's stick to this `VPC` everywhere, right now it's a mixture of `VPC`, 'Client`, and `VPCClient` in different names.

pkg/gardener/terraformer/terraformer.go
stoyanr: How about changing the `Interface` type so that `*gardenerterraformer.Terraformer` conforms? Go would do automatic casts on the concrete types returned by the methods in this case.

controllers/provider-alicloud/pkg/alicloud/client/types.go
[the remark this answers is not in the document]
adracus: I can't change the interface of the factory because then I'm not able to provide a mock implementation of the client if the factory always returns the concrete type.

pkg/gardener/terraformer/terraformer.go
adracus: I don't think you can change the interface that way: Take the `InitializeWith` method for example: In Gardener, this is `InitializeWith(initializer Initializer) *Terraformer`. It always returns the concrete `*Terraformer` type. Since we want to be able to return arbitrary mock implementations with the `Interface`, our definition for the `InitializeWith` method has to return the `Interface`, otherwise it's useless. Also I see that the methods might be the same, but as said: go does *not* do automatic casts, there simply isn't a thing like that. If you see a way of doing this let me know.

stoyanr: That's right, I didn't think about it. OK, please ignore these comments then.

stoyanr: I'll think about it when I have time, but that might take a few days, I am at KubeCon this week.

rfranzke:
rebase is needed now

controllers/provider-alicloud/charts/internal/alicloud-infra/templates/main.tf
```diff
@@ -0,0 +1,115 @@
...
+  vpc_id            = "{{ required "vpc.id is required" $.Values.vpc.id }}"
+  cidr_block        = "{{ required "zone.cidr.worker is required" $zone.cidr.worker }}"
+  availability_zone = "{{ required "zone.name is required" $zone.name }}"
+}
+
+// Create a new EIP.
+resource "alicloud_eip" "eip_natgw_z{{ $index }}" {
+  name                 = "{{ required "clusterName is required" $.Values.clusterName }}-eip-natgw-z{{ $index }}"
+  bandwidth            = "20"
+  internet_charge_type = "PayByBandwidth"
```
rfranzke: https://github.com/gardener/gardener/pull/994 is not yet reflected here in this PR - please implement these changes in this extension controller as well

rfranzke:
AFAIK you're busy with another task, so I'll try to finish this by pushing against your branch.

adracus:
Alright, feel free to go ahead!

PR #86, opened 2019-05-14
Implement Controller and Webhook Switches

adracus:
This allows turning individual controllers / webhooks on and off per command line flag.

**Release note**:

```improvement operator
It is now possible to turn individual webhooks and controllers on and off via the command line flag `--disable-controllers` and `--disable-webhooks`.
```


controllers/provider-aws/pkg/cmd/options.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/gardener/gardener-extensions/pkg/controller/cmd"
+	extensionscontrolplanecontroller "github.com/gardener/gardener-extensions/pkg/controller/controlplane"
+	extensionsinfrastructurecontroller "github.com/gardener/gardener-extensions/pkg/controller/infrastructure"
+	extensioncontrolplanewebhook "github.com/gardener/gardener-extensions/pkg/webhook/controlplane"
+)
+
+// ControllerSwitchOptions are the cmd.ControllerSwitchOptions for the provider controllers.
+func ControllerSwitchOptions() *cmd.ControllerSwitchOptions {
+	opts := cmd.NewControllerSwitchOptions(
+		cmd.ControllerSwitch(extensionsinfrastructurecontroller.ControllerName, infrastructurecontroller.AddToManager),
```
rfranzke: Didn't we agree that we want to hold a map of controller names to their `AddToManager` functions in `controllers/provider-*/pkg/controller/controller.go`, and simply iterate over it here (similar for webhooks)?

This would prevent the double maintenance of the list of controllers and make it easier when adding new controllers to a provider.

Right now, the [list in `../controllers.go`](https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-aws/pkg/controller/controller.go) still exists, and another list has been introduced here.

adracus: Since we'll only use the `cmd.ControllerSwitchOptions` actually having `AddToManager` in the controllers is unused. I removed that and now we again only have one list to maintain.

rfranzke: OK, thank you!

PR #93, opened 2019-05-23
[WIP] Add support for Flatcar Linux (both stable and edge channels)

### LINKED ISSUE #92: Add an option for Flatcar Linux with OperatingSystemConfig
Following up a discussion at KubeCon Barcelona:

Gardener now defines a resource of kind "OperatingSystemConfig" and it could be used to choose the alternative [Flatcar Linux](https://www.flatcar-linux.org/).

As far as I understand, the file [controllers/provider-aws/charts/provider-aws/values.yaml](https://github.com/gardener/gardener-extensions/blob/d9aef6c88ecf188d4d3c97dc65de9f59b79eea00/controllers/provider-aws/charts/provider-aws/values.yaml#L19) defines the AWS AMIs for "coreos". We could add an entry for "flatcar" here with its [4 channels](https://www.flatcar-linux.org/releases/): stable, beta, alpha, [edge](https://kinvolk.io/blog/2019/05/introducing-the-flatcar-linux-edge-channel/)

This was motivated by [Inspektor Gadget](https://github.com/kinvolk/inspektor-gadget) that currently works on the edge channel of Flatcar Linux.

This would also require to pass a [new parameter on the Kubelet command line](https://github.com/kinvolk/inspektor-gadget/blob/0e6c3fd71b4775c0e385ccd36afc08773a8ec594/Documentation/install.md#on-another-kubernetes-distribution) (`--cgroup-driver=systemd`) or [KubeletConfiguration](https://github.com/kinvolk/lokomotive-kubernetes/commit/22fb201c1e21034e5f5e81dd0f63d633be66370c#diff-97942f9b4a5f15c0b66ee126ef497e26R170) (`cgroupDriver: systemd`). But I am not sure where to specify that in the gardener extensions. Any pointers?

/cc @vasu1124

alban:
This adds an option for Flatcar Linux as an alternative to CoreOS Container Linux. I added both the stable channel (for production workload) and the edge channel (for experimentation only).

Fixes #92

As mentioned in #92, we need to pass an additional parameter to the Kubelet:
- `--cgroup-driver=systemd` on the edge channel
- not necessary on the stable channel.
I am not sure where I would specify this.

**Release note**:

```improvement user
Add support for Flatcar Linux (both stable and edge channels)
```

### DIFF
--- a/controllers/provider-aws/charts/provider-aws/values.yaml
+++ b/controllers/provider-aws/charts/provider-aws/values.yaml
@@ -16,6 +16,9 @@ controllers:
 
 config:
   machineImages:
+  # AMIs could be found on:
+  # https://coreos.com/releases/
+  # https://coreos.com/os/docs/latest/booting-on-ec2.html
   - name: coreos
     version: 1967.5.0
     regions:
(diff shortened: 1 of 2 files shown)


rfranzke:
Hi @alban, thanks for the PR, however, this alone won't help supporting flatcar. You'll have to submit an extension controller (see https://github.com/gardener/gardener/blob/master/docs/extensions/operatingsystemconfig.md), similar to https://github.com/gardener/gardener-extensions/tree/master/controllers/os-coreos or https://github.com/gardener/gardener-extensions/tree/master/controllers/os-suse-jeos.

rfranzke:
Closing this one for now as it cannot be used as is and as it is not quickly to fix.

PR #94, opened 2019-05-27
Implement customizing of etcd stateful sets via webhooks for AWS and GCP

### LINKED ISSUE #55: Implement `ControlPlane` controller for GCP provider
Similar to how we have implemented the `ControlPlane` extension resource controller for the [AWS provider](https://github.com/gardener/gardener-extensions/tree/master/controllers/provider-aws/pkg/controller/controlplane) let's please now do it for GCP.

Based on [the current implementation](https://github.com/gardener/gardener/blob/master/example/90-shoot-gcp.yaml) the `ControlPlaneConfig` should look like this:

```yaml
apiVersion: gcp.provider.extensions.gardener.cloud/v1alpha1
kind: ControlPlaneConfig
cloudControllerManager:
  zone: europe-west1
  featureGates:
    CustomResourceValidation: true
```

No `ControlPlaneStatus` needs to be implemented right now (not needed yet).

stoyanr:
Adds generic code for customizing etcd stateful sets via webhooks and implements the actual customizing of etcd stateful sets via webhooks for AWS and GCP.

Fixes #55 

The code for customizing etcd stateful sets via webhooks is split between 2 types of webhooks:
* `controlplaneexpsure` for replacing the default storage class with a non-default one, if configured.
* `controlplanebackup` for injecting the `backup-restore` container.

This PR also:
* Fixes broken webhook startup by previous PRs (#86).
* Refactors slightly webhook switches for clarity and consistency with controllers.
* Moves `NoopEnsurer` to `genericmutator` package.

**Release note**:

```improvement operator
Implement customizing of etcd stateful sets via webhooks for AWS and GCP.
```


controllers/provider-aws/charts/provider-aws/templates/configmap.yaml
```diff
@@ -15,3 +15,8 @@ data:
     machineImages:
 {{ toYaml .Values.config.machineImages | indent 4 }}
 {{- end }}
+    etcd:
```
rfranzke: Cosmetics: WDYT about

```yaml
etcd:
  storage:
    className: ...
    capacity: ...
  backup:
    schedule: ...
```

?

controllers/provider-aws/charts/provider-aws/values.yaml
```diff
@@ -19,6 +19,7 @@ disableControllers:
 disableWebhooks:
 - controlplane
 - controlplaneexposure
+- controlplanebackup
```
rfranzke:
```suggestion
- etcdbackup
```

To make it more precise what it is used for

controllers/provider-aws/pkg/apis/config/types.go
```diff
@@ -46,3 +51,20 @@ type RegionAMIMapping struct {
 	// AMI is the AMI for the machine image.
 	AMI string
 }
+
+// ETCD is an etcd configuration.
+type ETCD struct {
+	// StorageClassName is the name of the storage class used in etcd-main volume claims.
+	// +optional
```
rfranzke: `// +optional` tags are not needed in the internal version

controllers/provider-aws/pkg/webhook/controlplaneexposure/add.go
```diff
@@ -15,6 +15,7 @@
 package controlplaneexposure
```
rfranzke: Can we now maybe find a better name than `controlplaneexposure`?

pkg/webhook/controlplane/genericmutator/mutator.go
```diff
@@ -92,6 +97,17 @@ func (m *mutator) Mutate(ctx context.Context, obj runtime.Object) error {
 		case common.KubeSchedulerDeploymentName:
 			return m.ensurer.EnsureKubeSchedulerDeployment(ctx, x)
 		}
+	case *appsv1.StatefulSet:
+		switch x.Name {
+		case common.EtcdMainStatefulSetName, common.EtcdEventsStatefulSetName:
```
rfranzke: Maybe I missed this part: How do we ensure that the `etcd-events` stateful set does get the backup-restore sidecar container, however, configured *WITHOUT* backup?

zanetworker (commented):
Thanks for the PR @stoyanr, not much comments from my side. 

/lgtm

controllers/provider-aws/charts/provider-aws/templates/configmap.yaml
stoyanr: I think it's better, done.

controllers/provider-aws/charts/provider-aws/values.yaml
stoyanr: As discussed, we should rather keep the `controlplane` prefix for these webhooks as they are referred to as "controlplane" webhooks in several places.

controllers/provider-aws/pkg/apis/config/types.go
stoyanr: Removed.

pkg/webhook/controlplane/genericmutator/mutator.go
stoyanr: As discussed, it's me who missed this. Now it should be ok, changed it in `controlplane.GetBackupRestoreContainer`.

controllers/provider-aws/pkg/webhook/controlplaneexposure/add.go
stoyanr: As discussed, we don't have a good proposal now. If we come up with one, the renaming should be done in a separate PR.

controllers/provider-aws/pkg/apis/config/types.go
```diff
@@ -46,3 +51,20 @@ type RegionAMIMapping struct {
 	// AMI is the AMI for the machine image.
 	AMI string
 }
+
+// ETCD is an etcd configuration.
+type ETCD struct {
```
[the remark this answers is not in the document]
stoyanr: I think `ETCD` is better. `Config` would be redundant:
* The package is `config`, so externally it's visible as `config.ETCD`.
* The YAML field is also `etcd` as it's part of a `Configuration` resource.

controllers/provider-aws/pkg/apis/config/types.go
```diff
@@ -46,3 +51,20 @@ type RegionAMIMapping struct {
...
+	// StorageClassName is the name of the storage class used in etcd-main volume claims.
+	// +optional
+	StorageClassName *string
+	// StorageCapacity is the storage capacity used in etcd-main volume claims.
+	// +optional
+	StorageCapacity *resource.Quantity
+}
+
+// ETCDBackup is an etcd backup configuration.
+type ETCDBackup struct {
```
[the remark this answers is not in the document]
stoyanr: I think `ETCDBackup` is better, see above.

controllers/provider-aws/pkg/apis/config/types.go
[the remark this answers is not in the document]
zanetworker: oh yes, didn't notice that the package name is config.

controllers/provider-aws/charts/provider-aws/values.yaml
```diff
@@ -37,3 +38,8 @@ config:
       ami: ami-07739b17529e8c1d0
     - name: ap-southeast-2
       ami: ami-02d7d488d701a460e
+  etcd:
+    storageClass: gardener.cloud-fast
```
rfranzke: To avoid confusion I would also change the values API to match the new ConfigMap API.

controllers/provider-gcp/charts/provider-gcp/values.yaml
```diff
@@ -19,9 +19,15 @@ disableControllers:
...
 - controlplaneexposure
+- controlplanebackup
 
 config:
   machineImages:
   - name: coreos
     version: 2023.5.0
     image: projects/coreos-cloud/global/images/coreos-stable-2023-5-0-v20190312
+  etcd:
+    storageClass: gardener.cloud-fast
```
rfranzke: To avoid confusion I would also change the values API to match the new ConfigMap API.

pkg/webhook/controlplane/etcd.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	// Ensure additional command line args
+	for k, v := range args {
+		c.Command = EnsureStringWithPrefix(c.Command, fmt.Sprintf("--%s=", k), v)
+	}
+
+	// Ensure additional env variables
+	for _, envVar := range env {
+		c.Env = EnsureEnvVarWithName(c.Env, envVar)
```
rfranzke: Should we not add any env variables for backup if we are mutating the etcd-events statefulset? The empty provider flag should already prevent the sidecar from taking backups, so it would be little cleaner to not add the configuration (= env variables).

stoyanr: Is it certain that all env variables are only needed if backup is enabled? If yes, this is a good idea. If not, then the logic for checking the "role" should be moved to the invoking code, which would be a bigger change.

rfranzke: Yes, the env variables are only needed if backup is enabled. That's how it works today.

controllers/provider-aws/charts/provider-aws/values.yaml
stoyanr: Done.

controllers/provider-gcp/charts/provider-gcp/values.yaml
stoyanr: Done.

pkg/webhook/controlplane/etcd.go
stoyanr: Still, I moved the role checking to the ensurers, I believe it's cleaner this way. I also added tests for both etcd-main and etcd-events.

rfranzke (approved):
/lgtm

PR #124, opened 2019-06-11
Enhance controlplane webhooks to do more provider-specific modifications

### LINKED ISSUE #113: Enhance controlplane webhooks to add provider-specific settings to /etc/sysctl.d/99-k8s-general.conf
For some providers (namely AWS and GCP), Gardener adds provider-specific settings to the file `/etc/sysctl.d/99-k8s-general.conf` which is part of the `OperatingSystemConfig` resource, see https://github.com/gardener/gardener/blob/eb083ff1f87b131f2c27ddb488e4ca16d24a9b39/charts/seed-operatingsystemconfig/original/templates/config/_99-k8s-general.conf#L57-L64.
The corresponding controlplane webhooks for AWS and GCP should add these settings as well.

svetlinas:
Enhance ControlPlane webhooks to do additional provider-specific modifications

Fixes #113 

**Release note**:

```improvement operator
Enhance ControlPlane webhooks to do additional provider-specific modifications
```

### DIFF
--- a/controllers/provider-aws/pkg/webhook/controlplane/ensurer.go
+++ b/controllers/provider-aws/pkg/webhook/controlplane/ensurer.go
@@ -15,8 +15,8 @@
 package controlplane
 
 import (
+	"bytes"
 	"context"
-
 	"github.com/gardener/gardener-extensions/controllers/provider-aws/pkg/aws"
 	"github.com/gardener/gardener-extensions/pkg/webhook/controlplane"
 	"github.com/gardener/gardener-extensions/pkg/webhook/controlplane/genericmutator"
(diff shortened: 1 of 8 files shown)


controllers/provider-aws/pkg/webhook/controlplane/ensurer.go
```diff
@@ -172,3 +175,17 @@ func (e *ensurer) EnsureKubeletConfiguration(ctx context.Context, kubeletConfig
 	delete(kubeletConfig.FeatureGates, "CSIDriverRegistry")
 	return nil
 }
+
+// EnsureKubernetesGeneralConfiguration ensures that the kubernetes general configuration conforms to the provider requirements.
+func (e *ensurer) EnsureKubernetesGeneralConfiguration(ctx context.Context, fci *extensionsv1alpha1.FileContentInline) error {
```
stoyanr: I think it will be a bit better to use `*string` instead of `*extensionsv1alpha1.FileContentInline`, since we are not using anything from the struct except its `Data` member.

PR #127, opened 2019-06-11
Add checksum annotations in controlplane webhooks

### LINKED ISSUE #115: Enhance controlplane webhooks to inject cheksums as pod annotations
Gardener injects checksums as pod annotations for all secrets / configmaps used by a deployment / stateful set. Now that certain provider-specific secrets and configmaps are injected by controlplane webhooks, these webhooks should take care of injecting the corresponding checksums as well. 

This includes the following checksums for providers that actually inject the corresponding secrets / configmaps as volumes or sources for env variables:

* `checksum/secret-etcd-backup` in `etcd-main` pods
* `checksum/secret-cloudprovider` and `checksum/configmap-cloud-provider-config` in `kube-apiserver` pods
* `checksum/secret-cloudprovider` and `checksum/configmap-cloud-provider-config` in `kube-controller-manager` pods

stoyanr:
Controlplane webhooks now add checksum annotations for all secrets / configmaps that are being mounted as volumes or used as a source of env vars.

Fixes #115 

It also removes some volumes mounted by controlplane webhooks that are actually not needed.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/controllers/provider-alicloud/pkg/webhook/controlplanebackup/ensurer.go
+++ b/controllers/provider-alicloud/pkg/webhook/controlplanebackup/ensurer.go
@@ -29,6 +29,7 @@ import (
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // NewEnsurer creates a new controlplaneexposure ensurer.
@@ -44,7 +45,14 @@ type ensurer struct {
 	genericmutator.NoopEnsurer
 	etcdBackup  *config.ETCDBackup
 	imageVector imagevector.ImageVector
+	client      client.Client
 	logger      logr.Logger
+}
+
+// InjectClient injects the given client into the ensurer.
+func (e *ensurer) InjectClient(client client.Client) error {
+	e.client = client
+	return nil
 }
 
 // EnsureETCDStatefulSet ensures that the etcd stateful sets conform to the provider requirements.
@@ -52,7 +60,7 @@ func (e *ensurer) EnsureETCDStatefulSet(ctx context.Context, ss *appsv1.Stateful
 	if err := e.ensureContainers(&ss.Spec.Template.Spec, ss.Name, cluster); err != nil {
 		return err
 	}
-	return nil
+	return e.ensureChecksumAnnotations(ctx, &ss.Spec.Template, ss.Namespace, ss.Name)
 }
 
 func (e *ensurer) ensureContainers(ps *corev1.PodSpec, name string, cluster *extensionscontroller.Cluster) error {
@@ -61,6 +69,13 @@ func (e *ensurer) ensureContainers(ps *corev1.PodSpec, name string, cluster *ext
 		return err
 	}
 	ps.Containers = controlplane.EnsureContainerWithName(ps.Containers, *c)
+	return nil
+}
+
+func (e *ensurer) ensureChecksumAnnotations(ctx context.Context, template *corev1.PodTemplateSpec, namespace, name string) error {
+	if name == common.EtcdMainStatefulSetName {
+		return controlplane.EnsureSecretChecksumAnnotation(ctx, template, e.client, namespace, alicloud.BackupSecretName)
+	}
 	return nil
 }
 
--- a/controllers/provider-aws/pkg/webhook/controlplane/ensurer.go
+++ b/controllers/provider-aws/pkg/webhook/controlplane/ensurer.go
@@ -27,6 +27,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // NewEnsurer creates a new controlplane ensurer.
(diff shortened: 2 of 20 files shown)


controllers/provider-alicloud/pkg/webhook/controlplanebackup/ensurer.go
```diff
@@ -64,6 +75,15 @@ func (e *ensurer) ensureContainers(ps *corev1.PodSpec, name string, cluster *ext
 	return nil
 }
 
+func (e *ensurer) ensureChecksumAnnotations(ctx context.Context, template *corev1.PodTemplateSpec, namespace, name string) error {
+	if name == common.EtcdMainStatefulSetName {
+		if err := controlplane.EnsureSecretChecksumAnnotation(ctx, template, e.client, namespace, alicloud.BackupSecretName); err != nil {
```
rfranzke: Who is actually adding this secret volume to the `StatefulSet` spec? Is this missing at all? I guess Gardener cannot add the secret by default. In the future, the backup extension controller will create the secret anyways (and not Gardener anymore).

Similarly, I think Gardener should not add + mount the client certificate secret anymore, right? It's only needed by the backup-restore sidecar which we are injecting. So, this is still missing (apart from the PR), is this assumption correct?

pkg/webhook/controlplane/checksums.go
```diff
@@ -0,0 +1,57 @@
...
+// EnsureSecretChecksumAnnotation ensures that the given pod template has an annotation containing the checksum of the
+// secret with the given name and namespace.
+func EnsureSecretChecksumAnnotation(ctx context.Context, template *corev1.PodTemplateSpec, c client.Client, namespace, name string) error {
+	// Get secret from cluster
+	secret := &corev1.Secret{}
+	if err := c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, secret); err != nil {
+		return errors.Wrapf(err, "could not get secret '%s/%s'", namespace, name)
+	}
+
+	// Add checksum annotation
```
rfranzke: Let's use https://github.com/kubernetes/apimachinery/blob/de88ae2d04de1dcd8c34bedffd88a2b8efeac007/pkg/apis/meta/v1/helpers.go#L195?

pkg/webhook/controlplane/checksums.go
```diff
@@ -0,0 +1,57 @@
...
+// EnsureConfigMapChecksumAnnotation ensures that the given pod template has an annotation containing the checksum of the
+// configmap with the given name and namespace.
+func EnsureConfigMapChecksumAnnotation(ctx context.Context, template *corev1.PodTemplateSpec, c client.Client, namespace, name string) error {
+	// Get configmap from cluster
+	cm := &corev1.ConfigMap{}
+	if err := c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, cm); err != nil {
+		return errors.Wrapf(err, "could not get configmap '%s/%s'", namespace, name)
+	}
+
+	// Add checksum annotation
```
rfranzke: Let's use https://github.com/kubernetes/apimachinery/blob/de88ae2d04de1dcd8c34bedffd88a2b8efeac007/pkg/apis/meta/v1/helpers.go#L195?

controllers/provider-aws/pkg/webhook/controlplane/ensurer.go
```diff
@@ -38,30 +39,45 @@ func NewEnsurer(logger logr.Logger) genericmutator.Ensurer {
...
-	ps := &dep.Spec.Template.Spec
+	template := &dep.Spec.Template
+	ps := &template.Spec
 	if c := controlplane.ContainerWithName(ps.Containers, "kube-controller-manager"); c != nil {
 		ensureKubeControllerManagerCommandLineArgs(c)
 		ensureEnvVars(c)
 		ensureVolumeMounts(c)
 	}
 	ensureVolumes(ps)
+	if err := e.ensureChecksumAnnotations(ctx, template, dep.Namespace); err != nil {
```
rfranzke:
```suggestion
	return e.ensureChecksumAnnotations(ctx, template, dep.Namespace)
```

(replicate for other providers)

controllers/provider-aws/pkg/webhook/controlplane/ensurer.go
```diff
@@ -38,30 +39,45 @@ func NewEnsurer(logger logr.Logger) genericmutator.Ensurer {
...
-	ps := &dep.Spec.Template.Spec
+	template := &dep.Spec.Template
+	ps := &template.Spec
 	if c := controlplane.ContainerWithName(ps.Containers, "kube-apiserver"); c != nil {
 		ensureKubeAPIServerCommandLineArgs(c)
 		ensureEnvVars(c)
 		ensureVolumeMounts(c)
 	}
 	ensureVolumes(ps)
+	if err := e.ensureChecksumAnnotations(ctx, template, dep.Namespace); err != nil {
```
rfranzke:
```suggestion
	return e.ensureChecksumAnnotations(ctx, template, dep.Namespace)
```

(replicate for other providers)

controllers/provider-alicloud/pkg/webhook/controlplanebackup/ensurer.go
rfranzke:
```suggestion
		return controlplane.EnsureSecretChecksumAnnotation(ctx, template, e.client, namespace, alicloud.BackupSecretName)
```

(replicate for other providers)

controllers/provider-aws/pkg/webhook/controlplane/ensurer.go
```diff
@@ -126,26 +139,24 @@ var (
...
 )
 
 func ensureVolumeMounts(c *corev1.Container) {
 	c.VolumeMounts = controlplane.EnsureVolumeMountWithName(c.VolumeMounts, cloudProviderConfigVolumeMount)
-	c.VolumeMounts = controlplane.EnsureVolumeMountWithName(c.VolumeMounts, cloudProviderSecretVolumeMount)
 }
 
 func ensureVolumes(ps *corev1.PodSpec) {
 	ps.Volumes = controlplane.EnsureVolumeWithName(ps.Volumes, cloudProviderConfigVolume)
-	ps.Volumes = controlplane.EnsureVolumeWithName(ps.Volumes, cloudProviderSecretVolume)
```
rfranzke: Just to double check: It's removed because it's not needed to be mounted as it's only used in the env as reference, right?

controllers/provider-alicloud/pkg/webhook/controlplanebackup/ensurer.go
stoyanr: In this particular case the `etcd-backup` secret is not mounted as a volume but is still used as a source of env variables, injected by the webhook, which warrants adding a checksum annotations (but only for `etcd-main`, not `etcd-events`)
Regarding the client certificate secret `etcd-client-tls`, I see it currently mounted to the `etcd` container as well. Do you mean that it's not needed there, or do you mean a different secret (which one)? Looking at the chart in Gardener it seems to me that the only secret that  is not added by Gardener is the `etcd-backup` secret.

controllers/provider-aws/pkg/webhook/controlplane/ensurer.go
stoyanr: Exactly, and the same is true for some of the other providers. But the checksum is needed in both cases.

controllers/provider-alicloud/pkg/webhook/controlplanebackup/ensurer.go
rfranzke: OK, thank you, makes sense!

> Do you mean that it's not needed there, or do you mean a different secret (which one)?
It's not needed I think (@swapnilgm can you confirm that, please?) - at least, I don't see it anywhere used by etcd itself.

controllers/provider-aws/pkg/webhook/controlplane/ensurer.go
rfranzke: Thanks!

controllers/provider-alicloud/pkg/webhook/controlplanebackup/ensurer.go
swapnilgm: > I think Gardener should not add + mount the client certificate secret anymore, right?

Its needed in etcd container for it's liveness probe. So, it should come from Gardener.

Regarding  etcd-backup secret, based on the cloud provider its either mounted or referenced in environment variable. It is required only by backup sidecar. Though currently Gardener creates it. In future  backup extension controller will take responsibility to deploy secret for etcd.

rfranzke: Ah okay, right, thanks for pointing this out!

pkg/webhook/controlplane/checksums.go
stoyanr: Done

pkg/webhook/controlplane/checksums.go
stoyanr: Done

controllers/provider-aws/pkg/webhook/controlplane/ensurer.go
stoyanr: Done

controllers/provider-aws/pkg/webhook/controlplane/ensurer.go
stoyanr: Done

controllers/provider-alicloud/pkg/webhook/controlplanebackup/ensurer.go
stoyanr: Done

PR #142, opened 2019-06-13
Ensure worker CRDs conditionally at startup

rfranzke:
Ensure worker CRDs conditionally at startup to prevent REST mapping issues and not recreate them in every reconciliation loop.

**Release note**:

```noteworthy operator
All provider controllers do now have a new flag `--worker-deploy-crds` which defaults to `true`. If enabled then all worker-relevant CRDs like `MachineDeployment`, etc. will be deployed during startup time of the controller.
```

### DIFF
--- a/pkg/controller/worker/machine_crds.go
+++ b/pkg/controller/worker/machine_crds.go
@@ -12,15 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package genericactuator
+package worker
 
 import (
 	"context"
 
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	apiextensionsscheme "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/scheme"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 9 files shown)


pkg/controller/worker/machine_crds.go
```diff
@@ -297,20 +298,36 @@ func init() {
 	}
 }
 
+// EnsureMachineResourcesForConfig ensures that all well-known machine CRDs are created or updated.
+func EnsureMachineResourcesForConfig(ctx context.Context, config *rest.Config) error {
+	scheme := runtime.NewScheme()
```
adracus: This scheme should only be initialized once during e.g. an `init` function.

pkg/controller/worker/machine_crds.go
```diff
@@ -297,20 +298,36 @@ func init() {
 	}
 }
 
+// EnsureMachineResourcesForConfig ensures that all well-known machine CRDs are created or updated.
+func EnsureMachineResourcesForConfig(ctx context.Context, config *rest.Config) error {
```
adracus: How about having this in k8s style and calling it `ApplyMachineResourcesWithConfig`? Same goes for the function `EnsureMachineResources` -> `ApplyMachineResources`.

PR #156, opened 2019-06-20
Fix azure infra name retrieval

mkoynov:
Fixes a bug when retrieving configs from controlplane.infrastructurestatus

Updates the tests and examples

**Release note**:
```improvement operator
NONE
```


controllers/provider-azure/pkg/webhook/controlplaneexposure/ensurer.go
```diff
@@ -58,6 +58,7 @@ func (e *ensurer) EnsureKubeAPIServerService(ctx context.Context, svc *corev1.Se
 	if err != nil {
 		return err
 	}
+	// TODO: get seed version instead of shoot version
```
stoyanr: Remove the entire `comparedVersion` logic and put the annotation always. Leave a TODO to only do this for seeds with version >= 1.12 when getting the seed version becomes possible.

PR #169, opened 2019-06-25
Fixes deletion of firewall rules on gcp

### LINKED ISSUE #158: [GCP] Deletion of a shoot cluster deletes all firewall rules in the same VPC and svc account
If two or more shoot clusters share the same VPC and are created by the same service account and then one of these clusters gets deleted all firewall rules for the VPC and service account also get deleted. 

Check:
https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-gcp/pkg/internal/infrastructure/infrastructure.go#L35 
and
https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-gcp/pkg/internal/infrastructure/infrastructure.go#L70

We should add an additional filter including the shoot's name when listing firewall rules for deletion, currently we only use the following prefixes: 

```go
const (
	KubernetesFirewallNamePrefix string = "k8s"
	shootPrefix                  string = "shoot--"
)
```

plkokanov:
This PR adds an additional check when deleting firewall rules which have `k8s` at the start of their name - now the TargetTags are checked if they include the shoot's controlplane namespace before the rules are added for deletion.

Fixes #158 

I've only ran the unit tests to check if this works.

**Release note**:

```improvement operator
Deleting a shoot on GCP will now only delete the firewall rules for that shoot when multiple shoots share the same VPC network.
```

### DIFF
--- a/controllers/provider-gcp/pkg/controller/infrastructure/actuator_delete.go
+++ b/controllers/provider-gcp/pkg/controller/infrastructure/actuator_delete.go
@@ -34,6 +34,7 @@ func (a *actuator) cleanupKubernetesFirewallRules(
 	client gcpclient.Interface,
 	tf *terraformer.Terraformer,
 	account *internal.ServiceAccount,
+	shootSeedNamespace string,
 ) error {
 	state, err := infrastructure.ExtractTerraformState(tf, config)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


rfranzke:
Thanks! As discussed, let's wait with merging until you tested it.

plkokanov:
@rfranzke Tested successfully. Deleting a shoot cluster only deleted firewall rules for that cluster.

rfranzke:
Awesome!

PR #170, opened 2019-06-27
Fixes deletion of routes on GCP

plkokanov:
When deleting a shoot cluster on GCP all routes in the same VPC network as the shoot cluster are also deleted. This PR adds an additional filter when chosing which routes to delete by checking if the route's NextHopInstance contains the shoot's seed namespace

Additional fix for #158. Routes on the same VPC were also affected by single cluster deletion

The names of routes can have equal prefixes if the shoot's seed namespaces are too long. Eg.: instead of `shoot--my-proj--test-shoot1` the prefix for the route name would only be `shoot--my-proj--test-shoot`. This is why I use the `NextHopInstance` url to determine if a route should be deleted.

**Release note**:

```improvement user
Deleting a shoot on GCP will now only delete the routes that shoot when multiple shoots share the same VPC network.
```


plkokanov:
@rfranzke can you have a look?

PR #171, opened 2019-07-01
Fix contains string with prefix bug

mkoynov:
Fixes a possible bug/Makes it easier to read when there is need to ensure an element in a string list with no elements to remove a leading comma.
Removes duplication elements with wrong value for the command line arg "--feature-gates=" in the kube apiserver for packet and alicloud providers.

**Release note**:
```improvement operator
NONE
```

### DIFF
--- a/pkg/webhook/controlplane/utils.go
+++ b/pkg/webhook/controlplane/utils.go
@@ -108,7 +108,11 @@ func EnsureStringWithPrefixContains(items []string, prefix, value, sep string) [
 	if i := StringWithPrefixIndex(items, prefix); i < 0 {
 		items = append(items, prefix+value)
 	} else {
-		values := strings.Split(strings.TrimPrefix(items[i], prefix), sep)
+		valuesList := strings.TrimPrefix(items[i], prefix)
+		var values []string
+		if valuesList != "" {
+			values = strings.Split(valuesList, sep)
+		}
 		if j := StringIndex(values, value); j < 0 {
 			values = append(values, value)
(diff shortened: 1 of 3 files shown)


pkg/webhook/controlplane/utils.go
```diff
@@ -108,7 +108,11 @@ func EnsureStringWithPrefixContains(items []string, prefix, value, sep string) [
 	if i := StringWithPrefixIndex(items, prefix); i < 0 {
 		items = append(items, prefix+value)
 	} else {
-		values := strings.Split(strings.TrimPrefix(items[i], prefix), sep)
+		valuesList := strings.TrimPrefix(items[i], prefix)
+		var values []string
+		if valuesList != "" {
```
timuthy: Does this really fix a bug? If `strings.Split` is called on an empty string the result will be an empty slice. 
If your argument was that there is no necessity to call this on empty values, you could also argue that it's not necessary to call it on a single value, as well.

mkoynov: Most likely not, but it will definitely improve readability. `string.Split` will return a slice with one element - an empty string. If `StringIndex` does not find the passed argument this algorithm will append it to the values slice. This will result in two elements instead of one and the end result will be an element that has a leading comma. [example](https://play.golang.org/p/_Bt8XuUWp8u)

timuthy: Fair enough! Thanks for pointing that out.

PR #178, opened 2019-07-02
Implement operation reconcile annotation

timuthy:
This PR adjusts controllers to only reconcile when `gardener.cloud/operation: reconcile` is set in the respective resources.

Relevant resources are:
- controlplane
- extension
- infrastructure
- operatingsystemconfig
- worker

- Race conditions have been observed when using `ShootFailedPredicate` with a client instance injected by `Controller-Runtime`. Hence, an explicit `WaitForCacheSync` is required.
- Ignore operation annotation was changed to serve as a global parameter for all controllers.
- Watching `clusters` is still necessary to notice if Shoots transfer from `Failed` to any other state. This becomes especially important in deletion cases.

**Release note**:

```noteworthy operator
All extension controllers do now only reconcile their extension resources if the `gardener.cloud/operation=reconcile` annotation is set in the respective resource. Beginning with Gardener version `0.27.0` the gardener-controller-manager will annotate these extension resources when reconciling a shoot. This is to give Gardener control when the extension controllers should act (e.g., only in the shoot's maintenance time window). All extension controllers must implement this contract (see also [this document](https://github.com/gardener/gardener/blob/master/docs/extensions/conventions.md)).
```


pkg/controller/controlplane/mapper.go
```diff
@@ -16,8 +16,8 @@ package controlplane
 
 import (
 	"context"
-
-	extensionscontroller "github.com/gardener/gardener-extensions/pkg/controller"
+	handler2 "github.com/gardener/gardener-extensions/pkg/handler"
```
rfranzke:
```suggestion
	extensionshandler "github.com/gardener/gardener-extensions/pkg/handler"
```

pkg/controller/extension/mapper.go
```diff
@@ -15,7 +15,7 @@
 package extension
 
 import (
-	"github.com/gardener/gardener-extensions/pkg/controller"
+	handler2 "github.com/gardener/gardener-extensions/pkg/handler"
```
rfranzke:
```suggestion
	extensionshandler "github.com/gardener/gardener-extensions/pkg/handler"
```

pkg/controller/infrastructure/mapper.go
```diff
@@ -16,8 +16,8 @@ package infrastructure
 
 import (
 	"context"
-
-	extensionscontroller "github.com/gardener/gardener-extensions/pkg/controller"
+	handler2 "github.com/gardener/gardener-extensions/pkg/handler"
```
rfranzke:
```suggestion
	extensionshandler "github.com/gardener/gardener-extensions/pkg/handler"
```

pkg/controller/worker/mapper.go
```diff
@@ -16,8 +16,8 @@ package worker
 
 import (
 	"context"
-
-	extensionscontroller "github.com/gardener/gardener-extensions/pkg/controller"
+	handler2 "github.com/gardener/gardener-extensions/pkg/handler"
```
rfranzke:
```suggestion
	extensionshandler "github.com/gardener/gardener-extensions/pkg/handler"
```

timuthy:
Thanks for your review @rfranzke. PTAL.

PR #192, opened 2019-07-10
Create Credentialless Cloud-Config for Azure Workers

zanetworker:
Remove Azure credentials from cloud-config for 1.15 as they are not needed anymore. For more informtaion, please refer to: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#azure 

**Release note**:

```noteworthy operator
Azure cloud-config no longer requires credentials. 
```

### DIFF
--- a/controllers/provider-azure/charts/internal/cloud-provider-config/templates/cloud-provider-worker-config.yaml
+++ b/controllers/provider-azure/charts/internal/cloud-provider-config/templates/cloud-provider-worker-config.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cloud-provider-worker-config
+  namespace: {{ .Release.Namespace }}
+data:
+  cloudprovider.conf: |
+    {{- include "cloud-provider-config" .  | indent 4}}
+    {{- if semverCompare "< 1.15" .Values.kubernetesVersion }}
+    {{- include "azure-credentials" .   | indent 4 }}
+    {{- else }}
+    useInstanceMetadata: true
+    {{- end}}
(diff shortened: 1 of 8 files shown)


mvladev:
What about clusters `< v1.15`?

zanetworker:
@mvladev credentials will be used.

controllers/provider-azure/charts/internal/cloud-provider-config/templates/cloud-provider-worker-config.yaml
```diff
@@ -0,0 +1,38 @@
...
+    cloudProviderBackoffJitter: 1.0
+    cloudProviderRateLimit: true
+    cloudProviderRateLimitQPS: 10.0
+    cloudProviderRateLimitBucket: 100
+    cloudProviderRateLimitQPSWrite: 10.0
+    cloudProviderRateLimitBucketWrite: 100
+    {{- if semverCompare ">= 1.14" .Values.kubernetesVersion }}
+    cloudProviderBackoffMode: v2
+    {{- end}}
+    {{- if semverCompare ">= 1.15" .Values.kubernetesVersion }}
```
rfranzke: I suggest to bundle this together with the other if condition above:

```yaml
    {{- if semverCompare "< 1.15" .Values.kubernetesVersion }}
    aadClientId: "{{ .Values.aadClientId }}"
    aadClientSecret: "{{ .Values.aadClientSecret }}"
    tenantId: "{{ .Values.tenantId }}"
    subscriptionId: "{{ .Values.subscriptionId }}"
    {{- else }}
    useInstanceMetadata: true
    {{- end }}
```

Again, the fields for the client id, tenant id, etc. could be moved out into a template and then included here.

PR #194, opened 2019-07-11
enable encrypted flag in storageclass for Alicloud

jia-jerry:
Refer to gardener/gardener#1199. To fix SAP compliance issue.


zanetworker:
@jia-jerry  Why are the controller-registration of other controllers changed?

jia-jerry:
> @jia-jerry Why are the controller-registration of other controllers changed?

This is because we have changed the lates tag. After run /hack/generate.sh, verison of all controllerregistrations will be changed.

PR #211, opened 2019-07-23
Scale down/up cluster-autoscaler instead of deleting it

rfranzke:
Most parts of the code in the generic worker actuator have been copied from the code in Gardener that was written over a year ago and before we started extensibility.

Back then when Gardener handled the machines it could delete the cluster-autoscaler deployment during a rolling update of worker machines because it knew that the task that would re-deploy the cluster-autoscaler would be executed after the rolling update has completed successfully.

Now, with extensibility, the reconciliation of the worker actuator and Gardener might not be in sync. Consequently, it might happen that the worker actuator reconciles and deletes  the cluster-autoscaler, but Gardener does not reconcile in order to bring it back (only later).

Let's improve the code such that the cluster-autoscaler deployment is not deleted by the worker actuator, instead, it's just scaled down and scaled up again at the end (without the necessity to wait for Gardener to do the same).

**Release note**:

```improvement operator
The worker generic actuator is now scaling down and up the cluster-autoscaler instead of deleting the deployment (and relying on Gardener to re-deploy it).
```

### DIFF
--- a/pkg/controller/worker/genericactuator/actuator_reconcile.go
+++ b/pkg/controller/worker/genericactuator/actuator_reconcile.go
@@ -23,7 +23,6 @@ import (
 	extensionscontroller "github.com/gardener/gardener-extensions/pkg/controller"
 	"github.com/gardener/gardener-extensions/pkg/controller/worker"
 	"github.com/gardener/gardener-extensions/pkg/util"
-	"github.com/pkg/errors"
 
 	gardencorev1alpha1 "github.com/gardener/gardener/pkg/apis/core/v1alpha1"
 	gardencorev1alpha1helper "github.com/gardener/gardener/pkg/apis/core/v1alpha1/helper"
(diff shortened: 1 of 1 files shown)


pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -81,12 +85,19 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
...
+				}
+			} else {
+				if err := util.ScaleDeployment(ctx, a.client, deployment, 0); err != nil {
+					return err
+				}
 			}
-			if err := util.WaitUntilResourceDeleted(ctx, a.client, deployment, 5*time.Second); err != nil {
-				return err
+
+			if rollingUpdate {
```
timuthy:
```suggestion
mustScaleUpClusterAutoscaler = rollingUpdate
```

I would also move to either line 84 or 156. Or leave out `mustScaleUpClusterAutoscaler` and use `rollingUpdate` instead.

PR #218, opened 2019-07-26
Update controller-runtime to 0.2.0-beta.2

### LINKED ISSUE #163: Update controller-runtime to 0.2.0-beta.4
We have to update the controller-runtime to `v0.2.0-beta.4` in order to be able to vendor the `master` branch of Gardener again.

There is one larger problem with the `v0.2.0-beta.4` version, namely the removal of the certificate generation and automated webhook registration.

> Of particular note is that automatic certificate generation for webhooks has been removed, and webhooks will no longer self-register. Use controller-tools to generate a webhook configuration, and use something like cert-manager to generate your certificates (if you don't already have a solution). The next release of Kubebuilder will support this out-of-the-box.

We decided the most convenient way to fix this is to move the certificate generation as well as the webhook registration to the Helm chart of the providers themselves. For the local development case we might want to run a script as part of `make start-provider-<type>` that renders this part of the Helm chart and applies the configuration to the local Kubernetes cluster.

/cc @stoyanr

rfranzke:
- Update `sigs.k8s.io/controller-runtime` to `0.2.0-beta.2`.
- Update `k8s.io/*` to `kubernetes-1.14.0`
- Update `gardener/gardener` to `master` branch again. This allowed using a lot of new exported constants from gardener repository.
- Adapt code to cope with new controller-runtime version
- Implement certificate creation and webhook registration as it was removed from the controller-runtime with `0.2.x`.

Fixes #163 

**Special notes for your reviewer(s)**:
- I had to rename all the controllers (replace `-` with `_`) because the metric names are no longer valid with dashes anymore.
- The `--webhook-config-mode`, `--webhook-config-name`, and `--webhook-config-service-selectors` flags have been removed (no longer needed).

**Release note**:

```noteworthy developer
The controller-runtime library has been upgraded to `0.2.0-beta.2`.
```


pkg/webhook/registration.go
```diff
@@ -0,0 +1,229 @@
...
+		Rule: admissionregistrationv1beta1.Rule{
+			APIGroups:   []string{gvk.Group},
+			APIVersions: []string{gvk.Version},
+			Resources:   []string{mapping.Resource.Resource},
+		},
+	}, nil
+}
+
+func buildClientConfigFor(webhook *Webhook, namespace, providerName string, port int, caBundle []byte) admissionregistrationv1beta1.WebhookClientConfig {
+	path := "/" + webhook.Path
```
zanetworker: Maybe use `filepath.Join` ?

pkg/webhook/cmd/options.go
```diff
@@ -146,9 +199,19 @@ func (c *AddToManagerConfig) AddToManager(mgr manager.Manager) error {
 	}
 
 	webhookServer := mgr.GetWebhookServer()
+	webhookServer.CertDir = c.Server.CertDir
+
+	for _, wh := range webhooks {
+		webhookServer.Register("/"+wh.Name, wh.Webhook)
```
zanetworker: Maybe use `filepath.Join`

Makefile
```diff
@@ -111,69 +112,75 @@ start-os-ubuntu:
...
+	@LEADER_ELECTION_NAMESPACE=garden WEBHOOK_CONFIG_URL=$(WEBHOOK_CONFIG_URL) \
+	go run \
 		-ldflags $(LD_FLAGS) \
 		./controllers/provider-aws/cmd/gardener-extension-provider-aws \
 		--config-file=./controllers/provider-aws/example/00-componentconfig.yaml \
 		--ignore-operation-annotation=$(IGNORE_OPERATION_ANNOTATION) \
 		--leader-election=$(LEADER_ELECTION) \
-		--webhook-server-host=0.0.0.0 \
-		--webhook-server-port=8443
+		--webhook-config-server-host=0.0.0.0 \
```
zanetworker: Maybe use '--webhook-config-server-url` instead of an environment variable?

Makefile
```diff
@@ -111,69 +112,75 @@ start-os-ubuntu:
...
+		--webhook-config-server-port=8443
 
 .PHONY: start-provider-azure
 start-provider-azure:
-	@LEADER_ELECTION_NAMESPACE=garden go run \
+	@LEADER_ELECTION_NAMESPACE=garden WEBHOOK_CONFIG_URL=$(WEBHOOK_CONFIG_URL) \
```
zanetworker: Maybe use '--webhook-config-server-url` instead of an environment variable?

Makefile
```diff
@@ -111,69 +112,75 @@ start-os-ubuntu:
...
 		--leader-election=$(LEADER_ELECTION) \
-		--webhook-server-host=0.0.0.0 \
-		--webhook-server-port=8443
+		--webhook-config-server-host=0.0.0.0 \
+		--webhook-config-server-port=8443
 
 .PHONY: start-provider-openstack
 start-provider-openstack:
-	@LEADER_ELECTION_NAMESPACE=garden go run \
+	@LEADER_ELECTION_NAMESPACE=garden WEBHOOK_CONFIG_URL=$(WEBHOOK_CONFIG_URL) \
```
zanetworker: Maybe use '--webhook-config-server-url` instead of an environment variable?

Makefile
```diff
@@ -111,69 +112,75 @@ start-os-ubuntu:
...
 		--leader-election=$(LEADER_ELECTION) \
-		--webhook-server-host=0.0.0.0 \
-		--webhook-server-port=8443
+		--webhook-config-server-host=0.0.0.0 \
+		--webhook-config-server-port=8443
 
 .PHONY: start-provider-alicloud
 start-provider-alicloud:
-	@LEADER_ELECTION_NAMESPACE=garden go run \
+	@LEADER_ELECTION_NAMESPACE=garden WEBHOOK_CONFIG_URL=$(WEBHOOK_CONFIG_URL) \
```
zanetworker: Maybe use '--webhook-config-server-url` instead of an environment variable?

Makefile
```diff
@@ -111,69 +112,75 @@ start-os-ubuntu:
...
 		--leader-election=$(LEADER_ELECTION) \
-		--webhook-server-host=0.0.0.0 \
-		--webhook-server-port=8443
+		--webhook-config-server-host=0.0.0.0 \
+		--webhook-config-server-port=8443
 
 .PHONY: start-provider-packet
 start-provider-packet:
-	@LEADER_ELECTION_NAMESPACE=garden go run \
+	@LEADER_ELECTION_NAMESPACE=garden WEBHOOK_CONFIG_URL=$(WEBHOOK_CONFIG_URL) \
```
zanetworker: Maybe use '--webhook-config-server-url` instead of an environment variable?

pkg/controller/cmd/options.go
```diff
@@ -35,9 +35,9 @@ const (
 	// LeaderElectionNamespaceFlag is the name of the command line flag to specify the leader election namespace.
 	LeaderElectionNamespaceFlag = "leader-election-namespace"
 	// WebhookServerHostFlag is the name of the command line flag to specify the webhook config host for 'url' mode.
-	WebhookServerHostFlag = "webhook-server-host"
+	WebhookServerHostFlag = "webhook-config-server-host"
 	// WebhookServerPortFlag is the name of the command line flag to specify the webhook server port.
-	WebhookServerPortFlag = "webhook-server-port"
+	WebhookServerPortFlag = "webhook-config-server-port"
```
zanetworker: Maybe add `WebhookServerURLFlag = "webhook-config-server-url"`?

rfranzke:
> I liked `--webhook-config-mode` much more since it is easy to use.

The problem is that even with this flag we would still need `--webhook-config-server-url` in order to properly register the webhooks for the environment (we don't know the URL and can't make any assumptions). That's why I removed the `mode` flag. 

> Furthermore, it's opaque that by specifying `--webhook-config-server-url` the service mode is automatically disregarded.

When using a URL (or, "url mode"), a `Service` is not needed (that's why it's no longer created automatically but part of the Helm charts now).

What is your suggestion? Renaming `--webhook-config-server-url` to make it more clear (if so, to what)? Or do you suggest to reintroduce `--webhook-config-mode`? If set to `service` then the service names are part of the SANs of the certificate, if set to `url` then the URL provided by `--webhook-config-server-url` is the SAN?

timuthy:
> What is your suggestion? Renaming --webhook-config-server-url to make it more clear (if so, to what)? Or do you suggest to reintroduce --webhook-config-mode? If set to service then the service names are part of the SANs of the certificate, if set to url then the URL provided by --webhook-config-server-url is the SAN?

I thought of using `--webhook-config-mode` and for the CN/SAN you could reuse the information from `--webhook-config-host` which is also the way the 'old' implementation did it, isn't it?

rfranzke:
`--webhook-config-host` is the host that is used for advertising the HTTP server. It is different from the URL that is used to register the webhooks (host is usally `0.0.0.0`).

pkg/webhook/registration.go
rfranzke: This is not a file path, it's a URL. Good enough imho.

pkg/webhook/cmd/options.go
rfranzke: This is not a file path, it's a URL. Good enough imho.

PR #223, opened 2019-07-30
Implement support for OSC customization

pablochacin:
Gardener allows the specification of per-worker set customization
of OSC. This customization may contain additional files and units
to be added to the cloudconfig.

When reconciling the OSC, the actuator retrieves the
customization and adds to the cloudconfig before passing
it to the generator.

Fixes #

This PR has not yet been tested due to issues setting a development environment.

**Release note**:

```improvement operator
Additional units and files defined in the OSC configuration customizations for a worker pool will be applied to the cloudconfig for worker nodes. 
```

### DIFF
--- a/pkg/controller/operatingsystemconfig/oscommon/actuator/actuator_types.go
+++ b/pkg/controller/operatingsystemconfig/oscommon/actuator/actuator_types.go
@@ -1,0 +1,35 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package actuator
+
+import (
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type OperatingSystemCustomization struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Units is a list of unit for the operating system configuration (usually, a systemd unit).
+	// +patchMergeKey=name
+	// +patchStrategy=merge
+	// +optional
+	Units []extensionsv1alpha1.Unit `json:"units,omitempty"`
+	// Files is a list of files that should get written to the host's file system.
+	// +optional
+	Files []extensionsv1alpha1.File `json:"files,omitempty"`
+}
--- a/pkg/controller/operatingsystemconfig/oscommon/actuator/actuator_util.go
+++ b/pkg/controller/operatingsystemconfig/oscommon/actuator/actuator_util.go
@@ -33,19 +33,25 @@ import (
 
 	"github.com/gardener/gardener-extensions/pkg/controller/operatingsystemconfig/oscommon/cloudinit"
 	commonosgenerator "github.com/gardener/gardener-extensions/pkg/controller/operatingsystemconfig/oscommon/generator"
-
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
-
+	yaml "gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
-
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // CloudConfigFromOperatingSystemConfig generates a CloudConfig from an OperatingSystemConfig
 // using a Generator
 func CloudConfigFromOperatingSystemConfig(ctx context.Context, cli runtimeclient.Client, config *extensionsv1alpha1.OperatingSystemConfig, generator commonosgenerator.Generator) ([]byte, *string, error) {
+	var customization = OperatingSystemCustomization{}
+	if config.Spec.ProviderConfig != nil {
+		err := yaml.Unmarshal(config.Spec.ProviderConfig.Raw, &customization)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+
 	files := make([]*commonosgenerator.File, 0, len(config.Spec.Files))
-	for _, file := range config.Spec.Files {
+	for _, file := range append(config.Spec.Files, customization.Files...) {
 		data, err := DataForFileContent(ctx, cli, config.Namespace, &file.Content)
 		if err != nil {
 			return nil, nil, err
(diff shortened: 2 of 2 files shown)


rfranzke:
Hi Pablo, thanks for the PR, but I'm confused - that's not what we discussed.

Similar to how the AWS (or any other provider) has a specific infrastructure configuration the JeOS (or any other OS controller) could have a specific OS configuration.

```yaml
apiVersion: suse-jeos.operatingsystemconfig.extensions.gardener.cloud/v1alpha1
kind: Customization
someSpecific: flag
whatEver: setting
```

What was built now is not like that. It's a way to generally merge units/files into the existing config, but maybe some providers don't want to support that? Also, end-users would need to declare all these units/files they want to merge - also not desirable at the moment.

pablochacin:
@rfranzke Not sure If I'm following. This PR is basically the `ProviderConfig` that [was introduced in the OSC](https://github.com/gardener/gardener/pull/1261#issue-301129155) and using it to customize the OSC provided by Gardener. 

This logic is implemented in the `oscommon` library which was introduced as a helper for OSC controllers that only need a template for translating the `CloudConfig', as JeOS. This is because JeOS controller itself doesn't have any logic besides translating the CloudConfig usig a template. 

Not sure what you mean by `some providers don't want to support it`. Same will apply to any other feature in the library: if you want a different behavior, you can implement the controller without using the library. The library is not mandatory. 

That said, we could add some flag for enabling the feature at the controller level.

rfranzke:
> After writing the comment above I realized maybe your concern is about allowing arbitrary cloudconfig units to be defined by users, and you would prefer more restricted options that could then be mapped to cloudconfig, right?

Exactly, yes.

> If that is the case, I need to figure out how to do this at the controller level, as the oscommon-based controllers has little logic, as they rely the common logic for processing cloudconfig and only provide the template for the final transformation.

Yes, please. If we would like to support that simply additional units or files could be added to the OSC then we could do that directly in Gardener (as Gardener defines the extensions API and creates the OSC, hence, it could easily merge the user-defined units/files with its defaults). For this use-case we don't need a provider config that is controlled by the actual extension controllers.

Please open up the oscommon library such that actuators like suse-jeos can have their own, suse-jeos specific provider config.

Thank you very much!

pablochacin:
>  If we would like to support that simply additional units or files could be added to the OSC then we could do that directly in Gardener (as Gardener defines the extensions API and creates the OSC

This seems to be the case: what we have been discussing regarding OSC customization is rather open, so instead of keep adding more and more options and flags to the JeOS' customization, I think it makes more sense to be additional units and files.

rfranzke:
> This seems to be the case: what we have been discussing regarding OSC customization is rather open, so instead of keep adding more and more options and flags to the JeOS' customization, I think it makes more sense to be additional units and files.

I don't this makes sense at this point in time. End-users would need to define the very same units/files again and again for every shoot. Yes, that's not different from defining the same provider config for every shoot, but looking at the usability aspect I guess well-defined options/flags to customize the JeOS config would improve the user experience.

pablochacin:
>  I guess well-defined options/flags to customize the JeOS config would improve the user experience.

The point is that the options are not for JeOS itself, but for setting the environment for the worker pools or shoots and are application or user dependant. Therefore, IMHO it makes no sense to translate every possible user requirement to the JeOS operator.

rfranzke:
> Therefore, IMHO it makes no sense to translate every possible user requirement to the JeOS operator.

Agreed, but what I understood from the initial discussion was that HANA/HaaS needs some very special customization that most likely no other users need. Did this requirement change? What are the concrete units/files that HANA/HaaS needs?

pablochacin:
> Agreed, but what I understood from the initial discussion was that HANA/HaaS needs some very special customization that most likely no other users need. Did this requirement change

The initial idea came before the HANA/HaaS needs you mention. Since the very initial implementation of the JeOS controller I proposed this approach due to different requirements from other groups in SAP that needed some customization (I don't recall exactly which).

Later, there were some specific changes that HANA needed (and therefore are application dependant) and more recently other new changes appeared (setting the hostname correctly at boot time in AWS environments)

As you see, the idea of having specific options for each case seems not to be the best approach, as the changes are specific of some use case or platform and requiring changes in the controller seams not correct to me.

rfranzke:
> (setting the hostname correctly at boot time in AWS environments)

This is nothing that should be solved such a provider config mechanism. The AWS provider should do that if it's needed on AWS (similar like [this](https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-gcp/pkg/webhook/controlplane/ensurer.go#L172)).

> As you see, the idea of having specific options for each case seems not to be the best approach, as the changes are specific of some use case or platform and requiring changes in the controller seams not correct to me.

Still, I don't like the approach too much. It requires that the user knowledge about internal Gardener APIs and knowledge about what Gardener already sets up/configures, etc. It's not very end-user friendly.

pablochacin:
> APIs and knowledge about what Gardener already sets up/configures, etc. It's not very end-user friendly.

I partially agree that is not the most user friendly. However, the only knowledge required is about using cloud-init like units, what is not gardener-specific.

I'm afraid that the alternative is to define an abstraction on top of this, which would be specific of the JeOS controller, when the problem is in my opinion, more generic.

rfranzke:
> However, the only knowledge required is about using cloud-init like units, what is not gardener-specific.

It's not Gardener specific, but most users don't (want to) know this stuff, I assume. They simply want to set a flag "prepareMyClusterForHANA=true` (or whatever, maybe even more fine-grained to enable/disable specific settings - I don't know the exact requirements).

I understand that some users might want to inject complete new files/units, but as said, if we want to enable this then it should be part of the Shoot API without any provider config in between.

> I'm afraid that the alternative is to define an abstraction on top of this, which would be specific of the JeOS controller, when the problem is in my opinion, more generic.

Abstractions are not always good, but sometimes they are. They simplify the usability. Take https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-aws/example/30-infrastructure.yaml#L42-L53 as a similar example - the provider config of the `Infrastructure` CRD for AWS. Following your arguments the better approach would have been to allow the end-users to define custom Terraform instructions here instead of using this easy-to-understand provider config API. That's not what we want(ed). Gardener has an opinion on how shoots are created, and we don't want users to tamper with it.
The worker nodes are slightly different because they are "owned" by the end-users and run in their account.

Still, I favour the provider config approach with a defined API. However, if end-users (in this case HANA) prefer to define whole new units and files in their shoot resources instead of using an abstraction - fine, I can live with that. Can you maybe also get their feedback in the meantime?

Of course, your arguments are totally valid. We would not need to have implementation effort for each and every end-user requirement, true. But, with a provider config API we can make sure that the clusters are healthy because we only support such configuration that doesn't break anything. With being fully open to accept additional units/files end-users can (more) easily break their clusters, resulting in ops effort and alerts.

@timuthy @zanetworker @vlerenc opinions on your end?

timuthy:
> But, with a provider config API we can make sure that the clusters are healthy because we only support such configuration that doesn't break anything. With being fully open to accept additional units/files end-users can (more) easily break their clusters, resulting in ops effort and alerts.

Is definitely an argument that makes me vote for the provider config API.

pablochacin:
> The worker nodes are slightly different because they are "owned" by the end-users and run in their account

That was the main reason for not been restrictive regarding what can or cannot be configured.

> Abstractions are not always good, but sometimes they are.
Certainly, but as I said before, based on the different requirements we have discussed, the abstractions should be things like installing systemctl services (which in some cases will require creating a file with the service's script) and running commands at boot time, which are certainly close to using cloudconfig unis. I don't particularly see any advantage on abstracting this out, but can be done.

rfranzke:
> That was the main reason for not been restrictive regarding what can or cannot be configured.

As said above, there are various reasons to be restrictive. The main reasons are stability/SLOs.
The above example regarding additional Terraform configurations is still a good one, but there is also another one: We don't expose all flags for the control plane and the kubelet in the shoot spec, only those we carefully selected beforehand. Sure, with some of the currently exposed flags end-users might still break their clusters, however, at least we can validate the values and prevent them from doing so, providing some help, etc.

For the sake of consistency I'd follow the same motivation/principles also in the OSC customization requirements.
Can we continue with the provider config approach now?

pablochacin:
> Can we continue with the provider config approach now?

Do we have the exact requirements from Hana users?

pablochacin:
Superceeded by https://github.com/gardener/gardener-extensions/pull/248

PR #226, opened 2019-07-31
OSC controller now listens for secret events

mvladev:
Introduced a new fieldselector `secretRef.name` which returns`OperatingSystemConfigurations` which have a specific secret in either their specification or their status.

If the secret generated by the controller is removed then nothing reconciles it back.

Fixes # n/a

**Release note**:

```improvement operator
fix operating-system-controller to react on related secret changes.
```

### DIFF
--- a/pkg/controller/operatingsystemconfig/controller.go
+++ b/pkg/controller/operatingsystemconfig/controller.go
@@ -15,13 +15,21 @@
 package operatingsystemconfig
 
 import (
+	"context"
+
 	extensionspredicate "github.com/gardener/gardener-extensions/pkg/predicate"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
-
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
@@ -33,6 +41,8 @@ const (
 	ControllerName = "operatingsystemconfig_controller"
 
 	name = "operatingsystemconfig-controller"
+
+	specFilesSecretFieldKey = "secretRef.name"
 )
 
 // AddArgs are arguments for adding an operatingsystemconfig controller to a manager.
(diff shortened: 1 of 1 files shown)


timuthy:
> If the secret generated by the controller is removed then nothing reconciles it back.

I want to understand your use-case: What process removes that secret?

mvladev:
`kubectl delete secret osc-pool-1`.

Or any other logic which accidentally removes / edits the generated secret

mvladev:
Another thing which this PR enable is to optimize the controller - atm if one of the secrets in `spec.files.content.secretRef` is missing the controller throws errors constantly:

```
{"level":"error","ts":"2019-07-31T13:30:08.688+0300","logger":"controller-runtime.controller","msg":"Reconciler error","controller":"operatingsystemconfig_controller","request":"default/pool-01-original","error":"could not generate cloud config: Secret \"some-referenced-secret\" not found"
```

which is really bad, because the reconciled object is being put into a rate-limited queue.

What needs to happen in such cases is to set the status to error and simply stop with the reconciliation. When the secret is created eventually, it'll trigger the reconcile loop.

timuthy:
> Or any other logic which accidentally removes / edits the generated secret

The extension controllers used to have additional watches for any "what happens if someone deletes/edits something" use-cases. With #178 we decided to simplify watches and predicates for every controller because Gardener dictates when reconciliation is suppose to happen (makes it possible to only reconcile during shoot maintenance).

So we should decide whether an additional watch on `Secrets` is really critical or if it's good enough to let Gardener reconcile to eventually reconcile the extensions again. If an administrator accidentally deletes the secret, he can also trigger the extension reconciliation by `gardener.cloud/operation: reconcile`.

Today you add this watch for the `OSC` controller and tomorrow you'll end up with a bunch of other watches on different other controllers again.

mvladev:
> With #178 we decided to simplify watches and predicates for every controller because Gardener dictates when reconciliation is suppose to happen (makes it possible to only reconcile during shoot maintenance).

The purpose of all controllers in K8S is to **keep** the system in a desired state and not to keep the system the system at desired state at **specific time**. #178 is making the entire system brittle and prone to failure when the higher actor (gardener-controller-manager) is not working or misbehaving. 

> If an administrator accidentally deletes the secret, he can also trigger the extension reconciliation by gardener.cloud/operation: reconcile.

Which proves my point - why a manual intervention is required, when the system can heal itself?

timuthy:
> The purpose of all controllers in K8S is to keep the system in a desired state and not to keep the system at desired state at specific time. #178 is making the entire system brittle and prone to failure when the higher actor (gardener-controller-manager) is not working or misbehaving.

I totally get your point here and I also agree with you from a theoretical point of view. After all I'm concerned about two points:

1. If we'd follow your pattern to keep the system at the desired anytime, all of our extension controllers needed to have so many active watches, actually on every resource type they've ever created. Take the certificate service as an example: It creates `Deployments`, `ServiceAccounts`, `ClusterRoles`, `ClusterRolebindings`, `CustomResourceDefinitions`, `ClusterIssuers` and `Secrets`. If I got your concern correctly, it means that the Certificate Service Controller needs at least seven different watches to ensure the desired state at all times.

How is that suppose to work at runtime as well as from maintenance perspective for all of our extension controllers?

2. In operations we experienced that we run into rate limits of our supported cloud providers again and again (volume operations, LB operations, etc.). As a consequence, certain shoot clusters were unavailable for a longer period.

Wouldn't you also consider this as "_...making the entire system brittle and prone to failure..._" ?

We need to restrict these **expensive operations**, hence distribute them over a larger period --> reconciliation in maintenance only.

What would be your design approach that addresses your doubts as well as tackles the rate limit issues?

mvladev:
>> If we'd follow your pattern to keep the system at the desired anytime, all of our extension controllers needed to have so many active watches, actually on every resource type they've ever created. Take the certificate service as an example: It creates Deployments, ServiceAccounts, ClusterRoles, ClusterRolebindings, CustomResourceDefinitions, ClusterIssuers and Secrets. If I got your concern correctly, it means that the Certificate Service Controller needs at least seven different watches to ensure the desired state at all times.
>
> How is that suppose to work at runtime as well as from maintenance perspective for all of our extension controllers?

If "X" different resources are created then you are left with 2 choices - open a watch for those individual resources or use some other custom-resource which accepts templates (for all those resources above) and then just watch for it.

>> In operations we experienced that we run into rate limits of our supported cloud providers again and again (volume operations, LB operations, etc.). As a consequence, certain shoot clusters were unavailable for a longer period.
>
> Wouldn't you also consider this as "...making the entire system brittle and prone to failure..." ?

Those are transient errors - eventually those operations will succeed. The entire K8S is eventually consistent - now you might not have enough Nodes to run your workload, but after some time this might change.

> We need to restrict these expensive operations

100% agreed

> hence distribute them over a larger period --> reconciliation in maintenance only.

Or we can design our controllers to execute remote calls only when needed for example:

https://github.com/gardener/gardener-extensions/blob/969ee4adcfed3e2db7286748d2d093de8665f53f/controllers/provider-aws/pkg/controller/infrastructure/actuator_reconcile.go#L116-L120

Here we fetch the IGW id, but this ID is not persisted anywhere in the system, so we are forced to make the same call in some other reconciliation loop.

Bellow is the same problem - it always tries to create the bucket even if this BackupBucket has been created before: 

https://github.com/gardener/gardener-extensions/blob/6f30b7774a8f0d5521c52d5aa07097f9377f40b8/controllers/provider-aws/pkg/controller/backupbucket/actuator.go#L46-L53

timuthy:
> If "X" different resources are created then you are left with 2 choices - open a watch for those individual resources or use some other custom-resource which accepts templates (for all those resources above) and then just watch for it.

Do you have an example for the template concept? It rather sounds like deferring the problem because in the end there must be some kind of k8s resource, e.g. `Deployment`.

> Or we can design our controllers to execute remote calls only when needed for example:

This applies to actions we execute in our controllers which would be fine.

Image another case:
The `etcd sts .spec` is changed which makes the Pods to be recreated and as a consequence there are a lot of disk attach and detach operations happening --> run into rate limit. Since these operations are initiated by the `Kube-Controller-Manager` Gardener has no means to slow down or do an ad-hoc intervention. But at the same time the Shoot cluster is down and w/o a doubt is the worst thing that can happen even if "_...eventually those operations will succeed_".

Sure, we could think of another etcd concept. In the end I think the reconciliation in maintenance only is the most effective solution we cloud come up, atm.

rfranzke:
I agree with @timuthy, the use-case you explained is something that is not supported by Gardener or any extension controller at all. Operators that manually delete/tweak stuff in the seed should a) not do that in the first place, and b) cannot expect that everything is reset to Gardener's desired state immediately.

It was a deliberate decision to not implement all these `WATCH`es for the resources deployed to the seed. The argument back then was that only operators/administrators have access to the seed and no end-user can (accidentally) change/tweak anything, so there is no need for this additional network i/o, cpu/memory consumption.

Just thinking it through - of course, you have a valid point that controllers should always reconcile back the world to their desired state, but in our case this wouldn't bring us many benefits. If we would consequently implement this then not only all the extensions would have to watch every resource they ever create (see example of @timuthy), but also Gardener would have to watch almost every resource that is available in the Kubernetes API. And why? Just because administrators/operators tweak stuff manually in the seed what they shouldn't do anyways in the first place? I guess you see that this doesn't make much sense.

Originally, I thought this PR originated from a bug/problem you found when implementing support for machine images per worker pool in Gardener together with @KristianZH. You said when switching the OS the OSC controllers are not reconciling anymore? Do we really have this problem? If yes, this PR would probably not solve it, correct? Is there anything else that justifies such inconsistent change (inconsistent because only the OSC is watching its secrets, but no other controller does as described above).

zanetworker:
@mvladev any update based on what @rfranzke said?

PR #235, opened 2019-08-01
Add shoot webhooks library and implement for AWS + Packet

rfranzke:
Some providers might not only need to mutate resources deployed to the seed but also those deployed to the shoot. Examples are AWS which needs to modify the nginx-ingress configuration, Packet which needs to inject an init container into the vpn-shoot deployment, and soon OpenStack which might need to annotate the vpn-shoot service.

In order to get rid of https://github.com/gardener/gardener/blob/master/pkg/operation/cloudbotanist/types.go#L29-L32 we are now providing an easy-to-use library which registers webhooks in the shoots. The mutators only need to implement their business logic and can directly operate on the objects that need to be mutated.

The way it works is the following: All extension controllers are already running in the seed clusters, i.e., the same cluster where the kube-apiserver of the shoots is running. When a provider extension registers a shoot webhook then a `NetworkPolicy` gets created that allows communication from the respective shoot apiserver to the extension controller. This network policy is deployed by the `ControlPlane` controller. Additionally, it deploys a `ManagedResource` which contains the `MutatingWebhookConfiguration` that is to be created in the shoot cluster.
The kube-apiserver of the shoot can cluster-internally talk to the extension provider.

- @timuthy @zanetworker Along the way I moved some of the webhook functions outside of the `controlplane` package to allow reusing them also in other webhooks types (e.g., `shoot` or soon the new `networking` type). Also, the webhook server certificates are not regenerated after every restart. Instead, they are stored in a Kubernetes secret and read from there when the extension controllers are restarting.
- @afritzler @mandelsoft Take a look at https://github.com/gardener/gardener-extensions/commit/bd57e92311a3d658dcd533caf2485faf84d500f7 or https://github.com/gardener/gardener-extensions/commit/df05e5d2a31b74abf7227bfdd7ed73c38e930ce8 to see what needs to be implemented in order to annotate the vpn-shoot service.

**Release note**:

```noteworthy developer
The webhook package does now provide means to register webhooks for shoot clusters.
```


controllers/provider-packet/pkg/cmd/options.go
```diff
@@ -44,5 +46,6 @@ func WebhookSwitchOptions() *webhookcmd.SwitchOptions {
 		webhookcmd.Switch(extensioncontrolplanewebhook.WebhookName, controlplanewebhook.AddToManager),
 		webhookcmd.Switch(extensioncontrolplanewebhook.ExposureWebhookName, controlplaneexposurewebhook.AddToManager),
 		webhookcmd.Switch(extensioncontrolplanewebhook.BackupWebhookName, controlplanebackupwebhook.AddToManager),
+		webhookcmd.Switch(extensionshootwebhook.WebhookName, shootwebhook.AddToManager),
```
zanetworker: maybe move this up under the other webhook?

controllers/provider-packet/pkg/webhook/shoot/mutator.go
```diff
@@ -41,6 +41,7 @@ func (m *mutator) Mutate(ctx context.Context, obj runtime.Object) error {
 	case *appsv1.Deployment:
 		switch x.Name {
 		case "vpn-shoot":
+			logger.Info("Mutating resource", "kind", x.Kind, "namespace", x.Namespace, "name", x.Name)
```
zanetworker: maybe use `logMutation` here as well?

pkg/webhook/controlplane/genericmutator/mutator.go
```diff
@@ -103,20 +103,25 @@ func (m *mutator) Mutate(ctx context.Context, obj runtime.Object) error {
 	case *corev1.Service:
 		switch x.Name {
 		case gardencorev1alpha1.DeploymentNameKubeAPIServer:
+			logMutation(m.logger, x.Kind, x.Namespace, x.Name)
 			return m.ensurer.EnsureKubeAPIServerService(ctx, x)
 		}
 	case *appsv1.Deployment:
 		switch x.Name {
 		case gardencorev1alpha1.DeploymentNameKubeAPIServer:
+			logMutation(m.logger, x.Kind, x.Namespace, x.Name)
```
zanetworker: maybe use `logMutation` here as well?

zanetworker (commented):
Nice PR :)

pkg/webhook/cmd/options.go
```diff
@@ -214,6 +214,8 @@ type AddToManagerConfig struct {
 // AddToManager instantiates all webhooks of this configuration. If there are any webhooks, it creates a
 // webhook server, registers the webhooks and adds the server to the manager. Otherwise, it is a no-op.
 func (c *AddToManagerConfig) AddToManager(mgr manager.Manager) error {
+	ctx := context.Background()
```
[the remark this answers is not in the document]
rfranzke: At this point in time we don't have any context yet. The manager is not yet started, so we have to first create a background context.

controllers/provider-packet/pkg/cmd/options.go
rfranzke: What do you mean? It is already in the `WebhookSwitchOptions()` functions?!

zanetworker (approved):
/lgtm

PR #243, opened 2019-08-02
Add control plane exposure controller

### LINKED ISSUE #233: Control plane exposure controller
We have introduced a `ControlPlane` CRD whose `spec` contains a `type` field for the respective shoot provider. It's purpose is to install any provider-specific components into the seed (or shoot) that are required to make the cluster functional. Examples of those components are the cloud-controller-manager, CSI controllers/plugins, or storage classes.

Some providers have additional needs: They might need to deploy components that are required to expose the control plane. The kube-apiserver of a shoot is exposed via a load balancer in the seed. The seed provider might be different from the shoot provider (for example, you can create an AWS shoot whose control plane is running on a GCP seed). Examples for those components are the [`aws-lb-readvertiser`](https://github.com/gardener/aws-lb-readvertiser). While most seed providers don't need to do anything here some might have this requirement.

The plan is to enhance the current `ControlPlane` resource (instead of creating an entirely new CRD named `ControlPlaneExposure`). Let's introduce a new `.spec.purpose` field whose possible values are `{exposure,normal}` (naming suggestions/improvements are welcome):

```yaml
apiVersion: extensions.gardener.cloud/v1alpha1
kind: ControlPlane
metadata:
  ...
spec:
  type: aws
  purpose: exposure
  ...
```

When seeing such a resource the AWS control plane exposure controller shall deploy the `aws-lb-readvertiser`.

The resource doesn't need to have any infrastructure information (because the infrastructure is created for the shoot cloud provider).

svetlinas:
Adds functionality that cloud providers may use in case an exposure of the shoot control plane is needed

Fixes #233

**Release note**:

```improvement operator
All provider extensions now support `ControlPlane` CRDs of with `.spec.type=exposure`.
```


controllers/provider-aws/pkg/controller/controlplane/valuesprovider.go
```diff
@@ -188,6 +188,28 @@ func (vp *valuesProvider) GetStorageClassesChartValues(
...
+	cluster *extensionscontroller.Cluster,
+) (map[string]interface{}, error) {
+
+	replicas := 0
+	if extensionscontroller.IsHibernated(cluster.Shoot) {
+		replicas = 1
+	}
+
+	return map[string]interface{}{
+		"domain":   "", // TODO loadBalancerIngress is computed after kube-apiserver service is ready and it is assigned to botanist.operation.APIServerAddress
```
rfranzke: You could use [this function](https://github.com/gardener/gardener-extensions/blob/ba20cf1c9d2aeb21c38bdc8606f0cbdfc4298120/controllers/provider-alicloud/pkg/webhook/controlplaneexposure/ensurer.go#L60-L64).

controllers/provider-aws/example/kube-apiserver.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v1
```
rfranzke: Isn't https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-aws/example/10-fake-shoot-controlplane.yaml#L84-L99 good enough already?

pkg/controller/controlplane/genericactuator/actuator.go
```diff
@@ -155,6 +160,53 @@ func (a *actuator) Reconcile(
...
+func (a *actuator) reconcileControlPlaneExposure(
+	ctx context.Context,
+	cp *extensionsv1alpha1.ControlPlane,
+	cluster *extensionscontroller.Cluster,
+) (bool, error) {
+	if a.controlPlaneExposureChart == nil {
+		return false, nil
+	}
+
+	// Deploy secrets
```
rfranzke: Probably it's safer to also do this if `exposureSecrets != nil`. Some actuators might not need them.

pkg/controller/controlplane/genericactuator/actuator.go
```diff
@@ -268,12 +320,47 @@ func (a *actuator) Reconcile(
...
+	cp *extensionsv1alpha1.ControlPlane,
+	cluster *extensionscontroller.Cluster,
+) error {
+	// Delete control plane objects
+	a.logger.Info("Deleting control plane exposure with objects", "controlplane", util.ObjectName(cp))
+	if err := a.controlPlaneExposureChart.Delete(ctx, a.client, cp.Namespace); err != nil {
+		return errors.Wrapf(err, "could not delete control plane exposure objects for controlplane '%s'", util.ObjectName(cp))
+	}
+
+	// Delete secrets
```
rfranzke: Same `if` condition suggestion here.

controllers/provider-aws/example/kube-apiserver.yaml
svetlinas: Sorry, I didn't need to upload this file. I will remove it.

pkg/controller/controlplane/genericactuator/actuator.go
svetlinas: I will fix it

Issue #135, opened 2018-04-11
Consider Azure AZs


vlerenc:
We do not support AZs for Azure, but there seem to be some now: https://azure.microsoft.com/de-de/blog/introducing-azure-availability-zones-for-resiliency-and-high-availability/ -> https://azure.microsoft.com/en-us/global-infrastructure/availability-zones
What are the advantages of those AZs and should we support them (instead of the update and fault domains)?

dkistner:
**Current Status**
- [x] 1. Move to Standard LoadBalancers
- [ ] 2. New `core.gardener.cloud/v1alpha1.Shoot` resource is implemented to allow zone assignments per workerpool on Azure
- [ ] 3. MCM supports Azure Availability Zones https://github.com/gardener/machine-controller-manager/issues/332
- [ ] 4. Azure Provider extension handle `infrastructure`, `controlplane` and `worker` CRs properly to support zones (depends partly on 2.) 
- [ ] 5. Gardener creates proper CRs in the Seed which the Azure provider extension can react on (depends on 2.)

PR #246, opened 2019-08-06
Azure: Move to standard loadbalancer

### LINKED ISSUE #207: Should we switch to standard SKU load balancers in Azure?
We are using basic SKU load balancers for our Shoots in Azure, however, Azure supports also standard SKU load balancers which provide more features: https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-standard-overview

Should we switch to them?

Hint: To use this feature in Kubernetes we require version 1.11 for the Seed clusters, and must set `{"loadBalancerSku": "standard"}` in the Azure cloud provider config. See https://github.com/kubernetes/kubernetes/pull/61884 and https://github.com/kubernetes/kubernetes/pull/62707.

dkistner:
As prerequisite to support Availability Zones on Azure (#135) we need to move to LoadBalancers with the `standard` SKU instead of the `basic` SKU, because only this type of LoadBalancer supports load balancing across zones.

Fixes #207

Update the `ControllerRegistration` for the Azure Provider extension to use a custom build image which contains this PR (dominickistner/gardener-extension-hyper:0.11.0-az-standard-lb).
After the Azure extensions on the Seed cluster are updated, create an Azure Shoot and check after creation has been completed, if the LoadBalancer is of sku `standard`.

To test the upgrade procedure. Create upfront an Azure Shoot with basic LoadBalancer SKU.
Then upgrade the `ControllerRegistration` as above described and follow the instruction in the guide controllers/provider-azure/docs/migrate-loadbalancer.md

**Release note**:

```improvement user
Gardener supports now load balancers with standard sku. New cluster use automatically standard sku load balancers. Existing clusters remain with basic sku load balancers until they are manually migrated.
```
```action user
Gardener supports now load balancers with standard sku. To migrate the LoadBalancer from the basic sku to standard sku follow the instructions in https://github.com/gardener/gardener-extensions/tree/master/controllers/provider-azure/docs/migrate-loadbalancer.md
```

### DIFF
--- a/controllers/provider-azure/docs/migrate-loadbalancer.md
+++ b/controllers/provider-azure/docs/migrate-loadbalancer.md
@@ -1,0 +1,89 @@
+# Migrate Azure Shoot Load Balancer from basic to standard SKU
+
+This guide descibes how to migrate the Load Balancer of an Azure Shoot cluster from the basic SKU to the standard SKU.<br/>
+**Be aware:** You need to delete and recreate all services of type Load Balancer, which means that the public ip addresses of your service endpoints will change.<br/>
+Please do this only if the Stakeholder really needs to migrate this Shoot to use standard Load Balancers. All new Shoot clusters will automatically use Azure Standard Load Balancers.
+
+1. Disable temporarily Gardeners reconciliation.<br>
+The Gardener Controller Manager need to be configured to allow ignoring Shoot clusters.
+This can be configured in its the `ControllerManagerConfiguration` via the field `.controllers.shoot.respectSyncPeriodOverwrite="true"`.
+
+```sh
+# In the Garden cluster.
+kubectl annotate shoot <shoot-name> shoot.garden.sapcloud.io/ignore="true"
+
+# In the Seed cluster.
+kubectl -n <shoot-namespace> scale deployment gardener-resource-manager --replicas=0
+```
+
+2. Backup all Kubernetes services of type Load Balancer.
+```sh
+# In the Shoot cluster.
+# Determine all Load Balancer services.
+kubectl get service --all-namespaces | grep LoadBalancer
+
+# Backup each Load Balancer service.
+echo "---" >> service-backup.yaml && kubectl -n <namespace> get service <service-name> -o yaml >> service-backup.yaml
+```
+
+3. Delete all Load Balancer services.
+```sh
+# In the Shoot cluster.
+kubectl -n <namespace> delete service <service-name>
+```
+
+4. Wait until until Load Balancer is deleted.
+Wait until all services of type Load Balancer are deleted and the Azure Load Balancer resource is also deleted.
+Check via the Azure Portal if the Load Balancer within the Shoot Resource Group has been deleted.
+This should happen automatically after all Kubernetes Load Balancer service are gone within a few minutes.
+
+Alternatively the Azure cli can be used to check the Load Balancer in the Shoot Resource Group.
+The credentials to configure the cli are available on the Seed cluster in the Shoot namespace.
+```sh
+# In the Seed cluster.
+# Fetch the credentials from cloudprovider secret.
+kubectl -n <shoot-namespace> get secret cloudprovider -o yaml
+
+# Configure the Azure cli, with the base64 decoded values of the cloudprovider secret.
+az login --service-principal --username <clientID> --password <clientSecret> --tenant <tenantID>
+az account set -s <subscriptionID>
+
+# Fetch the constantly the Shoot Load Balancer in the Shoot Resource Group. Wait until the resource is gone.
+watch 'az network lb show -g shoot--<project-name>--<shoot-name> -n shoot--<project-name>--<shoot-name>'
+
+# Logout.
+az logout
+```
+
+5. Modify the `cloud-povider-config` configmap in the Seed namespace of the Shoot.<br/>
+The key `cloudprovider.conf` contains the Kubernetes cloud-provider configuration.
+The value is a multiline string. Please change the value of the field `loadBalancerSku` from `basic` to `standard`.
+Iff the field does not exists then append `loadBalancerSku: \"standard\"\n` to the value/string.
+```sh
+# In the Seed cluster.
+kubectl -n <shoot-namespace> edit cm cloud-provider-config
+```
+
+6. Enable Gardeners reconcilation and trigger a reconciliation.
+```
+# In the Garden cluster
+# Enable reconcilation
+kubectl annotate shoot <shoot-name> shoot.garden.sapcloud.io/ignore-
+
+# Trigger reconcilation
+kubectl annotate shoot <shoot-name> shoot.garden.sapcloud.io/operation="reconcile"
+```
+Wait until the cluster has been reconciled.
+
+6. Recreate the services from the backup file.<br/>
+Probably you need to remove some fields from the service defintions e.g. `.spec.clusterIP`, `.metadata.uid` or `.status` etc.
+```sh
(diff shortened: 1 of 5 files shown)


dkistner:
Please do not merge yet, due to some ongoing test activities. 
I raised the pr upfront to maintain the release notes :)

rfranzke:
> To migrate from the basic sku to standard it is required to remove all services with type `LoadBalancer` in the cluster. When all services are removed then Gardener needs to reconcile the cluster. After that the services can be recreated.

Can you explain again how this needs to be done? I as an end-user need to delete all my load balancer services, ok. What happens if I create new services of type load balancer (without deleting the old ones (because I don't want to?)).

>  When all services are removed then Gardener needs to reconcile the cluster.

What does this mean? When this change is rolled out then the cloud provider config is updated for all existing clusters automatically. I don't understand this part.

Does this configuration work for all Kubernetes versions (1.10.x - 1.15.x)?

rfranzke:
OK, great, thanks for the explanation! 

> This is not really needed if the user recreates also the Gardener managed lb services e.g. vpn-shoot.

I don't think we can implement anything because we don't know when the user wants to migrate (if he/she wants to at all). I think the solution you described is good: If the user wants to migrate, simply delete all load balancer services (also ours in the `kube-system` namespace).

> Let's discuss if we need to implement migration paths within the azure extension to move from basic to standard load balancers.

Same, I think your solution is good if this really works for all Kubernetes versions. Let's keep it like this.

dkistner:
**Ok Update:**
When the cloud-provider-config already contains the `standard` lb sku config, but the load balancers is not migrated (all services deleted and lb recreated), then it is not possible to add new lb services, because k8s is creating new pubic ips of type `standard`, but the existing `basic` load balancer is not supporting this type (only public ips of type `basic`).

Therefore I see possible options:
1. All new Shoots using automatically `standard` lbs. Existing Shoots remain with `basic` lbs. But I think we want a harmonised solution and we would later need to move to `standard` -> technical depts.
2. Implement migration paths to move from `basic` to `standard` lbs (means we need to delete all lb service and recreate them again).
3. Make it configurable in the Shoot config, which type of load balancer should be used. Default would be `basic`. This field needs to be immutable (as long as no migration logic is implemented). We would support both types of lbs.

rfranzke:
Any plan going forward? If not yet and further design discussions are needed, can we close the PR and continue in an issue?

dkistner:
Yes, let's close it for the moment. Will reopen it when we have a decission and continue to work on it.

rfranzke:
OK, thanks 👍

dkistner:
Reopened this PR.
We decided to move forward and create new Azure Shoots with standard LoadBalancers instead of basic ones.
We do not implement an automatic migration and old cluster needs to migrated manually if they need to use standard LoadBalancers.

controllers/provider-azure/docs/migrate-loadbalancer.md
```diff
@@ -0,0 +1,68 @@
+# Migrate Azure Shoot LoadBalancer from basic to standard SKU
+
+This guide descibes how to migrate the LoadBalancer of a Azure Shoot cluster from the basic SKU to the standard SKU.<br/>
+**Be aware:** You need to delete and recreate all services of type LoadBalancer, which means that the public ip addresses of your service endpoints will change.<br/>
+Please do this only if you really need to migrate this Shoot to use standard LoadBalancers. All new Shoot cluster will automatically use standard LoadBalances.
```
rfranzke:
```suggestion
Please do this only if you really need to migrate this Shoot to use standard LoadBalancers. All new Shoot clusters will automatically use standard LoadBalances.
```

controllers/provider-azure/docs/migrate-loadbalancer.md
```diff
@@ -0,0 +1,68 @@
...
+
+1. Disable temporarily Gardeners reconciliation.
+```sh
+# In the Garden cluster.
+kubectl annotate shoot <shoot-name> shoot.garden.sapcloud.io/ignore="true"
```
rfranzke: This only works if it is enabled in the GCM config.

controllers/provider-azure/pkg/controller/controlplane/valuesprovider.go
```diff
@@ -274,3 +286,30 @@ func getInfraNames(infraStatus *apisazure.InfrastructureStatus) (string, string,
 
 	return nodesSubnet.Name, nodesAvailabilitySet.Name, nodesRouteTable.Name, nodesSecurityGroup.Name, nil
 }
+
+func determineLoadBalancerType(ctx context.Context, c client.Client, namespace string) (string, error) {
+	var (
+		cm    = &corev1.ConfigMap{}
+		cmRef = kutil.Key(namespace, "cloud-provider-config")
```
rfranzke: Use the constant for the cloud provider config.

controllers/provider-azure/docs/migrate-loadbalancer.md
```diff
@@ -0,0 +1,68 @@
+# Migrate Azure Shoot LoadBalancer from basic to standard SKU
+
+This guide descibes how to migrate the LoadBalancer of a Azure Shoot cluster from the basic SKU to the standard SKU.<br/>
```
MSSedusch: an Azure Shoot

controllers/provider-azure/docs/migrate-loadbalancer.md
MSSedusch: Azure Standard Load Balancers

controllers/provider-azure/docs/migrate-loadbalancer.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+3. Delete all LoadBalancer services.
+```sh
+# In the Shoot cluster.
+kubectl -n <namespace> delete service <service-name>
+```
+
+4. Wait until until LoadBalancer is deleted.
+Wait until all services of type LoadBalancer are deleted and the LoadBalancer on Azure is also deleted.
```
MSSedusch: Wait until all services of type LoadBalancer are deleted and the Azure Load Balancer resource is also deleted.

controllers/provider-azure/docs/migrate-loadbalancer.md
```diff
@@ -0,0 +1,68 @@
...
+Check via the Azure Portal if the LoadBalancer within the Shoot Resource Group has been deleted.
```
MSSedusch: Can we assume that a cluster operator has access to the Azure portal?

dkistner: No probably not. But an operator will have access to the Seed cluster and therefore also access to the infrastructure credentials and could use the azure cli to do the same. I will add this to the guide. Thanks!

controllers/provider-azure/docs/migrate-loadbalancer.md
dkistner: Hmm ok. Wasn't aware of that. To do the migration successfully it is required to avoid any reconciliation during the process. I will add a note which explains that the GCM needs to be configured to allow ignoring Shoots. Thanks!

dkistner: done.

controllers/provider-azure/docs/migrate-loadbalancer.md
dkistner: done.

dkistner:
Seems that some tests for the Azure Provider extension does not work. I will fix this.

PR #248, opened 2019-08-06
[WIP] Jeos customization

pablochacin:
Gardener allows the specification of per-worker set customization of OSC. 

When reconciling the OSC, the actuator retrieves the provider config and passes it to a customizer, which adds additional units, files, and commands to the cloudconfig before passing it to the generator.

This is a work in progress. Some tasks still pending:
- [ ] Define the customization options for JeOS based on actual use cases from users
- [ ] Translate the customization options to cloudconfig elements
- [ ] Change the brute force-based decoding used for ProviderConfig to JeOS OSC customization, by an schema-based approach.

Fixes #

This PR has not yet been tested due to issues setting a development environment.

**Release note**:

```improvement operator
Customization options defined in the JeOS configuration for a worker pool will be applied to the cloudconfig for worker nodes. 
```

### DIFF
--- a/controllers/os-suse-jeos/pkg/customizer/customizer.go
+++ b/controllers/os-suse-jeos/pkg/customizer/customizer.go
@@ -1,0 +1,88 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package customizer
+
+import (
+	customizer "github.com/gardener/gardener-extensions/pkg/controller/operatingsystemconfig/oscommon/customizer"
+	generator "github.com/gardener/gardener-extensions/pkg/controller/operatingsystemconfig/oscommon/generator"
+	yaml "gopkg.in/yaml.v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+)
+
+type JeosCustomizer struct {
+}
+
+func NewCustomizer() (customizer.Customizer, error) {
+	return &JeosCustomizer{}, nil
+}
+
+func (c *JeosCustomizer) Customize(osconf *generator.OperatingSystemConfig, extension *runtime.RawExtension) (*generator.OperatingSystemConfig, error) {
+
+	customization := JeosOperatingSystemCustomization{}
+	err := yaml.Unmarshal(extension.Raw, &customization)
+	if err != nil {
+		return nil, err
+	}
+
+	var units []*generator.Unit
+	var files []*generator.File
+	var commands []string
+
+	var customized = &generator.OperatingSystemConfig{}
+
+	for _, service := range customization.EnableServices {
+		cmd := "systemctl enable " + service
+		commands = append(commands, cmd)
+	}
+
+	for _, cmd := range customization.BootCommands {
+		commands = append(commands, cmd)
+	}
+
+	customized.Files = append(osconf.Files, files...)
+	customized.Units = append(osconf.Units, units...)
+	customized.Commands = append(osconf.Commands, commands...)
+	customized.Bootstrap = osconf.Bootstrap
+	customized.Path = osconf.Path
+
+	return customized, nil
+}
+
+type JeosOperatingSystemCustomization struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// systmctl services to enable
+	EnableServices []string
+
+	// commands to execute at boot time
+	BootCommands []string
+}
--- a/controllers/os-suse-jeos/cmd/gardener-extension-os-suse-jeos/app/app.go
+++ b/controllers/os-suse-jeos/cmd/gardener-extension-os-suse-jeos/app/app.go
@@ -16,6 +16,7 @@ package app
 
 import (
 	"context"
+	"github.com/gardener/gardener-extensions/controllers/os-suse-jeos/pkg/customizer"
 	"github.com/gardener/gardener-extensions/controllers/os-suse-jeos/pkg/generator"
 	"github.com/gardener/gardener-extensions/pkg/controller/cmd"
 	"github.com/gardener/gardener-extensions/pkg/controller/operatingsystemconfig/oscommon/app"
@@ -24,10 +25,15 @@ import (
 
 // NewControllerCommand returns a new Command with a new Generator
 func NewControllerCommand(ctx context.Context) *cobra.Command {
+	c, err := customizer.NewCustomizer()
+	if err != nil {
+		cmd.LogErrAndExit(err, "Could not create JeOS CloudInit customizer")
+	}
+
 	g, err := generator.NewCloudInitGenerator()
 	if err != nil {
 		cmd.LogErrAndExit(err, "Could not create Generator")
 	}
 
-	return app.NewControllerCommand(ctx, "suse-jeos", g)
+	return app.NewControllerCommand(ctx, "suse-jeos", g, c)
 }
--- a/controllers/os-suse-jeos/pkg/generator/templates/cloud-init-suse-jeos.template
+++ b/controllers/os-suse-jeos/pkg/generator/templates/cloud-init-suse-jeos.template
@@ -34,3 +34,6 @@ runcmd:
 {{ range $_, $unit := .Units -}}
 - systemctl enable '{{ $unit.Name }}' && systemctl restart '{{ $unit.Name }}'
 {{ end -}}
+{{ range $_, $command := .Commands -}}
+- '{{ command }}'
+{{ end -}}
--- a/pkg/controller/operatingsystemconfig/oscommon/actuator/actuator.go
+++ b/pkg/controller/operatingsystemconfig/oscommon/actuator/actuator.go
@@ -16,6 +16,7 @@ package actuator
 
 import (
 	"github.com/gardener/gardener-extensions/pkg/controller/operatingsystemconfig"
+	"github.com/gardener/gardener-extensions/pkg/controller/operatingsystemconfig/oscommon/customizer"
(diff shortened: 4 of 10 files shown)


controllers/os-suse-jeos/pkg/customizer/customizer.go
```diff
@@ -0,0 +1,88 @@
...
+	customized.Files = append(osconf.Files, files...)
+	customized.Units = append(osconf.Units, units...)
+	customized.Commands = append(osconf.Commands, commands...)
+	customized.Bootstrap = osconf.Bootstrap
+	customized.Path = osconf.Path
+
+	return customized, nil
+}
+
+type JeosOperatingSystemCustomization struct {
```
rfranzke: Take a lot at the https://github.com/gardener/gardener-extensions/tree/master/controllers/provider-aws/pkg/apis/aws package. It creates a new API group and a resource inside. I guess that's what you meant with 'scheme-based'. Please follow the same approach.

controllers/os-suse-jeos/pkg/customizer/customizer.go
```diff
@@ -0,0 +1,88 @@
...
+	metav1.TypeMeta `json:",inline"`
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// systmctl services to enable
+	EnableServices []string
```
rfranzke: Did you talk to the users? Do they need to control which services to enable? Or what's the reason for introducing this field?

rfranzke:
@adracus, please take a look as well here.

controllers/os-suse-jeos/pkg/customizer/customizer.go
pablochacin: This is actually an example, as I mention in the PR description, collecting user requirements is an ongoing task.

pablochacin:
@rfranzke: when discussing #223 you said
> but most users don't (want to) know this stuff, I assume. They simply want to set a flag "prepareMyClusterForHANA=true` 

I think a way to achieve this and still make the customization flexible to accommodate multiple applications and platforms without having to implement logic in the controller for each case would be tu use the concept of "OSC Profile", which is basically a CRD with the cunstomization required for a particular use case, and then specify this profile in the OSC

```
apiVersion: core.gardener.cloud/v1alpha1
kind: Shoot
metadata:
spec:
  workers:
  - name: hana-worker-pool
    machineImage:
      name: suse-jeos
      version: 1234
      providerConfig:
        apiVersion: suse-jeos.operatingsystemconfig.extensions.gardener.cloud/v1alpha1
        kind: Customization
        profile: hana
 ```
The profile can then have all the customization required using some level of abstraction above plain cloudinit:

```  
apiVersion: suse-jeos.operatingsystemconfig.extensions.gardener.cloud/v1alpha1
kind: Profile
metadata:
   name: hana
spec:
  bootcommads:
    - command
    - command
  osSettings:
     settingX: value
     settingY: value 
 ```

This is very much a mix what you suggested and the idea of having a CRD I proposed initially.

rfranzke:
I like the profile idea, that sounds good and helps with both requirements. What I understand from this proposal is that these operating system controller specific CRDs would be created in the Garden cluster so that end-users can see what is available? Actually, I don't see the need for the profile being a CRD, maybe a normal config map is sufficient? Also, it's probably hard to design the 'spec' of such a profile to make it general enough for everybody, but if you have ideas to achieve that then it's probably worth to follow this path.

pablochacin:
> What I understand from this proposal is that these operating system controller specific CRDs  would be created in the Garden cluster so that end-users can see what is available?

I think no. Based on our previous discussion, the end user shouldn't have to be aware or have the knowledge to understand the internals of the configuration. I would expect the application or cluster administrator to be the one defining it and providing the information to users. 

>Actually, I don't see the need for the profile being a CRD, maybe a normal config map is sufficient?

I think that having a CRD would allow for validation when the CRD is created and also allow the controller to watch for changes in CRD.

rfranzke:
> I think no. Based on our previous discussion, the end user shouldn't have to be aware or have the knowledge to understand the internals of the configuration. I would expect the application or cluster administrator to be the one defining it and providing the information to users.

OK.

> I think that having a CRD would allow for validation when the CRD is created and also allow the controller to watch for changes in CRD.

Using a CRD just because we want to validate it doesn't sound reasonable, IMHO. Watching for changes is also not a strong argument as you also watch the file system if the configuration changes (if this is really needed, but I guess it's not).

We had a similar use-case for our [certificate-service extension](https://github.com/gardener/gardener-extensions/tree/master/controllers/extension-certificate-service) and we also discussed about using a CRD because then we could (more easily) validate the content.
However, we then discarded this idea and decided to go with a [versioned configuration](https://github.com/gardener/gardener-extensions/blob/master/controllers/extension-certificate-service/example/00-config.yaml) for the following reasons:
* CRD validation only allows static validation, but if you have cross-field constraints then you cannot use this (e.g., field `a` may only have a certain value if field `b` is `true`).
* Additional CRD in the system that is no "normal" Kubernetes resource somebody is acting on (no reconcile/delete loop), it's just used as a config map, so why not using `ConfigMap`?
* No real benefit compared to a normal configuration file, but additional complexity inside the system.

Such a [versioned configuration](https://github.com/gardener/gardener-extensions/blob/master/controllers/extension-certificate-service/example/00-config.yaml) file can be validated as well during start-up of the controller, and if it's invalid, then it simply fails to start. That's how Kubernetes is doing it with its component configs.

Going forward, I like the profile approach you raised. What do you think about introducing a versioned config map (that's btw. what most provider extensions (and the mentioned certificate service) are doing today already, so the approach is working quite well) for the SUSE-JeOS controller:

```yaml
apiVersion: suse-jeos.operatingsystemconfig.extensions.gardener.cloud/v1alpha1
kind: Configuration
profiles:
- name: hana
  bootCommands:
  - command
  - command
  osSettings:
    settingX: value
    settingY: value 
```

And then, as you suggested, the `Shoot` may simply select a supported profile:

```yaml
apiVersion: core.gardener.cloud/v1alpha1
kind: Shoot
metadata:
spec:
  workers:
  - name: hana-worker-pool
    machineImage:
      name: suse-jeos
      version: 1234
      providerConfig:
        apiVersion: suse-jeos.operatingsystemconfig.extensions.gardener.cloud/v1alpha1
        kind: Customization
        profile: hana
```

pablochacin:
>Using a CRD just because we want to validate it doesn't sound reasonable, IMHO. 

Not saying this is the ONLY reason, but one reason in favor or this.

>Watching for changes is also not a strong argument as you also watch the file system if the configuration changes (if this is really needed, but I guess it's not).

It will actually needed, at some point. Plese notice this is not intended to support ONLY the specific HANA case (that would the case, we could just hard code the configuration in the operator)

> Additional CRD in the system that is no "normal" Kubernetes resource somebody is acting on (no reconcile/delete loop), it's just used as a config map, so why not using ConfigMap?

Maybe there's a misunderstanding regarding the purpose of the proposed CRD: it is not an configuration for the operator but for one application. We expect such profiles to be created or modified during the life of an operator without having to restart the operator. 
 
> Such a versioned configuration file can be validated as well during start-up of the controller, and if it's invalid, then it simply fails to start. That's how Kubernetes is doing it with its component configs.

I find unreasonable that a configuration error in one application profile prevents the operator to start.

rfranzke:
> Not saying this is the ONLY reason, but one reason in favor or this.

What are the others?

> It will actually needed, at some point. Plese notice this is not intended to support ONLY the specific HANA case (that would the case, we could just hard code the configuration in the operator)

I don't understand. Instead of using a CRD we would have the profile configuration in the controller's configmap, nothing is hard-coded anywhere. Operators can still change/modify/extend/shrink this information dynamically to support additional use-cases.

> Maybe there's a misunderstanding regarding the purpose of the proposed CRD: it is not an configuration for the operator but for one application.

That's clear, end-users can select a profile in their `Shoot` manifest as you suggested in a previous comment. Also, you explained that the profiles are maintained by operators/administrators, and no end-user has access to them. 

> We expect such profiles to be created or modified during the life of an operator without having to restart the operator.

Yes, a config map would require restarting the controller (if we don't want to implement a file system watch (which I wouldn't do in the beginning because it's not needed - most probably these profiles aren't touched/changed often and it doesn't harm if the controllers are not available for 10 seconds due to a restart)).

> I find unreasonable that a configuration error in one application profile prevents the operator to start.

I don't understand, why not? That's a well-known pattern that is implemented by Kubernetes (and tons of other applications) as well. If the configuration is invalid then don't start and tell what is wrong such that the operator can fix it. As mentioned above, CRD validation is very limited, and if you want to do it correctly you anyway need to have a validating webhook. All this complexity isn't needed if we use the CRD *just* for validation. If you have other use-cases then please provide arguments for them. I'm always open to adjust such design decisions if there are reasonable points for doing so.

pablochacin:
> What are the others?

Basically been able to add/update profiles without having to re-deploy the controller. I think the use case we have in mind is not the same: This PR intends to support application specific OS customizations. Please notice the plural form. Not only one customization but multiple customizations.  

To be honest, I don't see why using a CRD is such a big concern considering that there's already a controller which can handle it.

rfranzke:
> Basically been able to add/update profiles without having to re-deploy the controller.

We already talked about this one: a) I don't think we need this because it doesn't matter if the controller is offline for a few minutes, and b) if you still want to have it then use a file system watch mechanism.

> I think the use case we have in mind is not the same: This PR intends to support application specific OS customizations. Please notice the plural form. Not only one customization but multiple customizations.

Sure, that's clear. We can easily put multiple profiles into the configuration.

Your two points are "validation" and "no redeployment of the controller". I'm saying that you should not abuse the CRD mechanism just because you want to validate the content. You need real code anyways in order to do the validation correctly (what if validity of a value in field `a` depends on a value of field `b`? You cannot solve this with CRD validation and need a webhook). The second argument is diminished above.
Using a CRD doesn't give anybody any benefit because no end-user can and shall read them. No reconciliation happens on these objects, it just increases the complexity in the system. These profiles shall be just used as config maps, so let's just use real `ConfigMap`s. That's what they've been made for.

I suggest focusing on the actual content of the customizations/profiles instead of this meta discussion, please.

pablochacin:
> I suggest focusing on the actual content of the customizations/profiles instead of this meta discussion, please.

I agree this metadiscussion is not most productive way to use time, but I'm surprised to find that contributions MUST adhere to specific design options and alternatives sems not to considered possible, regardless of the criteria of the contributor. Something I wouldn't have expected. And definitely something I don't feel confortable with.

rfranzke:
> but I'm surprised to find that contributions MUST adhere to specific design options and alternatives sems not to considered possible, regardless of the criteria of the contributor. Something I wouldn't have expected. And definitely something I don't feel confortable with.

I’m focusing purely on the technical aspects and I want to achieve a solution that is both easily understandable and maintainable. If alternatives would not be possible then we could simply write down how we want it to get implemented and execute it. But that’s not how it’s done, quite the contrary, I’m spending time to exchange arguments with you and try to converge to a good design. Sure, there are more important discussions than this one, and most probably it doesn’t matter too much whether we use a CRD or a configmap. Still, imho it’s a pure abuse if the one and only motivation is validation. CRDs are meant for things that must be reconciled, i.e. desired state that somehow needs to be achieved. However, no controller would reconcile these profiles, because they don’t represent desired state. They are just configuration for `OperatingSystemConfig`s.

As said, if only static validation features are used it would probably not too much of a problem using a CRD, but as soon as it gets more complex (which often happens after some time) and a validating webhook would be needed then I definitely have a problem with this approach. So, why not directly doing it in a better way if such shortcomings are already discovered?
In the previous comment I summarized your and my points. My problem now is that I always try to find counter-arguments while you are ignoring mine. It is no problem to convince me if realistic, valid, and good arguments are put forward. This didn’t happen so far. Why don’t you consider having a validating webhook a problem/ugliness, why don’t you think a configmap is a cleaner approach (especially after reading https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#should-i-use-a-configmap-or-a-custom-resource), why do you consider it a problem that the controller would need a restart after a change? Are there things that you want to implement in the near future  (that I’m not knowing of) that would support the CRD approach?

Anyways, let’s try to come to a conclusion. The standard and easist way to achieve this is a simple vote. @timuthy @adracus @zanetworker @mvladev @ialidzhikov @vpnachev @vlvasilev what are your opinions here?

pablochacin:
> I’m spending time to exchange arguments with you and try to converge to a good design. 

Both of us have, and I appreciate your time.

Please don't confuse my reluctance to accept some of your proposal as a lack of understanding of design principles.  It may happen that simply that what we understand by "good design" and "easily understandable and maintainable" is not the same. Neither are the same the use cases we have in mind. Hence it is difficult to arrive to a common understanding.

PR #258, opened 2019-08-12
Webhooks don't mutate objects if deletion timestamp is set

rfranzke:
Webhooks no longer mutate objects if deletion timestamp is set

**Release note**:

```improvement operator
All webhooks don't mutate any object anymore if the respective deletion timestamp is already set.
```


timuthy (approved):
/lgtm

PR #262, opened 2019-08-13
Enable SSE of buckets on Alicloud

minchaow:
This PR is to enable Server Side Encryption (SSE) of OSS buckets on Alicloud. 
Fixes #
None
Buckets already created will not be affected, which means only buckets created after the change will enable SSE.

### DIFF
--- a/controllers/provider-alicloud/pkg/alicloud/client/client.go
+++ b/controllers/provider-alicloud/pkg/alicloud/client/client.go
@@ -117,6 +117,12 @@ func (c *storageClient) CreateBucketIfNotExists(ctx context.Context, bucketName 
 		}
 		return err
 	}
+
+	config := oss.ServerEncryptionRule{SSEDefault: oss.SSEDefaultRule{SSEAlgorithm: "KMS"}}
+	if err := c.client.SetBucketEncryption(bucketName, config); err != nil {
+		return err
+	}
+
 	return nil
 }
 


controllers/provider-alicloud/pkg/alicloud/client/client.go
```diff
@@ -117,6 +117,12 @@ func (c *storageClient) CreateBucketIfNotExists(ctx context.Context, bucketName
 		}
 		return err
 	}
+
+	config := oss.ServerEncryptionRule{SSEDefault: oss.SSEDefaultRule{SSEAlgorithm: "KMS"}}
```
swapnilgm: If i understood the API correctly, the default CMK generated by OSS will be used for encryption.  What will happen if this key is lost somehow? Will we still be able to decrypt the backups. ALso, one more thing i'm not clear from (doc)[https://www.alibabacloud.com/help/doc-detail/31871.htm?spm=a2c63.p38356.a1.2.380324d3RO5ivi] specially following line
> To encrypt an object when creating it, you can include the x-oss-server-side-encryption header in the request and specify its value to KMS 

 Does setting bucket encryption also encrypts the objects by default. Or for encrypting the object one  has to update the object request compulsory to have above mentioned header.

minchaow: @swapnilgm Thank for your good questions. For your 1st question, if we use KMS to encrypt OSS objects, one default CMK will be generated, and it cannot be deleted when being used, so it should not be lost. For your 2nd question, when the encryption of buckets is enabled, objects uploaded to these buckets will be encrypted by default. I will have a double check tomorrow.

swapnilgm: >one default CMK will be generated, and it cannot be deleted when being used, so it should not be lost. 

Cool. One more question then, as you mentioned in release notes 
> Buckets already created will not be affected, which means only buckets created after the change will enable SSE.

 On the similar line, will we be able to change the default generated CMK later on to custom CMK? What will happen to old encrypted objects in that case?

> when the encryption of buckets is enabled, objects uploaded to these buckets will be encrypted by default. I will have a double check tomorrow.

Sure. Please confirm and reply. I'll rely on you for this clarification

rfranzke:
Should we revert https://github.com/gardener/gardener/pull/1321 ?

swapnilgm:
> Should we revert [gardener/gardener#1321](https://github.com/gardener/gardener/pull/1321) ?

Yes. Anyways we will be switching soon to extension with new shared bucket with same release probably. So, i would prefer  to avoid unnecessary changes with existing backup bucket for existing shoot.

rfranzke:
Done with https://github.com/gardener/gardener/commit/dfa769d4f38d40d7b1a92707e22c286d6d5ed74f

controllers/provider-alicloud/pkg/alicloud/client/client.go
minchaow: First to correct one my yesterday answer after checked with Alicloud support engineers. Object uploaded will not be encrypted by default, even the bucket has been already enabled SSE. `x-oss-server-side-encryption` header is required. 

> On the similar line, will we be able to change the default generated CMK later on to custom CMK? What will happen to old encrypted objects in that case?

Code on the line is only to enable buckets encrypted using KMS, not directly related to CMK. According to my understanding, custom CMKs usage is on the **objects** level, which means 
we can use different CMKs to encrypt different objects in the same bucket. What need to do is set both both `x-oss-server-side-encryption` and `X-oss-server-side-encrpytion-key-id` in the header.  For exapmle, https://github.com/aliyun/aliyun-oss-go-sdk/blob/master/sample/put_object.go#L72

swapnilgm: May i know the immediate use case for having server side bucket encryption? Is there any specific requirement on alicloud support?

If not, i'll prefer closing this PR. And take this a task asa common task for all providers.

minchaow: The change is not hurry, so it is fine to close it.

PR #266, opened 2019-08-15
Fix AWS ControlPlane controller shoot delete flow

### LINKED ISSUE #233: Control plane exposure controller

svetlinas:
Currently AWS provider controller is failing during shoot deletion because of not fully descriptive  Unstructured object in one of the chart definitions. This Unstructured object needs to represent  VerticalPodAutoscaler. 
The PR is also adding Unstructured objects to all of the charts which contain VerticalPodAutoscalers in order for these objects to be specifically deleted during shoot deletion.

Fixes #233 

**Release note**:

```improvement operator
Fix AWS ControlPlane controller shoot delete flow
```


controllers/provider-aws/pkg/controller/controlplane/valuesprovider.go
```diff
@@ -292,3 +293,10 @@ func getCCMChartValues(
 
 	return values, nil
 }
+
+// getVerticalPodAutoscalerObject returns Unstructured object describing VerticalPodAutoscaler of group "autoscaling.k8s.io"
+func getVerticalPodAutoscalerObject() *unstructured.Unstructured {
+	obj := &unstructured.Unstructured{}
+	obj.SetGroupVersionKind(schema.GroupVersionKind{Group: "autoscaling.k8s.io", Kind: "VerticalPodAutoscaler", Version: "v1beta2"})
```
timuthy: Can you try to reuse `SchemeGroupVersion` from https://github.com/kubernetes/autoscaler/blob/5ff9f7b8ba65cfec027da647aff79f63c1176fbe/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2/register.go#L26?

svetlinas: I tried, but I cannot make gardener-extensions to depend on this project, I cannot build the correct project:version

PR #267, opened 2019-08-15
Add logrotate and jq package. Disable cloud-init network configuration.

vlvasilev:
This PR adds logrotate and jq package for the proper work of the logrotate service and monitoring the kubelet on the Ubuntu OS.
Also the network configuration by cloud-init is disabled. This is needed because when we don't explicitly set the network configuration the network-init module try to search for network version in None object which leads to panic. 
Also the test for the ubuntu cloud-init generator is moved to the same directory where the generator is. 

**Release note**:

```noteworthy operator
The `os-ubuntu` controller now explicitly adds the `logrotate` and `jq` packages. Also, the `cloud-init` module is now disabled.
```


rfranzke (approved):
/lgtm

vpnachev (approved):
/lgtm

PR #276, opened 2019-08-17
Fix networking-calico RBAC

zanetworker:
Add missing cluster RBAC role for the networking-calico controller. 

As part of the refactoring, code support for configuration of IPAM was added. 

**Release note**:

```improvement operator
None
```


controllers/networking-calico/charts/internal/calico/values.yaml
```diff
@@ -2,6 +2,8 @@ global:
   podCIDR: ""
 config:
   backend: bird
+  ipam:
+    type: "host-local"
```
rfranzke:
```suggestion
    type: "host-local"
  # subnet: "usePodCidr"
```

controllers/networking-calico/pkg/charts/utils.go
```diff
@@ -21,20 +21,67 @@ import (
...
+				configValues = map[string]interface{}{
+					"backend": string(config.Backend),
+					"ipam": map[string]interface{}{
+						"type": config.IPAM.Type,
+					},
+				}
+			}
+		}
+	} else {
+		configValues = map[string]interface{}{
```
rfranzke: Add this to the `calicoChartValues` above. This way you can get rid of the `else` branch (which makes reading easier).

controllers/networking-calico/pkg/charts/utils.go
```diff
@@ -21,20 +21,67 @@ import (
...
+				calico.NodeImageName:            imagevector.CalicoNodeImage(),
+			},
+			"global": map[string]string{
+				"podCIDR": network.Spec.PodCIDR,
+			},
+		}
+	)
+
+	var configValues = map[string]interface{}{}
+	if config.IPAM != nil {
```
rfranzke:
```suggestion
	if config.IPAM != nil && len(config.IPAM.Type) > 0 {
```

Get rid of one level of nesting

PR #282, opened 2019-08-21
Define `ProviderProfile`s for all providers

rfranzke:
With this PR the `ProviderProfile` objects are defined that will be embedded into the new `core.gardener.cloud/v1alpha1.ProviderProfile` resource as proposed in [GEP-4](https://github.com/gardener/gardener/pull/1284/files#diff-bf2774d9954baab517306db45a5b80bbR131-R218)

* The `ProviderProfile`s are not yet used because the first need to modify Gardener to use the `ProviderProfile` instead of the `CloudProfile` in its shoot reconciliation.
* We will move the machine image mapping from the componentconfig to the `ProviderProfile` soon, and we want to harmonize them to support multiple versions per machine image (like it is supported in the `CloudProfile` at the moment).
* Validation of the `ProviderProfile`s has been implemented in the `pkg/apis/<provider>/validation` packages.

**Release note**:

```improvement operator
NONE
```


rfranzke:
ping @timuthy @ialidzhikov @danielfoehrKn - any feedback? can we merge this?

controllers/provider-openstack/pkg/apis/openstack/validation/validation_providerprofile.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+	"time"
+
+	apisopenstack "github.com/gardener/gardener-extensions/controllers/provider-openstack/pkg/apis/openstack"
+
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// ValidateProviderProfileConfig validates a ProviderProfileConfig object.
+func ValidateProviderProfileConfig(providerProfile *apisopenstack.ProviderProfileConfig) field.ErrorList {
```
timuthy: Should we also have a validation for `DNSServers`, e.g. `net.ParseIP(dnsAddr) != nil` ?

rfranzke: Done, PTAL again.

PR #286, opened 2019-08-22
Add extension controllers logging configuration

svetlinas:
Add logging configuration of provider-specific components, e.g. for the cloud-controller-manager, to Extension controllers , because only they know these components and how they are displayed in the best way..
Every extension controller deploys a ConfigMap into the `garden` namespace in the seed cluster. This ConfigMap contains the respective information for Gardener's logging stacks. 

Part of https://github.com/gardener/gardener/issues/1351 

**Release note**:

```noteworthy operator
Add provider extension controllers are now deploying their logging configuration via a `ConfigMap`.
```


controllers/provider-aws/charts/provider-aws/templates/configmap-logging.yaml
```diff
@@ -0,0 +1,42 @@
...
+        Key_Name            log
+        Parser              lbReadvertiserParser
+        Reserve_Data        True
+
+    [FILTER]
+        Name record_modifier
+        Match *cloud-controller-manager*
+        Record type user
+
+  parsers.conf: |-
```
rfranzke: It should only be deployed if the seed provider is `aws`: https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-aws/charts/provider-aws/templates/daemonset-custom-mtu.yaml#L1

controllers/provider-aws/charts/provider-aws/templates/configmap-logging.yaml
```diff
@@ -0,0 +1,42 @@
...
+        Reserve_Data        True
+
+    [FILTER]
+        Name                parser
+        Match               kubernetes.machine-controller-manager*machine-controller-manager*
+        Key_Name            log
+        Parser              kubeapiserverParser
+        Reserve_Data        True
+
+    [FILTER]
```
rfranzke: It should only be deployed if the seed provider is `aws`: https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-aws/charts/provider-aws/templates/daemonset-custom-mtu.yaml#L1

controllers/provider-alicloud/charts/provider-alicloud/templates/configmap-logging.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+    name: {{ include "name" . }}-logging-config
```
rfranzke: Check the indentation (on all other configmaps as well), you're using three instead of two spaces

ialidzhikov:
> @ialidzhikov would it harm to have the same parsers multiple times in the fluent-bit configuration? As of today all extensions are installed into all seeds, so all of them would create the configmap.

fluent-bit can have multiple `[PARSER]` sections (but no duplication names allowed). From the PR I see that we only refer existing parsers (except 1 `[PARSER]`).
In the seed controller we know the seed provider type, we can pick the provider specific configmap and pass the values inside to the logging chart, right?

rfranzke:
> In the seed controller we know the seed provider type, we can pick the provider specific configmap and pass the values inside to the logging chart, right?

I don't think so, there might be multiple shoots with different providers on the same seed, and then we would have to take them all.

ialidzhikov:
> I don't think so, there might be multiple shoots with different providers on the same seed, and then we would have to take them all.

Ok then lets add multiple `[FILTER]` for the cloud-controller-manager and machine-controller-manager of each provider. I suggest to modify the ccm and mcm container names such that the provider is also part of the container name. The container name then can be used to match in the `Match` section of the `[FILTER]`. In this way each provider `[FILTER]` will match only his own ccm or mcm.
The match section of the filters will look like:
```
   Match               kubernetes.cloud-controller-manager*cloud-controller-manager-aws*
```

WDYT?

rfranzke:
Yes, sounds like a reasonable option (although, I would add the provider as prefix instead of suffix because that's [how the CCM executables are called as well](https://github.com/kubernetes/cloud-provider-aws/tree/master/cmd/aws-cloud-controller-manager), but that's just a trivial detail). @svetlinas can you follow @ialidzhikov's suggestion, please?

svetlinas:
Yes, on it.

controllers/provider-aws/charts/provider-aws/templates/configmap-logging.yaml
```diff
@@ -0,0 +1,45 @@
...
+        Name record_modifier
+        Match *aws-cloud-controller-manager*
+        Record type user
+
+{{/*TODO: should this be put in each extension provider?    */}}
```
rfranzke: Nope, AWS is the only control plane exposure controller that implements anything. All other providers deploy only components into the seed that are configured for the shoot provider (not the seed provider).

Issue #154, opened 2019-06-20
Add support for TenantID and DomainID in OpenStack infra secret


afritzler:
**What would you like to be added**:
Currently when using OpenStack we only allow the usage of `tenantName` and `domainName` in the infrastructure secret [0]. We should also support IDs instead of name here. 

The integration has to happen in the following components
- [ ] Machine Controller Manager as it creates machines using gophercloud
- [ ] Seed configuration as set up by Gardener since the credentials are used for the backup infra
- [ ] Control Plane as setup by Gardener. This will be done in the CP extension controller in the future.
- [ ] Infrastructure Extension as needed for the Terraform creation of networks, routers etc.

**Why is this needed**:
There might be OpenStack setups where users want to use IDs instead of names all along. This scenarios should also be supported. 

[0] https://github.com/gardener/gardener/blob/master/pkg/operation/cloudbotanist/openstackbotanist/types.go#L29

PR #287, opened 2019-08-22
Added additional OpenStack Authentication Values 

afritzler:
Added additional OpenStack Authentication Values 

This PR is part of https://github.com/gardener/gardener-extensions/issues/154

**Release note**:
```noteworthyuser
Added support for additional OpenStack authentication values: `TenantID`, `DomainID`, `UserDomainName`, `UserDomainID`. Check out the example manifests.
```

### DIFF
--- a/controllers/provider-openstack/charts/internal/openstack-infra/templates/main.tf
+++ b/controllers/provider-openstack/charts/internal/openstack-infra/templates/main.tf
@@ -1,7 +1,23 @@
 provider "openstack" {
   auth_url    = "{{ required "openstack.authURL is required" .Values.openstack.authURL }}"
-  domain_name = "{{ required "openstack.domainName is required" .Values.openstack.domainName }}"
-  tenant_name = "{{ required "openstack.tenantName is required" .Values.openstack.tenantName }}"
+  {{- if .Values.openstack.domainName }}
+  domain_name = "{{ .Values.openstack.domainName }}"
+  {{- end}}
+  {{- if .Values.openstack.domainID }}
+  domain_id   = "{{ .Values.openstack.domainID }}"
+  {{- end}}
+  {{- if .Values.openstack.tenantName }}
+  tenant_name = "{{ .Values.openstack.tenantName }}"
+  {{- end}}
+  {{- if .Values.openstack.tenantID }}
+  tenant_id = "{{ .Values.openstack.tenantID }}"
+  {{- end}}
+  {{- if .Values.openstack.userDomainName }}
+  user_domain_name = "{{ .Values.openstack.userDomainName }}"
+  {{- end}}
+  {{- if .Values.openstack.userDomainID }}
+  user_domain_id = "{{ .Values.openstack.userDomainID }}"
+  {{- end}}
   region      = "{{ required "openstack.region is required" .Values.openstack.region }}"
   user_name   = "${var.USER_NAME}"
   password    = "${var.PASSWORD}"
--- a/controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
+++ b/controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
@@ -42,6 +42,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apiserver/pkg/authentication/user"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	gutils "github.com/gardener/gardener/pkg/utils"
 )
 
 // Object names
@@ -229,15 +231,29 @@ func getConfigChartValues(
 	values := map[string]interface{}{
 		"kubernetesVersion": cluster.Shoot.Spec.Kubernetes.Version,
 		"domainName":        c.DomainName,
+		"domainID":          c.DomainID,
 		"tenantName":        c.TenantName,
+		"tenantID":          c.TenantID,
 		"username":          c.Username,
 		"password":          c.Password,
+		"userDomainID":      c.UserDomainID,
+		"userDomainName":    c.UserDomainName,
 		"lbProvider":        cpConfig.LoadBalancerProvider,
 		"floatingNetworkID": infraStatus.Networks.FloatingPool.ID,
 		"subnetID":          subnet.ID,
 		"authUrl":           cluster.CloudProfile.Spec.OpenStack.KeyStoneURL,
 		"dhcpDomain":        cluster.CloudProfile.Spec.OpenStack.DHCPDomain,
 		"requestTimeout":    cluster.CloudProfile.Spec.OpenStack.RequestTimeout,
+	}
+
+	// Only the 1.15 out of tree provider for OpenStack support the OS_DOMAIN_NAME/ID
+	// https://github.com/kubernetes/cloud-provider-openstack/pull/733
+	ok, err := gutils.CompareVersions(cluster.Shoot.Spec.Kubernetes.Version, ">=", "1.15")
+	if ok {
+		values = map[string]interface{}{
+			"userDomainName": c.UserDomainName,
+			"userDomainID":   c.UserDomainID,
+		}
 	}
 
 	if cpConfig.LoadBalancerClasses == nil {
--- a/controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
+++ b/controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
@@ -246,7 +246,7 @@ func getConfigChartValues(
 
 	// Only the 1.15 out of tree provider for OpenStack support the OS_DOMAIN_NAME/ID
 	// https://github.com/kubernetes/cloud-provider-openstack/pull/733
-	ok, err := gutils.CompareVersions(cluster.Shoot.Spec.Kubernetes.Version, ">=", "1.15")
+	ok, _ := gutils.CompareVersions(cluster.Shoot.Spec.Kubernetes.Version, ">=", "1.15")
 	if ok {
 		values["userDomainName"] = c.UserDomainName
 		values["userDomainID"] = c.UserDomainID
--- a/controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
+++ b/controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
@@ -236,8 +236,6 @@ func getConfigChartValues(
 		"tenantID":          c.TenantID,
 		"username":          c.Username,
 		"password":          c.Password,
-		"userDomainID":      c.UserDomainID,
-		"userDomainName":    c.UserDomainName,
 		"lbProvider":        cpConfig.LoadBalancerProvider,
 		"floatingNetworkID": infraStatus.Networks.FloatingPool.ID,
 		"subnetID":          subnet.ID,
@@ -250,10 +248,8 @@ func getConfigChartValues(
 	// https://github.com/kubernetes/cloud-provider-openstack/pull/733
 	ok, err := gutils.CompareVersions(cluster.Shoot.Spec.Kubernetes.Version, ">=", "1.15")
 	if ok {
-		values = map[string]interface{}{
-			"userDomainName": c.UserDomainName,
-			"userDomainID":   c.UserDomainID,
-		}
+		values["userDomainName"] = c.UserDomainName
+		values["userDomainID"] = c.UserDomainID
 	}
 
 	if cpConfig.LoadBalancerClasses == nil {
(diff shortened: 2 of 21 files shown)


controllers/provider-openstack/charts/internal/openstack-infra/templates/main.tf
```diff
@@ -1,7 +1,23 @@
...
+  domain_name = "{{ .Values.openstack.domainName }}"
+  {{- end}}
+  {{- if .Values.openstack.domainID }}
+  domain_id   = "{{ .Values.openstack.domainID }}"
+  {{- end}}
+  {{- if .Values.openstack.tenantName }}
+  tenant_name = "{{ .Values.openstack.tenantName }}"
+  {{- end}}
+  {{- if .Values.openstack.tenantID }}
+  tenant_name = "{{ .Values.openstack.tenantID }}"
```
rfranzke:
```suggestion
  tenant_id = "{{ .Values.openstack.tenantID }}"
```

controllers/provider-openstack/charts/internal/openstack-infra/templates/main.tf
```diff
@@ -1,7 +1,23 @@
...
+  {{- end}}
+  {{- if .Values.openstack.userDomainName }}
+  tenant_name = "{{ .Values.openstack.userDomainName }}"
```
rfranzke:
```suggestion
  user_domain_name = "{{ .Values.openstack.userDomainName }}"
```

controllers/provider-openstack/charts/internal/openstack-infra/templates/main.tf
```diff
@@ -1,7 +1,23 @@
...
+  {{- end}}
+  {{- if .Values.openstack.userDomainID }}
+  tenant_name = "{{ .Values.openstack.userDomainID }}"
```
rfranzke:
```suggestion
  user_domain_id = "{{ .Values.openstack.userDomainID }}"
```

rfranzke:
Can you check why the CI is failing?

controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
```diff
@@ -240,6 +246,16 @@ func getConfigChartValues(
 		"requestTimeout":    cluster.CloudProfile.Spec.OpenStack.RequestTimeout,
 	}
 
+	// Only the 1.15 out of tree provider for OpenStack support the OS_DOMAIN_NAME/ID
+	// https://github.com/kubernetes/cloud-provider-openstack/pull/733
+	ok, err := gutils.CompareVersions(cluster.Shoot.Spec.Kubernetes.Version, ">=", "1.15")
```
rfranzke: We cannot merge this PR yet because there is no release of the OpenStack CCM, right?

controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
```diff
@@ -240,6 +246,16 @@ func getConfigChartValues(
...
+	if ok {
+		values = map[string]interface{}{
```
rfranzke: Why are you overwriting the complete `values` map?

controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
```diff
@@ -229,9 +231,13 @@ func getConfigChartValues(
 	values := map[string]interface{}{
 		"kubernetesVersion": cluster.Shoot.Spec.Kubernetes.Version,
 		"domainName":        c.DomainName,
+		"domainID":          c.DomainID,
 		"tenantName":        c.TenantName,
+		"tenantID":          c.TenantID,
 		"username":          c.Username,
 		"password":          c.Password,
+		"userDomainID":      c.UserDomainID,
```
rfranzke: Why are these added here? According to below comparison they are only valid for a certain Kubernetes version and should only be added if the version constraint matches?

controllers/provider-openstack/pkg/webhook/controlplanebackup/ensurer.go
```diff
@@ -138,6 +138,15 @@ func (e *ensurer) getBackupRestoreContainer(name string, cluster *extensionscont
 						},
 					},
 				},
+				{
+					Name: "OS_DOMAIN_ID",
+					ValueFrom: &corev1.EnvVarSource{
+						SecretKeyRef: &corev1.SecretKeySelector{
+							Key:                  openstack.DomainID,
```
rfranzke: OK, so adding these env variables always doesn't harm, right?
However, what happens if this key does not exist in the referenced secret? Will the pod come up anyways and then the value of the env variable will be `""`?

controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
afritzler: Yes I would wait. The merge should happen tomorrow.

controllers/provider-openstack/pkg/webhook/controlplanebackup/ensurer.go
afritzler: The values are non pointer strings so yes, they will become "". Setting those envs doesn't harm, as the gophercloud authenticator reads and evaluates the environment variables correctly. https://github.com/gophercloud/gophercloud/blob/master/openstack/auth_env.go#L36

controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
afritzler: That was a mistake. Wanted to add those keys.

controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
rfranzke: Even if it's merged we have to wait for a new release. Do you know how often they are releasing?

controllers/provider-openstack/pkg/webhook/controlplanebackup/ensurer.go
rfranzke: Hm, actually, the pod won't come up if the referenced keys don't exist in the secret. Just ran into this problem (cc @swapnilgm).

swapnilgm: Yes. It shows "ContainerConfigError".

afritzler: Ok, then I need to make this conditional.

controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
afritzler: No, the domain and tenant ID is available for old cloud provider versions as well.

controllers/provider-openstack/pkg/webhook/controlplanebackup/ensurer.go
afritzler: So this part need to be completely refactored. I added here secrets file and dir handling to the etcd-backup pod. That way we can just dump the correct secrets into the pod without having to fiddle around with the pod spec validation. The OpenStack authentication mechanics are then handling the authentication correctly. https://github.com/gardener/gardener-extensions/pull/287

afritzler: @swapnilgm I just updated the PR with the correct handling of `cloud.yaml` in the etcd-backup pod spec.

controllers/provider-openstack/pkg/webhook/controlplanebackup/ensurer.go
```diff
@@ -110,57 +110,21 @@ func (e *ensurer) getBackupRestoreContainer(name string, cluster *extensionscont
 			provider = openstack.StorageProviderName
 			env = []corev1.EnvVar{
 				{
-					Name: "STORAGE_CONTAINER",
+					Name: "OS_CLIENT_CONFIG_FILE",
```
swapnilgm: You will still have to pass bucketName via STORAGE_CONTAINER env.

afritzler: Good point.

controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
afritzler: So they are releasing not that often. However we will push for a release once all PRs are merged (still pending) on the cloud controller side. However we could move forward with this even if the CCM isn't available yet. You are just not allowed to set the user-domain-name. :)

rfranzke: What would happen if an end-user still specifies the user-domain-name field? Would the current CCM crash, or would simply happen "nothing"?

rfranzke:
@swapnilgm PTAL: https://github.com/gardener/gardener-extensions/pull/287/files#diff-97489784c4edd1edc4cc864e568ed184R50

hack/test.sh
```diff
@@ -24,4 +24,8 @@ SKIP_TESTS=$(echo controllers/provider-{alicloud,aws,azure,gcp,openstack,packet}
 
 header_text "Test"
 
+if [ $# -gt 0 ]; then
```
rfranzke: What is this thing doing? Why do we need it?

controllers/provider-openstack/pkg/webhook/controlplanebackup/ensurer.go
```diff
@@ -121,46 +121,19 @@ func (e *ensurer) getBackupRestoreContainer(name string, cluster *extensionscont
 					},
 				},
 				{
-					Name: "OS_AUTH_URL",
+					Name: "OS_CLIENT_CONFIG_FILE",
 					ValueFrom: &corev1.EnvVarSource{
```
swapnilgm:
```suggestion
					Value: path.Join("~/.config/openstack/", openstack.CloudYAML),
```

controllers/provider-openstack/pkg/webhook/controlplanebackup/ensurer.go
```diff
@@ -121,46 +121,19 @@ func (e *ensurer) getBackupRestoreContainer(name string, cluster *extensionscont
...
-				},
-				{
-					Name: "OS_TENANT_NAME",
-					ValueFrom: &corev1.EnvVarSource{
-						SecretKeyRef: &corev1.SecretKeySelector{
-							Key:                  openstack.TenantName,
+							Key:                  openstack.CloudYAMLKey,
 							LocalObjectReference: corev1.LocalObjectReference{Name: openstack.BackupSecretName},
 						},
 					},
```
swapnilgm:
```suggestion
					},
			}
			volumeMounts = []corev1.VolumeMount{
				{
					Name:      openstack.BackupSecretName,
					MountPath: "~/.config/openstack",
				},
			}
```

controllers/provider-openstack/pkg/controller/controlplane/valuesprovider.go
afritzler: The CCM will crashloop since the config file marshalling can't handle unknown properties. :(

swapnilgm:
Hey sorry for late comment, but I was not sure so didn't pitch it earlier. But thinking more about i am more into favour of replacing [this](https://github.com/gardener/gardener/blob/master/hack/templates/resources/70-deprecated-secret-cloudprovider.yaml.tpl#L76) and [this](https://github.com/gardener/gardener/blob/master/hack/templates/resources/45-deprecated-secret-seed-backup.yaml.tpl#L73) with 
```Yaml
  clouds.yaml: base64(clouds.yaml)
``` 
- The `clouds.yaml` is standard openstack [configuration file](https://docs.openstack.org/os-client-config/latest/user/configuration.html) format. 
-  With current approach we will have to parse the different fields in auth info and pass it at different place. There will be too many `if else` block repeated at many places. 
- I'm not sure does terraform, MCM support it yet, if not we can add support there.
- This is on similar thought of google `serviceaccount.json`.
- For aws or azure or alicloud there is not much complexity so its fine. (like having different combination of auth fields).

cc: @rfranzke @afritzler

kayrus:
> clouds.yaml: base64(clouds.yaml)

I'd rather use plain text.

> I'm not sure does terraformmer, MCM (cc @prashanth26 ) support it yet, if not we can add support there.
1.15.x does, but not all the auth options. All the options are supported in master branch, and will be supported in 1.16.x

afritzler:
On hold as we first need to add the OpenStack CSI support to the OpenStack provider.

rfranzke:
OK, please reopen the PR once it's ready to be reviewed and merged again. Thank you!

PR #290, opened 2019-08-23
Add URN field

vpnachev:
With this PR, the azure provider will use the `urn` of the VM images, instead of the quadruple `publisher,offer,sku,version`. 
The goal of this change is to allow map image versions, so that the provider will be able to advertise version `X` as version `Y`, just like rest of the providers are doing.

Fixes #

You could a see discussion about this here https://github.com/gardener/gardener/pull/1284#discussion_r313946117

This change requires a change in MCM, too. See https://github.com/gardener/machine-controller-manager/pull/326

**Release note**:

```improvement operator
`provider-azure` is now using URN to specify machine images.
```


vpnachev:
This PR is ready for review.
Note, the URN feature will be working only with MCM `>=0.22.0`(or higher, depending when https://github.com/gardener/machine-controller-manager/pull/326 will be released), otherwise it will be using the current behaviour.

vpnachev:
The contributed defaulting of CRD resources was removed.

controllers/provider-azure/pkg/controller/worker/machine_images.go
```diff
@@ -61,28 +61,28 @@ func (w *workerDelegate) GetMachineImages(ctx context.Context) (runtime.Object,
 	return workerStatusV1alpha1, nil
 }
 
-func (w *workerDelegate) findMachineImage(name, version string) (publisher, sku, offer string, err error) {
+func (w *workerDelegate) findMachineImage(name, version string) (publisher, sku, offer, urn string, err error) {
 	machineImage, err := confighelper.FindImage(w.machineImageMapping, name, version)
 	if err == nil {
-		return machineImage.Publisher, machineImage.SKU, machineImage.Offer, nil
+		return machineImage.Publisher, machineImage.SKU, machineImage.Offer, *machineImage.URN, nil
```
rfranzke: What if URN is `nil`?

controllers/provider-azure/pkg/controller/worker/machine_images.go
```diff
@@ -61,28 +61,28 @@ func (w *workerDelegate) GetMachineImages(ctx context.Context) (runtime.Object,
...
 		}
 
 		machineImage, err := apisazurehelper.FindMachineImage(workerStatus.MachineImages, name, version)
 		if err != nil {
-			return "", "", "", errorMachineImageNotFound(name, version)
+			return "", "", "", "", errorMachineImageNotFound(name, version)
 		}
 
-		return machineImage.Publisher, machineImage.SKU, machineImage.Offer, nil
+		return machineImage.Publisher, machineImage.SKU, machineImage.Offer, *machineImage.URN, nil
```
rfranzke: Same question like above

controllers/provider-azure/pkg/controller/worker/machine_images.go
vpnachev: nice catch

PR #294, opened 2019-08-26
Decrease waiting times for worker + infrastructure controllers

rfranzke:
Decrease waiting times for worker + infrastructure controllers to allow faster propagation of errors. It's a follow-up of gardener/gardener#1364.

* /cc @ggaurav10 
* I had to revendor gardener/gardener and adapted a few breaking changes.

**Release note**:

```improvement user
The time the `Worker` and `Infrastructure` extension controllers are waiting for their operations to finish has been decreased in order to allow a faster error propagation to end-users.
```


ialidzhikov:
Could you please check `./hack/install.sh`?

rfranzke:
Done, please check again.

pkg/gardener/terraformer/terraformer.go
```diff
@@ -31,6 +33,31 @@ func (t *terraformer) SetVariablesEnvironment(tfVarsEnvironment map[string]strin
 	return &terraformer{t.tf.SetVariablesEnvironment(tfVarsEnvironment)}
 }
 
+// SetJobBackoffLimit implements Terraformer.
+func (t *terraformer) SetJobBackoffLimit(val int32) Interface {
```
ialidzhikov: I think we can improve the terraform options handling in g/gardener with the usage of [functional options](https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md).
However it is out of context of this PR.

pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -142,7 +142,7 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 	}
 
 	// Wait until all generated machine deployments are healthy/available.
-	timeoutCtx, cancel := context.WithTimeout(ctx, 30*time.Minute)
+	timeoutCtx, cancel := context.WithTimeout(ctx, 15*time.Minute)
```
ialidzhikov: With gardener/gardener#1364 we now have:
```golang
const WorkerDefaultTimeout = 5 * time.Minute
```

We wait 5min for worker to be `Ready`, and the worker waits 15 min for its machdeploys to be `Available`?
The same question applies also for the deletion.

pkg/gardener/terraformer/terraformer.go
rfranzke: Yes, good idea.

pkg/controller/worker/genericactuator/actuator_reconcile.go
rfranzke: Hm, yeah, good point, thank you. Set it to `5m` now, if it doesn't complete within the `5m` then it will be retried by the controller anyway and another `5m` will be waited. However, if an error occurred, it will now be propagated to the user very fast.

PR #310, opened 2019-09-10
Enable customized images for AliCloud

### LINKED ISSUE #309: [AliCloud] Support Customized image
CoreOS is not well supported by AliCloud. There might be security issues of outdated CoreOS version. We need to be assure Gardener clusters running on a latest CoreOS. A cusomized image will be a solution.

After adaption:
- Customized image IDs in different regions are different
- By default, a customized image is private. That means, for other sub-account, there needs some extra configuration for this account to consume customized image
- Since we can do everything in a customized image, we can support cloud-config when a VM is bootstrapped. That means we could use standard coreos extension to boot a node.
- We continue to support CoreOS image provided by AliCloud. As it is not updated version, it is not suggested to use in production.

jia-jerry:
Support customized Image on AliCloud
Fixes #309 

Support region and VM image ID for AliCloud. Corresponding configuration will be provide in landscape setup tools.
**Release note**:

```action operator
From this version, for AliCloud provider extensions, you need to configure available image ID for each region. Please refer to the [example configuration](https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-alicloud/example/00-componentconfig.yaml#L9) used by AliCloud provider extensions. If you are using a customised image: It is required to to make sure that the customised image exists in supported regions and that the customised images is accessible by Shoot cluster account. 
```


rfranzke:
How will the "migration" look like? Today, the `Worker` status does not contain the region. Will the worker controller still be able to find the image for the provided region? Can you please write a proper release note that action from the administrator is required in order to change the Alicloud config accordingly?

jia-jerry:
> How will the "migration" look like? Today, the `Worker` status does not contain the region. Will the worker controller still be able to find the image for the provided region?
No need to add region field in Worker. There is a logic for controller to find mapped images ID from the configmap *gardener-extension-provider-alicloud-configmap*. I tested OK with following senario:
Create a shoot with old version of alicloud-provider-extension.
Upgrade new version of  alicloud-provider-extension.
Increase replica in machinedeployment.

rfranzke:
Okay, alright. Please fix the release note, I think it's good to go then.

go.mod
```diff
@@ -56,6 +56,7 @@ require (
 )
 
 replace (
+	git.apache.org/thrift.git => github.com/apache/thrift v0.12.0
```
ialidzhikov: Minor nit - this should be no longer needed.

jia-jerry:
@rfranzke , just removed region field in worker just like AWS provider. Release note was updated.

PR #316, opened 2019-09-16
Check for ProviderConfig in CP Actuator to prevent manual intervention when updating

danielfoehrKn:
Updating Extensions & Gardener without manual intervention.

I recently updated my Gardener Setup using the newest GardenerAPi Server & ControllerManager & newest provider extension.
I already had a ControlPlane & ControlPlaneExposure resources for my shoot.

**Problem:**

- GardenerController Manager reconcile fails waiting for succesful reconcile of the ControlePlane resource
-  gcp provider - controlplane actuator ran into nil pointer exception reconciling the ControlePlane Exposure 

- Provider is in CrashLoop and never reconciles the actual ControlePlane -> Gardener reconcile fails before it can update the ControlePlaneExposure -> deadlock

**Cause:**
- the newest provider distinguishes between ControlePlane & ControlePlane Exposure via the Spec.Purpose field
- ControlePlane Exposure is not yet updated by GardnerControllerManager to include the Purpose field -> will be reconciled as normal ControlePlane -> nil pointer

**Solution**

Check for ProviderConfig in ControlePlane

Other workaround:
- delete the ControlePlaneExposure via & restart provider

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/controllers/provider-alicloud/pkg/controller/controlplane/valuesprovider.go
+++ b/controllers/provider-alicloud/pkg/controller/controlplane/valuesprovider.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"fmt"
 	"path/filepath"
 
 	"github.com/gardener/gardener-extensions/controllers/provider-alicloud/pkg/alicloud"
@@ -218,6 +219,9 @@ func (vp *valuesProvider) GetConfigChartValues(
 	cp *extensionsv1alpha1.ControlPlane,
 	cluster *extensionscontroller.Cluster,
 ) (map[string]interface{}, error) {
+	if cp.Spec.ProviderConfig == nil {
+		return nil, fmt.Errorf("cannot decode nil providerConfig of ControlePlane '%s'", util.ObjectName(cp))
+	}
 	// Decode providerConfig
 	cpConfig := &apisalicloud.ControlPlaneConfig{}
 	if _, _, err := vp.decoder.Decode(cp.Spec.ProviderConfig.Raw, nil, cpConfig); err != nil {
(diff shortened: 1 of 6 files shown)


ialidzhikov:
Could you please check `make test`?

pkg/controller/controlplane/genericactuator/actuator.go
```diff
@@ -165,6 +165,9 @@ func (a *actuator) Reconcile(
 	if cp.Spec.Purpose != nil && *cp.Spec.Purpose == extensionsv1alpha1.Exposure {
 		return a.reconcileControlPlaneExposure(ctx, cp, cluster)
 	}
+	if cp.Spec.ProviderConfig == nil {
```
rfranzke: We should not enforce this in the generic actuator. It's a provider-specific decision whether this is needed or not.

rfranzke: any follow up @danielfoehrKn ?

controllers/provider-alicloud/pkg/controller/controlplane/valuesprovider.go
```diff
@@ -218,6 +219,9 @@ func (vp *valuesProvider) GetConfigChartValues(
 	cp *extensionsv1alpha1.ControlPlane,
 	cluster *extensionscontroller.Cluster,
 ) (map[string]interface{}, error) {
+	if cp.Spec.ProviderConfig == nil {
```
rfranzke: Instead  of throwing an error, can you simply default the providerConfig to an empty object if it wasn't provided? The functions that actually use it should validate that the required fields are set/return an error in this case.

Something like

```golang
cpConfig := &apisalicloud.ControlPlaneConfig{}
if cp.Spec.ProviderConfig != nil && cp.Spec.ProviderConfig.Raw != nil {
	if _, _, err := vp.decoder.Decode(cp.Spec.ProviderConfig.Raw, nil, cpConfig); err != nil {
		return nil, errors.Wrapf(err, "could not decode providerConfig of controlplane '%s'", util.ObjectName(cp))
	}
}
```

WDYT?

danielfoehrKn: Makes sense, although I need to make sure that this validation happens at a later point in time in each provider.

rfranzke: Yes, I think this would be a more valuable PR. In the end this will allow end-users to e.g. not specify the `ControlPlaneConfig` in their `Shoots`.
Today, for certain providers (like AWS, for example), they **have** to specify

```yaml
apiVersion: aws.provider.extensions.gardener.cloud/v1alpha1
kind: ControlPlaneConfig
```

which basically has no content.

rfranzke:
Please reopen once adapted according to suggestions.

PR #322, opened 2019-09-19
Control plane generic actuator waits for managed resources to be deleted

rfranzke:
We have seen situations in which the shoot namespace in the seed was
hanging in "Terminating" state because the managed resources from the
control plane controller were still existing.
In the [shoot deletion flow](https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/shoot/shoot_control_delete.go)
the `ControlPlane` resource is destroyed, and right after that the
kube-apiserver deployment is deleted. If the gardener-resource-manager
is not cleaning up the control plane MRs fast enough then the
kube-apiserver is deleted to fast which will block the
gardener-resource-manager from gracefully cleaning up the MRs.
Thus, now with the waiting mechanism we ensure that the `ControlPlane`
resource is only deleted after the MRs are gone, hence, the
kube-apiserver deployment will also only be deleted after that.

**Release note**:

```improvement operator
The `ControlPlane` generic actuator now waits for its created `ManagedResource`s to be deleted during its deletion flow. This is so that it doesn't accidentally remove its finalizer although created artefacts still exist.
```

### DIFF
--- a/pkg/controller/controlplane/genericactuator/actuator.go
+++ b/pkg/controller/controlplane/genericactuator/actuator.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"time"
 
 	extensionscontroller "github.com/gardener/gardener-extensions/pkg/controller"
 	"github.com/gardener/gardener-extensions/pkg/controller/controlplane"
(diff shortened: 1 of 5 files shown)


pkg/controller/controlplane/genericactuator/actuator.go
```diff
@@ -388,11 +389,20 @@ func (a *actuator) deleteControlPlane(
...
 		return errors.Wrapf(err, "could not delete managed resource containing shoot chart for controlplane '%s'", util.ObjectName(cp))
 	}
 
+	timeoutCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancel()
+
+	if err := extensionscontroller.WaitUntilManagedResourceDeleted(timeoutCtx, a.client, cp.Namespace, storageClassesChartResourceName); err != nil {
+		return errors.Wrapf(err, "error while waiting for managed resource containing storage classes chart for controlplane '%s' to be deleted", util.ObjectName(cp))
+	}
+	if err := extensionscontroller.WaitUntilManagedResourceDeleted(timeoutCtx, a.client, cp.Namespace, controlPlaneShootChartResourceName); err != nil {
```
timuthy: I don't think it's a good idea to reuse `timeoutCtx` since `WaitUntilManagedResourceDeleted` blocks and quite some time has already passed from the previous wait. You could either wait in parallel or use a new timeout context.

pkg/controller/controlplane/genericactuator/actuator.go
```diff
@@ -422,6 +432,9 @@ func (a *actuator) deleteControlPlane(
 		if err := extensionscontroller.DeleteManagedResource(ctx, a.client, cp.Namespace, shootWebhooksResourceName); err != nil {
 			return errors.Wrapf(err, "could not delete managed resource containing shoot webhooks for controlplane '%s'", util.ObjectName(cp))
 		}
+		if err := extensionscontroller.WaitUntilManagedResourceDeleted(timeoutCtx, a.client, cp.Namespace, shootWebhooksResourceName); err != nil {
```
timuthy: See previous comment concerning `timeoutCtx`.

pkg/controller/controlplane/genericactuator/actuator.go
```diff
@@ -422,6 +432,9 @@ func (a *actuator) deleteControlPlane(
 		if err := extensionscontroller.DeleteManagedResource(ctx, a.client, cp.Namespace, shootWebhooksResourceName); err != nil {
```
timuthy: Regarding your PR description: Don't you think that it's also necessary to delete the this ManagedResource before deleting the control plane resources?

PR #331, opened 2019-09-23
Azure: Update deprecated terraform resources

dkistner:
The bindings of route tables and security groups directly to subnets is deprecated and will be removed in the next major version of the Azurerm terraform provider.
As replacement there are new association resources to bind route tables and security groups to a subnet. Migration will work out of the box.

**Which issue(s) this PR fixes**:
Fixes partly #320 

**Release note**:

```improvement operator
NONE
```

PR #332, opened 2019-09-23
Azure: Allow to configure service endpoints

### LINKED ISSUE #199: Enable Virtual network service endpoints for Azure
**What would you like to be added**:
[Virtual network service endpoints for Azure](https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-service-endpoints-overview) support.

Virtual network service endpoints for Azure requires a specific flag to be set on the subnet (not on VNET, on Subnet) of the nodes. Currently we can provide an existing VNET but Subnet is always created and managed by Gardener. In our case we want the cluster nodes to be able to use this feature and so related Subnet needs this flag set on. But the subnet is configured by Gardener, and our changes are being reverted during reconciles. 

We've faced this issue on Gardener 0.18, but verified that the same issue exists in new extension implementation: https://github.com/gardener/gardener-extensions/blob/c4c502de484d9ef6333d4f15c807d96740086550/controllers/provider-azure/charts/internal/azure-infra/templates/main.tf#L28

These are the options I can see with my brief investigation:
1. enable passing an existing subnet along with vnet (also mind #136 since it seems related), so we can take over the the vnet and subnet management from gardener and provide them when creating the shoot,
2. enable providing service_endpoints which will be used in terraformer if provided (https://www.terraform.io/docs/providers/azurerm/r/subnet.html#service_endpoints), since this will also allow doing on-the-fly changes for existing clusters. We have many clusters older than 1y, we dont want to reprovision to enable this feature.

I personally go for option 2, but open to discussion.

**Why is this needed**:

We have users blocked on moving in because they've been using Azure PostgreSQL. When not using VNET endpoint for PostgreSQL:
* there is a performance impact, on DB operations
* one has to whitelist all egress IPs of the cluster on the postgres service and its not easy to set kubernetes cluster's egress IP on Azure when using multiple non-static Service[type=LoadBalancers].

https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-service-endpoints-overview#key-benefits

dkistner:
Allow to configure Azure service endpoints for Gardener managed subnets.
See the Azure documentation for service endpoints: https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-service-endpoints-overview

If an `Infrastructure` CR reference an invalid service endpoint then Terraform will return an `SubnetHasServiceEndpointWithInvalidServiceName` error code.

See #199

Fixes #199 

**Release note**:

```improvement user
The Gardener Azure provider extension support now subnets configured with Azure Service Endpoints.
```

cc @bergerx

### DIFF
--- a/controllers/provider-azure/pkg/apis/azure/v1alpha1/types_infrastructure.go
+++ b/controllers/provider-azure/pkg/apis/azure/v1alpha1/types_infrastructure.go
@@ -42,6 +42,9 @@ type NetworkConfig struct {
 	VNet VNet `json:"vnet"`
 	// Workers is the worker subnet range to create (used for the VMs).
 	Workers string `json:"workers"`
+	// ServiceEndpoints is a list of Azure ServiceEndpoints which should be associated with the worker subnet.
+	// +optional
+	ServiceEndpoints []string `json:"serviceEndpoints,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/controllers/provider-azure/charts/internal/azure-infra/templates/main.tf
+++ b/controllers/provider-azure/charts/internal/azure-infra/templates/main.tf
@@ -30,6 +30,7 @@ resource "azurerm_subnet" "workers" {
   resource_group_name       = "{{ required "resourceGroup.name is required" .Values.resourceGroup.name }}"
   virtual_network_name      = "{{ required "resourceGroup.vnet.name is required" .Values.resourceGroup.vnet.name }}"
   address_prefix            = "{{ required "networks.worker is required" .Values.networks.worker }}"
+  service_endpoints         = [{{range $index, $serviceEndpoint := .Values.resourceGroup.subnet.serviceEndpoints}}{{if $index}},{{end}}"{{$serviceEndpoint}}"{{end}}]
 }
 
 resource "azurerm_route_table" "workers" {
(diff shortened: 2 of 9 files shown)


dkistner:
Let's wait before #331 is merged, then this PR needs a rebase.
Just opened it for review upfront.

dkistner:
Rebased as #331 is merged.

controllers/provider-azure/pkg/apis/azure/v1alpha1/types_infrastructure.go
```diff
@@ -42,6 +42,8 @@ type NetworkConfig struct {
 	VNet VNet `json:"vnet"`
 	// Workers is the worker subnet range to create (used for the VMs).
 	Workers string `json:"workers"`
+	// ServiceEndpoints is a list of Azure ServiceEndpoints which should be associated with the worker subnet.
+	ServiceEndpoints []string `json:"serviceEndpoints"`
```
rfranzke:
```suggestion
	// +optional
	ServiceEndpoints []string `json:"serviceEndpoints,omitempty"`
```

rfranzke:
Before this can be used with gardener you have to revendor the extensions, and extend the `garden.sapcloud.io/v1beta1.Shoot` API and add a new field in the `spec.cloud.azure` spec for the service endpoints. You also have to write proper conversion for `garden.sapcloud.io/v1beta1.Shoot <-> core.gardener.cloud/v1alpha1.Shoot`.

PR #335, opened 2019-09-24
enable watch for autoscaler cluster roles

zanetworker:
enable the horizontal typha autoscaler to watch nodes. 

```improvement operator
NONE
```


rfranzke:
Why is this added? It's not needed AFAIK?

PR #369, opened 2019-10-14
extension-shoot-cert-service: updated image tag for cert-management

MartinWeindel:
Updated cert-management image tag to 0.2.3.
With this new version, the watches on issuers and secrets on the default cluster 
are restricted to the issuer namespace. 

**Release note**:

```improvement operator
extension-shoot-cert-service: The cert-controller-manager only watches
issuers and secrets on the default cluster in the issuer namespace.
```


rfranzke:
closed in favour of #370

PR #374, opened 2019-10-16
Add IPIP and IP Autodetection Method for calico controller

mkoynov:

Makes sure IP_AUTODETECTION_METHOD env var for calico/node can be used. Makes CALICO_IPV4POOL_IPIP env var configurable for calico/node. Removes redundant FELIX_IPINIPENABLED env var from calico/node.

Includes unit tests

**Release note**:

```noteworthy user
The configuration for the `networking-calico` controller now allows to explicitly set `.spec.ipip`. It will be defaulted to `Always`.
```

### DIFF
--- a/controllers/networking-calico/charts/internal/calico/templates/calico.yaml
+++ b/controllers/networking-calico/charts/internal/calico/templates/calico.yaml
@@ -184,17 +184,16 @@ spec:
             - name: CALICO_IPV4POOL_CIDR
               value: "{{ .Values.global.podCIDR }}"
             # Enable IPIP
-            {{- if ne .Values.config.backend "none"}}
-            # Enable IP-in-IP within Felix.
-            - name: FELIX_IPINIPENABLED
-              value: "true"
-            - name: CALICO_IPV4POOL_IPIP
-              value: "Always"
-            {{- else }}
+            {{- if eq .Values.config.backend "none"}}
             - name: FELIX_IPINIPENABLED
               value: "false"
             - name: CALICO_IPV4POOL_IPIP
               value: "Never"
+            {{- else }}
+            - name: FELIX_IPINIPENABLED
+              value: "true"
+            - name: CALICO_IPV4POOL_IPIP
+              value: "{{.Values.ipip}}"
             {{- end }}
             # Choose the backend to use.
             - name: CALICO_NETWORKING_BACKEND
--- a/controllers/networking-calico/pkg/apis/calico/types_network.go
+++ b/controllers/networking-calico/pkg/apis/calico/types_network.go
@@ -25,6 +25,15 @@ const (
 	None Backend = "none"
 )
 
+type IPIP string
+
+const (
+	Always      IPIP = "Always"
+	Never       IPIP = "Never"
+	CrossSubnet IPIP = "CrossSubnet"
+	Off         IPIP = "Off"
+)
+
 type CIDR string
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
@@ -32,7 +41,7 @@ type CIDR string
 // NetworkConfig configuration for the calico networking plugin
 type NetworkConfig struct {
 	metav1.TypeMeta
-	// Backend defines whether a backend should be used or not (e.g., bird or None)
+	// Backend defines whether a backend should be used or not (e.g., bird or none)
 	Backend Backend
 	// IPAM to use for the Calico Plugin (e.g., host-local or Calico)
 	// +optional
@@ -41,6 +50,9 @@ type NetworkConfig struct {
 	// https://docs.projectcalico.org/v2.2/reference/node/configuration#ip-autodetection-methods
 	// +optional
 	IPAutoDetectionMethod *string
+	// IPIP is the IPIP Mode for the IPv4 Pool (e.g. Always, Never, CrossSubnet)
+	// +optional
+	IPIP *IPIP
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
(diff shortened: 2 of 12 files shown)


rfranzke:
Thanks for the PR!
@zanetworker can you please take a thorough look?

controllers/networking-calico/pkg/charts/utils.go
```diff
@@ -54,7 +55,10 @@ func ComputeCalicoChartValues(network *extensionsv1alpha1.Network, config *calic
 	)
 
 	if config != nil {
-		calicoConfigValues["backend"] = config.Backend
+		switch config.Backend {
+		case calicov1alpha1.Bird, calicov1alpha1.None:
+			calicoConfigValues["backend"] = config.Backend
+		}
```
zanetworker: This is only for validation correct?

controllers/networking-calico/pkg/charts/utils.go
```diff
@@ -64,6 +68,15 @@ func ComputeCalicoChartValues(network *extensionsv1alpha1.Network, config *calic
...
 		}
+
+		if config.IPAutoDetectionMethod != nil {
+			calicoChartValues["ipAutodetectionMethod"] = *config.IPAutoDetectionMethod
+		}
+
+		switch config.IPIP {
+		case calicov1alpha1.Always, calicov1alpha1.Never, calicov1alpha1.Off, calicov1alpha1.CrossSubnet:
+			calicoChartValues["ipip"] = config.IPIP
+		}
```
zanetworker: This is only for valdiation, correct?

controllers/networking-calico/charts/internal/calico/templates/calico.yaml
```diff
@@ -184,17 +184,12 @@ spec:
             - name: CALICO_IPV4POOL_CIDR
               value: "{{ .Values.global.podCIDR }}"
             # Enable IPIP
-            {{- if ne .Values.config.backend "none"}}
-            # Enable IP-in-IP within Felix.
-            - name: FELIX_IPINIPENABLED
```
rfranzke: Why was `FELIX_IPINIPENABLED` removed? It's never set anymore?

controllers/networking-calico/pkg/apis/calico/types_network.go
```diff
@@ -41,6 +50,9 @@ type NetworkConfig struct {
 	// https://docs.projectcalico.org/v2.2/reference/node/configuration#ip-autodetection-methods
 	// +optional
 	IPAutoDetectionMethod *string
+	// IPIP is the IPIP Mode for the IPv4 Pool (e.g. Always, Never, CrossSubnet)
+	// +optional
+	IPIP IPIP
```
rfranzke: If this shall be optional then make it a pointer type:

```suggestion
	IPIP *IPIP
```

controllers/networking-calico/pkg/apis/calico/v1alpha1/types_network.go
```diff
@@ -43,6 +52,9 @@ type NetworkConfig struct {
 	// https://docs.projectcalico.org/v2.2/reference/node/configuration#ip-autodetection-methods
 	// +optional
 	IPAutoDetectionMethod *string `json:"ipAutodetectionMethod,omitempty"`
+	// IPIP is the IPIP Mode for the IPv4 Pool (e.g. Always, Never, CrossSubnet)
+	// +optional
+	IPIP IPIP `json:"ipip,omitempty"`
```
rfranzke: Same suggestion as above.

rfranzke:
Also, CI is failing:

```
Check
Executing check-generate
go generate needs to be run:
 M controllers/networking-calico/hack/api-reference/calico.md
```

controllers/networking-calico/pkg/apis/calico/types_network.go
mkoynov: OK

controllers/networking-calico/pkg/apis/calico/v1alpha1/types_network.go
mkoynov: OK

controllers/networking-calico/pkg/charts/utils.go
mkoynov: Correct

controllers/networking-calico/pkg/charts/utils.go
mkoynov: Yes, it is

mkoynov: Those two validations should not be needed if we add two validating webhooks on Network but I felt that would be overkill?

rfranzke: We are planning to support validating webhooks soon, and we should implement one for Calico as well. The configuration may be tampered by the end-users as its exposed in the `Shoot`, hence, we should prevent that some undesired values are sent to us. 
For now, I guess "this validation" is ok because we don't have the webhooks yet, but as soon as we have them we should implement it properly.

controllers/networking-calico/charts/internal/calico/templates/calico.yaml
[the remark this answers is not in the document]
mkoynov: Yes, because of exactly that. It is automatically set to `true` when an IPIP-enabled pool is created. Do you have a case where where we want to set an IPIP pool to disabled while ` FELIX_IPINIPENABLED` is true or the other way around?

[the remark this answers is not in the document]
mkoynov: Very well

zanetworker (approved):

rfranzke (approved):

PR #376, opened 2019-10-17
Refactored Webhook Ensurer API

afritzler:
Co-authored-by: Uwe Krueger <uwe.krueger@sap.com>

This PR aligns the signature of the ensurer webhook APIs and simplifies further extensions with additional fields via wrapper context.

**Release note**:
```improvement developer
Refactored Webhook Ensurer API
```

### DIFF
--- a/pkg/webhook/controlplane/genericmutator/mutator.go
+++ b/pkg/webhook/controlplane/genericmutator/mutator.go
@@ -40,37 +40,36 @@ import (
 
 // EnsurerContext wraps the actual context and cluster object.
 type EnsurerContext interface {
-	GetContext() context.Context
-	GetCluster() (*extensionscontroller.Cluster, error)
+	GetCluster(ctx context.Context) (*extensionscontroller.Cluster, error)
 }
 
 // Ensurer ensures that various standard Kubernets controlplane objects conform to the provider requirements.
 // If they don't initially, they are mutated accordingly.
 type Ensurer interface {
 	// EnsureKubeAPIServerService ensures that the kube-apiserver service conforms to the provider requirements.
-	EnsureKubeAPIServerService(EnsurerContext, *corev1.Service) error
+	EnsureKubeAPIServerService(context.Context, EnsurerContext, *corev1.Service) error
 	// EnsureKubeAPIServerDeployment ensures that the kube-apiserver deployment conforms to the provider requirements.
-	EnsureKubeAPIServerDeployment(EnsurerContext, *appsv1.Deployment) error
+	EnsureKubeAPIServerDeployment(context.Context, EnsurerContext, *appsv1.Deployment) error
 	// EnsureKubeControllerManagerDeployment ensures that the kube-controller-manager deployment conforms to the provider requirements.
-	EnsureKubeControllerManagerDeployment(EnsurerContext, *appsv1.Deployment) error
+	EnsureKubeControllerManagerDeployment(context.Context, EnsurerContext, *appsv1.Deployment) error
 	// EnsureKubeSchedulerDeployment ensures that the kube-scheduler deployment conforms to the provider requirements.
-	EnsureKubeSchedulerDeployment(EnsurerContext, *appsv1.Deployment) error
+	EnsureKubeSchedulerDeployment(context.Context, EnsurerContext, *appsv1.Deployment) error
 	// EnsureETCDStatefulSet ensures that the etcd stateful sets conform to the provider requirements.
-	EnsureETCDStatefulSet(EnsurerContext, *appsv1.StatefulSet) error
+	EnsureETCDStatefulSet(context.Context, EnsurerContext, *appsv1.StatefulSet) error
 	// EnsureKubeletServiceUnitOptions ensures that the kubelet.service unit options conform to the provider requirements.
-	EnsureKubeletServiceUnitOptions(EnsurerContext, []*unit.UnitOption) ([]*unit.UnitOption, error)
+	EnsureKubeletServiceUnitOptions(context.Context, EnsurerContext, []*unit.UnitOption) ([]*unit.UnitOption, error)
 	// EnsureKubeletConfiguration ensures that the kubelet configuration conforms to the provider requirements.
-	EnsureKubeletConfiguration(EnsurerContext, *kubeletconfigv1beta1.KubeletConfiguration) error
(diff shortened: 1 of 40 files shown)


zanetworker:
Just a side note, did you have a look at the following: 

- https://golang.org/pkg/context/ -> especially > `Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx`
- https://github.com/golang/go/issues/22602 - > TLDR; wrapped context are not recommended in exported functions

rfranzke:
+1, this would also be my comment. The `ensurerContext` struct is fine, but please keep the `context.Context` the first argument of each function. I like the idea of the `ensurerContext` containing the cluster, client, and object information, so +1 for this, thank you!

afritzler:
Valid point. I will remove the context from the `EnsurerContext` wrapper.

jia-jerry:
This is a nice refactor. It will ease the whole world for webhook to get more information about the cluster. Looking forward your PR.

afritzler:
Thanks @jia-jerry! We also need it in order to add support for OpenStack CSI since we need to distinguish between different the Kubernetes version in a couple of components.

pkg/webhook/controlplane/genericmutator/mutator.go
```diff
@@ -102,6 +108,39 @@ func (m *mutator) InjectClient(client client.Client) error {
...
+// NewEnsurerContext creates a ensurer context object.
+func NewEnsurerContext(client client.Client, object metav1.Object) EnsurerContext {
+	return &ensurerContext{
+		client: client,
+		object: object,
+	}
+}
+
+// NewInternalEnsurerContext creates a ensurer context object.
+func NewInternalEnsurerContext(cluster *extensionscontroller.Cluster) EnsurerContext {
```
rfranzke: Why do we need this function? Also, why is it called "internal"? What shall this mean?

pkg/webhook/controlplane/genericmutator/mutator.go
```diff
@@ -102,6 +108,39 @@ func (m *mutator) InjectClient(client client.Client) error {
...
+	return &ensurerContext{
+		cluster: cluster,
+	}
+}
+
+// GetCluster returns the cluster object.
+func (c *ensurerContext) GetCluster(ctx context.Context) (*extensionscontroller.Cluster, error) {
```
rfranzke: Why don't we simply fetch it always and inject it into the ensurer context by default? Then the various webhook implementors don't need to call this function and handle the error, etc. WDYT?

afritzler: We wanted to lazy load the `cluster` object as needed.

pkg/webhook/controlplane/genericmutator/mutator.go
afritzler: Well, this is primarily needed for the tests.

zanetworker: Maybe we can use only `NewEnsurerContext` for everything and pass `nil` for un-necessary values. Instead of calling `NewInternalContext` to just get the cluster just call `NewEnsurerContext` with a nil value for the client but have always one instantiation function ?

pkg/webhook/controlplane/genericmutator/mutator.go
```diff
@@ -102,6 +108,39 @@ func (m *mutator) InjectClient(client client.Client) error {
 	return nil
 }
 
+type ensurerContext struct {
+	client  client.Client
+	object  metav1.Object
+	cluster *extensionscontroller.Cluster
+}
+
+// NewEnsurerContext creates a ensurer context object.
```
zanetworker:
```suggestion
// NewEnsurerContext creates an  ensurerContext object.
```

pkg/webhook/controlplane/genericmutator/mutator.go
afritzler: Not sure why this is a problem. The internal one is for our tests only as we need to mock a cluster object into the the whole thing in order to run some of the tests.

zanetworker: To me it is a little confusing to have exposed Initializers for the same object just to fill in different fields for a test. My proposal was to have one initializer, later in the test you would do something like:  
```go 
dummyContext = genericmutator.NewEnsurerContext(nil, nil, cluster)
``` 
Also Just my opinion :)

pkg/webhook/controlplane/genericmutator/mutator.go
afritzler: Thx for the hint here!

pkg/webhook/controlplane/genericmutator/mutator.go
zanetworker: sounds better to me, we had a similar approach here https://github.com/gardener/gardener-resource-manager/blob/7ceeb30219932467612119392067cacabb9aaf6f/pkg/manager/managedresources.go#L33. 

`NewEnsurerContext().WithCluster(..).WithClient(...)`

mandelsoft: A function signature should always be oriented on the intended semantic. For the API function this semantic explicitly does NOT allow to set a cluster object, therefor the signature has no such argument. Only for the test purposes we need to create fake contexts beside the official API usage (White box testing). Therfore we added an additional function for this purpose not intended be part of the Official API.

zanetworker: Sure, but here if I am an external user to the API and I find this function, I would not know that it is intended only for tests, also the name 'Internal' contradicts with the function scope. I guess for such use-case an `internal` package can be used to prevent exposing the function to other external actors. Or simply re-use the same initializer. 

That said, this is just an opinion on my side, feel free to ignore if you don't think it adds value to your PR :)

rfranzke: I'm okay getting it in as it is, there are also other PRs waiting on this one. @zanetworker also ok?

afritzler: Also, can't we just refactor the test later on? There seems to be different opinions on how to do that. Maybe we should discuss first?

zanetworker: Sure, please go ahead.

PR #379, opened 2019-10-21
Enable cloud-nat for GCP

zanetworker:
This PR adds CloudNat as a replacement to public IPs for GCP, this is to be consistent with other cloud providers and to provide more security not to have instances exposed to the outside. 

- Fixes  https://github.com/gardener/machine-controller-manager/issues/339
- Depends on  https://github.com/gardener/machine-controller-manager/pull/340 

The MCM [PR](https://github.com/gardener/machine-controller-manager/pull/340), otherwise, CloudNAT and public IPs will co-exist. 

**Release note**:

```action user
CloudNAT is now used instead of public IPs for GCP instances, please don't use the public IP of the VMs directly to access the VMs, instead use Loadbalancers or a bastion hosts to SSH.
```


prashanth26:
Hi @zanetworker ,

I apologize for the late changes, but we have inverted the logic for external IP attachment to maintain backward compatibility for existing MCM users. You can probably ignore the machineAPI changes to this PR. I shall create another PR on Gardener to adopt the new APIs and fix them. In the meanwhile this PR could just introduce the cloud NAT.

controllers/provider-gcp/pkg/controller/worker/machines.go
```diff
@@ -123,6 +123,7 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 				"region":             w.worker.Spec.Region,
 				"zone":               zone,
 				"canIpForward":       true,
+				"attachExternalIP":   false,
```
rfranzke: As discussed, this will roll all GCP machines in the next reconciliation.
/cc @vlerenc

zanetworker:
as discussed with @prashanth26, a small change needs to incorporated to use "disableExternalIP" instead of "attachExternalIP" as MCM needs to have it enabled by default for external users of MCM. 

/cc @rfranzke

prashanth26:
Also a quick reminder to the vendor in the new MachineAPIs into the `Gardener-extensions` with the `disableExternalIP` field. Also, handle the same in worker pool controller.

rfranzke:
Re-vendoring is not needed I guess because we don't work with the MCM types but only use the YAML specification.

prashanth26:
Okay, then I maybe misunderstood the flow earlier. Now it makes sense. Re vendoring might not be required then.

vlerenc:
Do you have more details @zanetworker? Is the plan roughly https://github.com/gardener/machine-controller-manager/issues/339#issuecomment-540979928 and will we basically "migrate" from external IPs to NAT GWs? If so and because of @rfranzke comment https://github.com/gardener/gardener-extensions/pull/379?email_source=notifications&email_token=ABF6M6Y4DS7TKYB72AWQQY3QPWA7BA5CNFSM4JCYHYUKYY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOCIS5MXY#pullrequestreview-304469599, please inform the end users in advance:
- Your cluster will be migrated from external IPs to NAT GWs and any code/processes/solutions that depend on external IPs will fail (e.g. opening firewall rules won't cut it anymore)
- Your worker nodes will be forcefully rolled outside your maintenance time window (unless we can prevent that from happening, which would be preferred, of course)
cc @zanetworker @DockToFuture @prashanth26 @hardikdr @rfranzke

zanetworker:
> Do you have more details @zanetworker? Is the plan roughly [gardener/machine-controller-manager#339 (comment)](https://github.com/gardener/machine-controller-manager/issues/339#issuecomment-540979928) and will we basically "migrate" from external IPs to NAT GWs? If so and because of @rfranzke comment https://github.com/gardener/gardener-extensions/pull/379?

Yes, machines will be migrated from externalIPs in the next reconciliation, because the machine class will change which will eventually leads to having newer machines. 

> * Your cluster will be migrated from external IPs to NAT GWs and any code/processes/solutions that depend on external IPs will fail (e.g. opening firewall rules won't cut it anymore)

That I already mentioned in the current release note but I can add that last bit `(e.g. opening firewall rules won't cut it anymore)` :) 

> * Your worker nodes will be forcefully rolled outside your maintenance time window (unless we can prevent that from happening, which would be preferred, of course)

I don't think this is needed, simply because the external IPs currently assigned for the VMs are ephemeral by nature meaning they change anyways every VM restart and can not be guranteed. For this PR, a new machine would simple be like a VM restart since the IP changes anyways, except this time there will be no ephemeral IPs.

The currently assigned IPs are ephemeral and only meant by design to allow external access for VMs. Relying on an ephemeral IP means having downtime for the application for every VM restart / hibernation / interface attachement and so on. Therefore, I don't think there is any production workload actively using it, or what do you think?

rfranzke:
> Your worker nodes will be forcefully rolled outside your maintenance time window (unless we can prevent that from happening, which would be preferred, of course)

The machines won't be rolled outside of the maintenance time window. This does only happen in case the bootstrap config changes, but here this is not the case.

controllers/provider-gcp/charts/internal/machineclass/templates/machineclass.yaml
```diff
@@ -19,6 +19,7 @@ metadata:
   namespace: {{ $.Release.Namespace }}
 spec:
   canIpForward: {{ $machineClass.canIpForward }}
+  disableExternalIP: {{ $machineClass.attachExternalIP }}
```
rfranzke:
```suggestion
  disableExternalIP: {{ $machineClass. disableExternalIP }}
```

PR #383, opened 2019-10-23
Azure: Avoid parallel LoadBalancer operations

dkistner:
Set concurrent-service-syncs to 1 for Azure Cloud Controller Managers to avoid parallel operations on the Shoot LoadBalancer on Azure which cause inconsistencies in the LoadBalancer configuration.
This is a workaround (probably also the fix) for https://github.com/kubernetes/kubernetes/issues/84052

**Release note**:

```improvement operator
An issue with parallel LoadBalancer updates on Azure which caused inconsistencies in the LoadBalancer configuration is now fixed by disabling concurrent LoadBalancer operation executed by the Azure Cloud Controller Manager. Latest k8s version can be consumed again.
```


rfranzke:
Thanks, can you please add a release note that a bug was fixed that will allow consuming latest Kubernetes patch releases?

PR #400, opened 2019-10-31
JeOS: ensure required packages are installed

### LINKED ISSUE #399: Extend SUSE JeOS support
Currently if you download the official [OpenSUSE JeOS image](https://download.opensuse.org/distribution/openSUSE-stable/jeos/openSUSE-Leap-15.1-JeOS.x86_64-OpenStack-Cloud.qcow2), you won't find the required packages installed there:
* docker
* socat
* jq
* nfs-client

[Here is the list](https://gist.github.com/kayrus/357f8dfbc3a8f737dee2fadd82c2323e) of packages installed in public JeOS.

From my perspective it's worth to add the code, which ensures that the above packages are installed.

kayrus:

The PR adds the code, which ensures that required packages are installed in SUSE OS.

Fixes #399

I put the package installation into the loop, because zypper lock file can be in place during the boot. Default cloud-init `packages:` can fail because of this and packages can be missing during the boot.

7 return code means that the zypper is locked: [ZYPPER_EXIT_ZYPP_LOCKED](https://en.opensuse.org/SDB:Zypper_manual_(plain))

Even if the `zypper install` fails, the further commands in the `runcmd:` list will be executed. If the packages are already installed and there are no repos configured, zypper will just return 0.

**Release note**:

```improvement operator
Ensure required packages are installed in JeOS
```

/cc @vpnachev

### DIFF
--- a/controllers/os-suse-jeos/pkg/generator/templates/cloud-init-suse-jeos.template
+++ b/controllers/os-suse-jeos/pkg/generator/templates/cloud-init-suse-jeos.template
@@ -26,6 +26,7 @@ write_files:
 {{- end -}}
 {{- end }}
 runcmd:
+- "until zypper -q install -y docker wget socat jq nfs-client; [ $? -ne 7 ]; do sleep 1; done"
 - systemctl daemon-reload
 {{ if .Bootstrap -}}
 - ln -s /usr/bin/docker /bin/docker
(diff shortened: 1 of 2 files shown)


vpnachev:
Thank for the PR @kayrus 
I will validate it and come back after that.

PR #401, opened 2019-10-31
Azure: Enable os disk configuration

dkistner:
Currently the Azure provider extension does not pass the configured os volume type to the `AzureMachineClass` and the machine-controller-manager will in consequence create machines always with the default os disk belonging to the requested machine type.

This will now change. The Azure provider will write the volume type now to the `AzureMachineClass`, but only if the passed volume type is a valid Azure volume type. Otherwise the default volume type of the machine type will be used.

**Release note**:

```improvement user
The os volume type configuration for worker machines of Azure Shoot cluster will be now passed correctly, but only if the selected volume type is a valid Azure volume type.

This will trigger a rolling update of all machines in the cluster during the maintenance window, if configured.
```

### DIFF
--- a/controllers/provider-azure/pkg/controller/worker/machines.go
+++ b/controllers/provider-azure/pkg/controller/worker/machines.go
@@ -135,6 +135,22 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
+		osDisk := map[string]interface{}{
+			"size": volumeSize,
+		}
+
+		// In the past the volume type information was not passed to the machineclass.
+		// In consequence the Machine controller manager has created machines always
+		// with the default volume type of the requested machine type. Existing clusters
+		// respectively their worker pools could have an invalid volume configuration
+		// which was not applied. To do not damage exisiting cluster we will set for
+		// now the volume type only if it's a valid Azure volume type.
+		// Otherwise we will still use the default volume of the machine type.
+		if pool.Volume.Type != nil {
+			if *pool.Volume.Type == "Standard_LRS" || *pool.Volume.Type == "StandardSSD_LRS" || *pool.Volume.Type == "Premium_LRS" {
+				osDisk["type"] = *pool.Volume.Type
+			}
+		}
 
 		image := map[string]interface{}{
 			"publisher": publisher,
@@ -173,7 +189,7 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 					},
 					"machineType":  pool.MachineType,
 					"image":        image,
-					"volumeSize":   volumeSize,
+					"osDisk":       osDisk,
 					"sshPublicKey": string(w.worker.Spec.SSHPublicKey),
 				}
 			)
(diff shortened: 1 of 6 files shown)


dkistner:
Do not merge until review by @AndreasBurger has been completed

controllers/provider-azure/pkg/controller/worker/machines.go
```diff
@@ -136,6 +136,18 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
...
 		}
+		osDisk := map[string]interface{}{
+			"size": volumeSize,
+		}
+
+		// If no volume type is passed or the passed volume type is not a valid Azure volume type
+		// then nothing will be written into the machineclass and the default volume type
+		// of the belonging machine type will be used.
+		if pool.Volume.Type != nil {
+			if *pool.Volume.Type == "Standard_LRS" || *pool.Volume.Type == "StandardSSD_LRS" || *pool.Volume.Type == "Premium_LRS" {
```
vpnachev: How about to remove this validation and let the gardener landscape operator to define the supported volume types in the `CloudProfile`? In this way the `provider-azure` extension will not need to be rebuild and released if a new disk type is added or current one is removed by Azure. Just get an example how `MachineType` is implemented.

ialidzhikov: Yep, sound reasonable.

rfranzke: +1, let's not hard-code the types here. Your documentation clearly states that only these types are valid, so using any other type is a configuration mistake and will probably prevent the machine from booting, right?

controllers/provider-azure/pkg/controller/worker/machines.go
```diff
@@ -174,7 +186,7 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 					},
 					"machineType":  pool.MachineType,
 					"image":        image,
-					"volumeSize":   volumeSize,
+					"osDisk":       osDisk,
```
rfranzke: Doing this will trigger a rolling update of all machines of all clusters (in their maintenance windows, if configured). Can you please add this to the release notes?

controllers/provider-azure/pkg/controller/worker/machines.go
dkistner: Yes, I'm completely with you and I don't like it too. The reason for adding this check was to not damage existing clusters.

Old clusters were created with the default volume type of the machineType, because we don't passed the volume type information to MCM. This will change now and if an user has already configured worker pools with an invalid volume type then this invalid one will be now passed to MCM in the next reconconcilation and will the new machine can't be created. 

New machines and worker pools will be prevented from creation by validation of course, when they use an invalid machine type.

rfranzke: OK, makes sense, thank you for the explanation. I think it makes sense to put this information as inline comment in the code itself as well.

dkistner: Done. I have extended the inline comment.

controllers/provider-azure/pkg/controller/worker/machines.go
dkistner: Done.

controllers/provider-azure/pkg/controller/worker/machines.go
vpnachev: Thanks for the details, @dkistner !

PR #417, opened 2019-11-06
Enable Quota friendly Cloud Router sharing for CloudNATs

### LINKED ISSUE #411: Implement Quota friendly Cloud Router sharing for CloudNATs.
**Problem**
Currently it is possible to utilize CloudNAT from GCP to allow external connectivity from VMs to the outside world without exposing their public IP. The problem with the current approach is that there is a hard quota limit of 5 Cloud Routers / VPC, this means that only 5 shoots can share the same VPC (https://cloud.google.com/router/quotas). 

**Possible solution**
To circumvent this issue, we can have two scenarios: 
- If a VPC is not created / does not exists, we simply create the Cloud Router for the VPC, if newer shoots would like to share the VPC, they can provide the created Cloud Router name during the shoot creation. 
- If the VPC already exists, it will be required to provide the CloudRouter name along with the VPC, in that case the CloudNAT for the new shoot will share the CloudRouter specified in the name field. 
- if a VPC already exists and a CloudRouter name is not provided, then VMs are created the old way (with a public IP, until a deadline which dictates the use of a CloudRouter). 
   - In two months time, this adapting code will be removed. 

**Changes required**
- Expose the CloudRouter Name in the API.
- Differentiate between the two cases in code (i.e., decide whether a CloudRouter needs to be created or re-used). 

part of #408

/cc @DockToFuture @rfranzke @vlerenc

zanetworker:
Co-authored-by: DockToFuture  <sebastian.stauch@sap.com>

Enable CloudNAT while allowing CloudRouters in the same VPC to be shared. 

Fixes #411 

**Release note**:

```action user
If a VPC is to be re-used for shoot clusters, a (manually pre-provisioned) GCP Cloud Router name needs to be specified as well, this is to avoid hitting quota limits (5 Cloud routers / VPC) and allowing all the Cloud NATs to share that Cloud Router. If no Cloud Router name is specified, all the nodes will be assigned a public IP.
```

```action user
After two months, no public IPs will be allowed and specifying a Cloud Router name will be enforced. Please adapt your shoot deployments into a existing VPCs as soon as possible.
```


rfranzke:
Thanks for the PR! What about the restriction you shared yesterday w.r.t. the quota of 1 cloud nat per network per region?

zanetworker:
> Thanks for the PR! What about the restriction you shared yesterday w.r.t. the quota of 1 cloud nat per network per region?

It is necessary to specify the source subnet for the NAT gateway, otherwise, the first NAT to be created in the VPC will cover all Subnet CIDRs in the same region and it won't be allowed to create another nat (for another shoot).

DockToFuture:
All test cases were tested for this PR:
- [x] no vpc, no cr -> both were created, nat entry was created, nodes have no external ip, internet can be reached
- [x] vpc, cr -> nat entry was created, nodes have no external ip, internet can be reached
- [x] vpc, no cr -> nodes have external ip,  internet can be reached

ialidzhikov:
```
Executing check-generate
go generate needs to be run:
 M controllers/provider-gcp/hack/api-reference/api.md
make: *** [Makefile:43: check] Error 1
```

controllers/provider-gcp/charts/internal/gcp-infra/templates/main.tf
```diff
@@ -31,6 +31,52 @@ resource "google_compute_subnetwork" "subnetwork-nodes" {
...
+  }
+
+  log_config {
+    enable = true
+    filter = "ERRORS_ONLY"
+  }
+}
+{{- end }}
+
+{{ if and (ne .Values.vpc.cloudRouter.name "none") (ne .Values.vpc.cloudRouter.name "")}}
```
rfranzke: Don't check for name `none`. Rather use the standard pattern `{{ if .Values.vpc.cloudRouter.name ... }}`

controllers/provider-gcp/charts/internal/gcp-infra/values.yaml
```diff
@@ -4,9 +4,12 @@ google:
 
 create:
   vpc: true
+  cloudRouter: false
 
 vpc:
   name: ${google_compute_network.network.name}
+  cloudRouter:
```
rfranzke:
```suggestion
# cloudRouter:
```

controllers/provider-gcp/charts/internal/gcp-infra/values.yaml
```diff
@@ -4,9 +4,12 @@ google:
...
+    name: "none"
```
rfranzke:
```suggestion
#   name: "none"
```

controllers/provider-gcp/charts/internal/gcp-infra/templates/main.tf
```diff
@@ -31,6 +31,52 @@ resource "google_compute_subnetwork" "subnetwork-nodes" {
...
+resource "google_compute_router_nat" "nat" {
```
rfranzke: Why is this resource duplicated? It already exists in line 41. Can we prevent the duplication?

controllers/provider-gcp/charts/internal/gcp-infra/templates/main.tf
```diff
@@ -121,6 +167,29 @@ output "{{ .Values.outputKeys.vpcName }}" {
...
+  value = "${google_compute_router.router.name}"
+}
+output "{{ .Values.outputKeys.cloudNAT }}" {
+  value = "${google_compute_router_nat.nat.name}"
+}
+{{ else if and (ne .Values.vpc.cloudRouter.name "none") (ne .Values.vpc.cloudRouter.name "")}}
+output "{{ .Values.outputKeys.cloudRouter }}" {
+  value = "{{.Values.vpc.cloudRouter.name}}"
+}
+output "{{ .Values.outputKeys.cloudNAT }}" {
```
rfranzke: same here - DRY

controllers/provider-gcp/charts/internal/gcp-infra/templates/main.tf
```diff
@@ -121,6 +167,29 @@ output "{{ .Values.outputKeys.vpcName }}" {
...
+  value = "${google_compute_router_nat.nat.name}"
+}
+{{ else }}
+output "{{ .Values.outputKeys.cloudRouter }}" {
+  value = "{{.Values.vpc.cloudRouter.name}}"
+}
+output "{{ .Values.outputKeys.cloudNAT }}" {
```
rfranzke: and here

controllers/provider-gcp/pkg/controller/worker/machines.go
```diff
@@ -132,6 +132,10 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 		}
 
 		for zoneIndex, zone := range pool.Zones {
+			var disableExternalIP = false
+			if infrastructureStatus.Networks.VPC.CloudRouter != nil && infrastructureStatus.Networks.VPC.CloudRouter.Name != "none" {
```
rfranzke: again - don't check for "none". The first condition should be good enough.

controllers/provider-gcp/pkg/controller/worker/machines_test.go
```diff
@@ -283,7 +283,8 @@ var _ = Describe("Machines", func() {
 						"machineType": machineType,
 						"networkInterfaces": []map[string]interface{}{
 							{
-								"subnetwork": subnetName,
+								"subnetwork":        subnetName,
+								"disableExternalIP": false,
```
rfranzke: Can you add a separate test case where `disableExternalIP` will be set to `true`?

controllers/provider-gcp/pkg/internal/infrastructure/terraform.go
```diff
@@ -61,25 +65,38 @@ func ComputeTerraformerChartValues(
...
+		cloudRouterName   string
 	)
 
 	if config.Networks.VPC != nil {
 		createVPC = false
+		createCloudRouter = false
+
 		vpcName = config.Networks.VPC.Name
+
+		cloudRouterName = "none"
```
rfranzke: remove

controllers/provider-gcp/pkg/internal/infrastructure/terraform.go
```diff
@@ -61,25 +65,38 @@ func ComputeTerraformerChartValues(
...
 		"google": map[string]interface{}{
 			"region":  infra.Spec.Region,
 			"project": account.ProjectID,
 		},
-		"create": map[string]interface{}{
-			"vpc": createVPC,
-		},
 		"vpc": map[string]interface{}{
 			"name": vpcName,
+			"cloudRouter": map[string]interface{}{
```
rfranzke: only provide this key to the terraform helm chart if a cloud router name was given

controllers/provider-gcp/pkg/internal/infrastructure/terraform.go
```diff
@@ -177,6 +204,9 @@ func StatusFromTerraformState(state *TerraformState) *gcpv1alpha1.Infrastructure
 			Networks: gcpv1alpha1.NetworkStatus{
 				VPC: gcpv1alpha1.VPC{
 					Name: state.VPCName,
+					CloudRouter: &gcpv1alpha1.CloudRouter{
```
rfranzke: Only write the cloud router name into the status if it was created or provided by the end-user

controllers/provider-gcp/charts/internal/gcp-infra/templates/main.tf
zanetworker: its not duplicated, this is a different case.

controllers/provider-gcp/charts/internal/gcp-infra/templates/main.tf
zanetworker: its not duplicated, this is a different case., whats DRY?

controllers/provider-gcp/charts/internal/gcp-infra/templates/main.tf
zanetworker: as we discussed, none would be better than empty.

controllers/provider-gcp/charts/internal/gcp-infra/templates/main.tf
rfranzke: The only thing that is different in this case is one value. I recommend to not duplicate it, and instead, to only change this one value depending on the case you want to handle. Will make it easier to maintain/read.

controllers/provider-gcp/pkg/internal/infrastructure/terraform.go
zanetworker: as discussed, none would be better discriptive to write in the terraform output, so it will be set to none if not given.

controllers/provider-gcp/charts/internal/gcp-infra/templates/main.tf
zanetworker: as we discussed, none would be better than empty.

controllers/provider-gcp/pkg/internal/infrastructure/terraform.go
rfranzke: I don't agree. What if a user has named its cloud router "none"? It's always better to set the pointer object to `nil` if the value is optional, see https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required.

zanetworker: I am aware of the conventions in this case :). But this was decided because of the terrafom output parsing when we discussed earlier. In anycase, if you still think having an empty string in the terraform output is okay then we can replace none.

rfranzke: Why do you want to have an output value at all in case no cloud router was created? I would simply omit the output value (<=> no router was created).

controllers/provider-gcp/charts/internal/gcp-infra/templates/main.tf
```diff
@@ -31,6 +31,39 @@ resource "google_compute_subnetwork" "subnetwork-nodes" {
...
+  region  = "{{ required "google.region is required" .Values.google.region }}"
+  network = "{{ required "vpc.name is required" .Values.vpc.name }}"
+}
+{{- end }}
+
+{{ if or  .Values.create.cloudRouter .Values.vpc.cloudRouter -}}
+resource "google_compute_router_nat" "nat" {
+  name                               = "{{ required "clusterName is required" .Values.clusterName }}-cloud-nat"
+  {{  if .Values.vpc.cloudRouter -}}
+  router                             = "{{ .Values.vpc.cloudRouter.name }}"
```
rfranzke:
```suggestion
  router                             = "{{ required "vpc.cloudRouter.name is required" .Values.vpc.cloudRouter.name }}"
```

controllers/provider-gcp/charts/internal/gcp-infra/templates/main.tf
```diff
@@ -121,6 +154,24 @@ output "{{ .Values.outputKeys.vpcName }}" {
   value = "{{ required "vpc.name is required" .Values.vpc.name }}"
 }
 
+{{ if .Values.create.cloudRouter -}}
+output "{{ .Values.outputKeys.cloudRouter }}" {
+  value = "${google_compute_router.router.name}"
+}
+{{- end }}
+
+{{ if  .Values.vpc.cloudRouter -}}
```
rfranzke: WDYT about the following:

```terraform
{{ if or  .Values.create.cloudRouter .Values.vpc.cloudRouter -}}
output "{{ .Values.outputKeys.cloudRouter }}" {
  {{ if .Values.create.cloudRouter -}}
  value = "${google_compute_router.router.name}"
  {{ else -}}
  value = "{{ .Values.vpc.cloudRouter.name }}"
  {{ end -}}
}

output "{{ .Values.outputKeys.cloudNAT }}" {
  value = "${google_compute_router_nat.nat.name}"
}
{{- end }}
```

This makes more clear that a cloud router + NAT gateway is only available as output if either a new cloud router is created or a name of an existing cloud router is given.

controllers/provider-gcp/docs/usage-as-end-user.md
```diff
@@ -43,6 +45,9 @@ The `networks.vpc` section describes whether you want to create the shoot cluste
 * If `networks.vpc.name` is given then you have to specify the VPC name of the existing VPC that was created by other means (manually, other tooling, ...).
 If you want to get a fresh VPC for the shoot then just omit the `networks.vpc` field.
 
+* If a VPC is to be re-used, and the `networkers.vpc.cloudRouter.Name` is given then you have to specify the name of a CloudRouter that was created by other means (manually, other tooling, ...),
```
rfranzke:
```suggestion
* If a VPC is to be re-used, and the `networkers.vpc.cloudRouter.name` is given then you have to specify the name of a CloudRouter that was created by other means (manually, other tooling, ...), 
```

rfranzke: I would not understand this as an end-user, I think.
Can we rephrase such that the following gets clear:

* If a VPC name is not given then we will create the cloud router + NAT gateway to ensure that worker nodes don't get external IPs.
* If a VPC name is given then
  * if a cloud router name is not given we won't do anything, i.e. not creating a cloud router + NAT ourselves, resulting in worker nodes having external IPs (only for limited amount of time, end-users are asked to migrate to the following case)
  * if a cloud router name is given we re-use this one and create a dedicated NAT for the shoot, allowing worker nodes to not have external IPs.

controllers/provider-gcp/pkg/apis/gcp/validation/infrastructure.go
```diff
@@ -68,6 +68,10 @@ func ValidateInfrastructureConfig(infra *apisgcp.InfrastructureConfig, nodesCIDR
 		allErrs = append(allErrs, field.Invalid(networksPath.Child("vpc", "name"), infra.Networks.VPC.Name, "vpc name must not be empty when vpc key is provided"))
 	}
 
+	if infra.Networks.VPC != nil && len(infra.Networks.VPC.Name) == 0 && infra.Networks.VPC.CloudRouter != nil {
+		allErrs = append(allErrs, field.Invalid(networksPath.Child("vpc", "cloudRouter"), infra.Networks.VPC.Name, "cloud router can not be configured when the VPC name is not specified"))
```
rfranzke:
```suggestion
		allErrs = append(allErrs, field.Invalid(networksPath.Child("vpc", "cloudRouter"), infra.Networks.VPC.CloudRouter, "cloud router can not be configured when the VPC name is not specified"))
```

controllers/provider-gcp/pkg/apis/gcp/validation/infrastructure_test.go
```diff
@@ -126,6 +126,23 @@ var _ = Describe("InfrastructureConfig validation", func() {
 					"Detail": Equal("must be valid canonical CIDR"),
 				}))
 			})
+			It("should forbid configuring CloudRouter if VPC name is not set", func() {
+				infrastructureConfig.Networks.VPC = &apisgcp.VPC{}
+				infrastructureConfig.Networks.VPC.CloudRouter = &apisgcp.CloudRouter{}
+
+				errorList := ValidateInfrastructureConfig(infrastructureConfig, &nodes, &pods, &services)
+
+				Expect(errorList).To(HaveLen(2))
```
rfranzke: Not needed, or?

controllers/provider-gcp/pkg/controller/worker/machines_test.go
```diff
@@ -176,7 +176,7 @@ var _ = Describe("Machines", func() {
 					},
 				}
 
-				w = &extensionsv1alpha1.Worker{
+				wrker = &extensionsv1alpha1.Worker{
```
rfranzke: Was the renaming, even with a typo, on purpose?

zanetworker: yes, having a one letter variable is quite annoying in tracing, worker is a package, so wrker was the next inline.

rfranzke:
/lgtm, thanks!
@zanetworker you mentioned that @DockToFuture wanted to perform some tests after your changes - is this already done? Can we merge this?

zanetworker:
> /lgtm, thanks!
> @zanetworker you mentioned that @DockToFuture wanted to perform some tests after your changes - is this already done? Can we merge this?

yes he already performed the tests, good to merge.

PR #418, opened 2019-11-06
Move terraformer library from gardener to gardener-extensions

tim-ebert:
This PR moves the terraformer library package from `gardener/gardener/pkg/operation/terraformer` to `gardener/gardener-extensions/pkg/terraformer`.
Gardener itself does no longer use it but only the extensions, hence, its correct home is in the gardener-extensions library.
It also merges the terraformer wrapper package (`gardener/gardener-extensions/pkg/gardener/terraformer`) with the moved package.
All extension controllers were adapted to only use the newly exposed `gardener/gardener-extensions/pkg/terraformer.Terraformer` interface.

See also gardener/gardener#1596

**Release note**:

```noteworthy developer
The `terraformer` package was moved from `gardener/gardener` to `gardener/gardener-extensions/pkg/terraformer`.
```


ialidzhikov:
There are some `golangci-lint` failures. Could you please check them?

tim-ebert:
Lint failures are fixed, PTAL. 👍

pkg/terraformer/types.go
```diff
@@ -82,3 +83,30 @@ const (
 	// TerraformerPurposeBackup is a constant for the complete Terraform setup with purpose 'etcd backup'.
 	TerraformerPurposeBackup = "backup"
```
ialidzhikov:
```suggestion

```

pkg/terraformer/types.go
```diff
@@ -15,14 +15,15 @@
 package terraformer
 
 import (
+	"k8s.io/client-go/rest"
```
ialidzhikov: You could move it below.

pkg/terraformer/types.go
```diff
@@ -82,3 +83,30 @@ const (
...
+	SetDeadlinePod(time.Duration) Terraformer
+	SetDeadlineJob(time.Duration) Terraformer
+	InitializeWith(initializer Initializer) Terraformer
+	Apply() error
+	Destroy() error
+	GetStateOutputVariables(variables ...string) (map[string]string, error)
+	ConfigExists() (bool, error)
+}
+
+// Initializer can initialize an Terraformer.
```
ialidzhikov:
```suggestion
// Initializer can initialize a Terraformer.
```

pkg/terraformer/types.go
tim-ebert: const removed 👍

tim-ebert:
Thanks for your reviews!
Suggestions applied and rebased onto master, PTAL.

PR #421, opened 2019-11-07
Remove deprecated fields

vpnachev:
As follow up on #290, remove deprecated fields before g/g-ext `1.0.0` is released.

Fixes #

There is a similar PR in the MCM https://github.com/gardener/machine-controller-manager/pull/353 

**Release note**:

```action operator
Since version `0.13.0` the fields `publisher`, `offer`, `sku` and `version` in the `AzureMachineClass` are deprecated and now support for them has been removed. Please switch your Azure provider extension to use URN as OS image reference, which is the replacement of the deprecated fields.
```


vpnachev:
Put it on `WIP`, cause of failing tests.

vpnachev:
The validations are fixed.
/needs review

PR #297, opened 2019-08-27
Generate bootstrap script from OSC for JeOS

pablochacin:
**What this PR does / why we need it**:
Add a template for generating bash scripts from OSC to JeOS OSC controller, and the ability to select either `cloud-init` or `script` when deploying the controller. 

The format of the bootstrap file is selected via the `OS_CONFIG_FORMAT` environment variable in the controller's deployment spec. The value of this variable is set in the `osConfig.Format` value in the controller deployment chart. 

The command to executed at boot time to apply the OSC may depend on the cloud provider, therefore the logic for setting it maybe should be placed elsewhere.

In this WIP the boot command is set in the `BOOT_COMMAND' environment variable in the controller deployment spec. The value of this value is set in the osConfig.bootCmd value in the controller deployment chart.  

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
1. This is an WIP submitted to gather feedback about:
- General approach (allow selecting format vs hardcoded template)
- Mechanism for selecting bootstrap format (cloud-init or script) using environment variable to directly name the format

2. The template generator has been tested locally, but no testing on live clusters has been tried.
A cluster deployed on a cloud provider which supports script based bootstrapping is required for testing.

**Release note**:

```improvement operator
Allows deployment of JeOS-based worker nodes on cloud providers on which cloud-init is not supported. By default bash `script` bootstrapping is used. 
```

```action operator
If you want to keep using `cloud-init` for SuSE-JeOS nodes bootstrapping, use `osConfig.format` and `osConfig.bootCmd` values to switch to cloud-init. Sample values can be found [here](https://github.com/gardener/gardener-extensions/blob/master/controllers/os-suse-jeos/charts/os-suse-jeos/values.yaml).
```

PR #423, opened 2019-11-08
Use defaults for env vars

vpnachev:
With the recently merged #297 env vars were introduced to allow configuration of the SuSE-JeOS controller. It turns out that these env vars are indirectly required by all extensions when the hyper binary is used. Now if these env vars are unset, default values will be used instead to fail.

Fixes #

**Release note**:

```improvement operator
Make the env vars `OS_CONFIG_FORMAT` and `BOOT_COMMAND` optional to allow the hyper binary to run without them.
```

### DIFF
--- a/controllers/os-suse-jeos/pkg/generator/generator.go
+++ b/controllers/os-suse-jeos/pkg/generator/generator.go
@@ -15,7 +15,6 @@
 package generator
 
 import (
-	"fmt"
 	"os"
 	"strings"
 	"text/template"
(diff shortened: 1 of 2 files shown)


ialidzhikov:
I think I reproduced locally..

```
{"level":"error","ts":"2019-11-09T21:11:27.987Z","msg":"Could not create Generator","error":"Environment variable OS_CONFIG_FORMAT must be defined","stacktrace"}
```

/kind bug

rfranzke:
Why have you done it this way?  Isn't "making the variables optional" just an ugly work-around for the actual problem (which is that these OS controller related functions are called in all cases (even if no OS controller command is triggered))?

Isn't the correct solution this one:

1. Instead of expecting a generator [here](https://github.com/gardener/gardener-extensions/blob/master/pkg/controller/operatingsystemconfig/oscommon/app/app.go#L34:6), this function should accept a function that creates a generator.
1. This function should be called at an appropriate place in the [`Run` section of the cobra command](https://github.com/gardener/gardener-extensions/blob/master/pkg/controller/operatingsystemconfig/oscommon/app/app.go#L62).

This way, the environment variables can still be mandatory and the related code is only executed if the `Run` function is called.

PR #426, opened 2019-11-11
Switch from job to pod for terraform apply/destroy

### LINKED ISSUE #182: Duplicate resources prevent the creation of a cluster
**What happened**:

We are using gardener with an istio extension. During our [integration tests](https://concourse.istio.sapcloud.io/teams/garden/pipelines/gardener/jobs/integration-test), we create and delete a shoot cluster. We are always using the same cluster name. After a few loops we encountered the following error in the logs of a `<shoot>.infra.tf-job` 

```
Error: Error applying plan:

3 errors occurred:
        * aws_iam_role.nodes: 1 error occurred:
        * aws_iam_role.nodes: Error creating IAM Role shoot--core--integration-test-nodes: EntityAlreadyExists: Role with name shoot--core--integration-test-nodes already exists.
        status code: 409, request id: cd4798f1-9e59-11e9-9a03-27542a2f5747


        * aws_key_pair.kubernetes: 1 error occurred:
        * aws_key_pair.kubernetes: Error import KeyPair: InvalidKeyPair.Duplicate: The keypair 'shoot--core--integration-test-ssh-publickey' already exists.
        status code: 400, request id: d1b878e3-c75a-4fa6-82ab-8a50d11a2227


        * aws_iam_role.bastions: 1 error occurred:
        * aws_iam_role.bastions: Error creating IAM Role shoot--core--integration-test-bastions: EntityAlreadyExists: Role with name shoot--core--integration-test-bastions already exists.
        status code: 409, request id: cd4749f9-9e59-11e9-82f6-8f5e7da59408

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```

Duplicates should be either ignored or deleted.

Not sure how to reproduce this.

**Anything else we need to know?**:

**Environment**:

- Gardener version: master
- Kubernetes version (use `kubectl version`): 1.14.0
- Cloud provider or hardware configuration: AWS
- Others:

### LINKED ISSUE #202: Duplicate KeyPair/Roles on AWS
**What happened**:
Shoot gets stuck during creation with the error message that a key pair and two roles (one with "nodes" and one with "bastion") belonging to the shoot are already present on the infrastructure and thus can't be created.
(Currently trying to reproduce the problem, will add exact error message afterwards)

The shoot is created.

Not completely sure, but it seems that this happens when the seed needs to scale up during shoot creation.

**Anything else we need to know?**:
Has been observed for AWS shoots only. Seems to happen when the shoot is created, but the seed is full and during shoot creation, the auto-scaling of the seed cluster kicks in and scales up.
The error can be resolved by manually removing the duplicate key pair and roles from the infrastructure.

**Environment**:
- Gardener version: 0.26.2
- Cloud provider or hardware configuration: AWS

### LINKED ISSUE #428: Remove terraformer validation Pod
Currently we do deploy a terraformer Pod which executes the validate command in terraformer (terraform validate and terraform plan):
https://github.com/gardener/terraformer/blob/993cebd92e69205adc978489f75005dd5dd170b4/terraform.sh#L122-L138

@rfranzke suggested that we probably can remove the terraformer validate and do terraform apply directly.

ialidzhikov:
Currently we use Job for terraform apply/destroy operations. The Job controller starts a  new Pod if the first Pod fails or is deleted. When the first `tf-job` Pod is deleted, it is being gracefully terminated for given period of `activeDeadlineSeconds` (630 in our case). However the Job controller start a new `tf-job` Pod which at the end of its execution overrides the correct terraform state of the first `tf-job` Pod.
This PR adapts the terraformer package (`pkg/terraformer`) to use Pod to prevent multiple Pods executing simultaneously terraform apply/destroy.

Fixes #182
Fixes #202
Fixes #428

**Release note**:

```improvement operator
`pkg/terraformer` package does now use a Pod for `terraform apply/destroy` operations. The motivation for this changes is to prevent multiple Pods to execute `apply/destroy` commands simultaneously in some cases.
```
```improvement operator
`pkg/terraformer` package does no longer deploy a Terraformer validation pod. It executes directly `terraform apply/destroy`.
```
```action developer
`pkg/terraformer` package has renamed and deleted funcs. If you are vendoring this package in your project, make sure to adapt to these changes when adopting the new version of `gardener/gardener-extensions`.
```

### DIFF
--- a/pkg/terraformer/config.go
+++ b/pkg/terraformer/config.go
@@ -22,7 +22,6 @@ import (
 	"time"
 
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
-
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -46,13 +45,7 @@ func (t *terraformer) SetVariablesEnvironment(tfvarsEnvironment map[string]strin
 	return t
 }
 
-// SetJobBackoffLimit configures the backoff limit for the Terraformer job.
-func (t *terraformer) SetJobBackoffLimit(limit int32) Terraformer {
-	t.jobBackoffLimit = limit
-	return t
-}
-
-// SetActiveDeadlineSeconds configures the active deadline seconds for the Terraformer pod and job.
+// SetActiveDeadlineSeconds configures the active deadline seconds for the Terraformer pod.
 func (t *terraformer) SetActiveDeadlineSeconds(adl int64) Terraformer {
 	t.activeDeadlineSeconds = adl
 	return t
@@ -64,15 +57,9 @@ func (t *terraformer) SetDeadlineCleaning(d time.Duration) Terraformer {
 	return t
 }
 
-// SetDeadlinePod configures the deadline while waiting for the Terraformer pod.
+// SetDeadlinePod configures the deadline while waiting for the Terraformer apply/destroy pod.
 func (t *terraformer) SetDeadlinePod(d time.Duration) Terraformer {
 	t.deadlinePod = d
-	return t
-}
-
-// SetDeadlineJob configures the deadline while waiting for the Terraformer job.
-func (t *terraformer) SetDeadlineJob(d time.Duration) Terraformer {
-	t.deadlineJob = d
 	return t
 }
 
@@ -202,8 +189,14 @@ func (t *terraformer) prepare(ctx context.Context) (int, error) {
 		return -1, errors.New("no Terraform variables environment provided")
 	}
 
+	// Clean up possible existing pod artifacts from previous runs
+	if err := t.ensureCleanedUp(ctx); err != nil {
+		return -1, err
+	}
+
 	// Clean up possible existing job/pod artifacts from previous runs
-	if err := t.ensureCleanedUp(); err != nil {
+	// TODO: Remove after several releases.
+	if err := t.ensureCleanedUpDeprecated(); err != nil {
 		return -1, err
 	}
 
@@ -261,9 +254,26 @@ func (t *terraformer) cleanupConfiguration(ctx context.Context) error {
 	return nil
 }
 
-// ensureCleanedUp deletes the job, pods, and waits until everything has been cleaned up.
-func (t *terraformer) ensureCleanedUp() error {
+// ensureCleanedUp deletes the Terraformer pods, and waits until everything has been cleaned up.
+func (t *terraformer) ensureCleanedUp(ctx context.Context) error {
+	podList, err := t.listTerraformerPods(ctx)
+	if err != nil {
+		return err
+	}
+	if err := t.deleteTerraformerPods(ctx, podList); err != nil {
+		return err
+	}
+
+	return t.waitForCleanEnvironment(ctx)
+}
+
+// ensureCleanedUpDeprecated deletes the job, pods, and waits until everything has been cleaned up.
+//
+// Deprecated: Terraformer does no longer uses a Job. Kept for backwards compatibility.
+// TODO: Remove after several releases.
+func (t *terraformer) ensureCleanedUpDeprecated() error {
 	ctx := context.TODO()
+
 	jobPodList, err := t.listJobPods(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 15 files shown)


rfranzke:
Thanks for the PR! What do you think about removing the validator pod now as well? It doesn't make sense to have it anymore, and by removing it we actually halve the number of calls to the cloud APIs we are doing during infrastructure reconciliation. I can't even recall why exactly we had introduced back then, but as we are now switching to a single pod execution anyways, it doesn't make sense to keep it.

ialidzhikov:
Your suggestion sounds reasonable. If you don't mind, I will open a separate issue about it as it is somehow a separate cleanup/performance improvement - #428 .
I would rather prefer to do not mix up both changes in this PR.

rfranzke:
Hm, is it really that much of additional work? It's mostly deleting code, isn't it? You have already touched the relevant spots. Though, if you consider it's too much additional work now then let's do it separately. Please check again.

rfranzke:
@ialidzhikov can you provide an update please? Are you looking into it and working on it, or do you consider it to be to much and still prefer to do it separately (i.e., allowing this PR to be reviewed now)?

ialidzhikov:
I am working on it.

rfranzke:
OK, thank you.

pkg/terraformer/config.go
```diff
@@ -202,11 +189,17 @@ func (t *terraformer) prepare(ctx context.Context) (int, error) {
 		return -1, errors.New("no Terraform variables environment provided")
 	}
 
-	// Clean up possible existing job/pod artifacts from previous runs
+	// Clean up possible existing pod artifacts from previous runs
 	if err := t.ensureCleanedUp(); err != nil {
 		return -1, err
 	}
 
+	// Clean up possible existing job/pod artifacts from previous runs
```
rfranzke: nice!

pkg/terraformer/terraformer.go
```diff
@@ -490,11 +465,26 @@ func (t *terraformer) retrievePodLogs(jobPodList *corev1.PodList) (map[string]st
...
 		return result, nil
 	case <-time.After(2 * time.Minute):
-		return nil, fmt.Errorf("Timeout when reading the logs of all pods created by Terraform job '%s'", t.jobName)
+		return nil, fmt.Errorf("timeout when reading the logs of all pods created by Terraformer '%s'", t.name)
+	}
+}
+
+func (t *terraformer) deleteTerraformerPods(ctx context.Context, podList *corev1.PodList) error {
+	for _, pod := range podList.Items {
+		if err := t.client.Delete(ctx, &pod); err != nil && !apierrors.IsNotFound(err) {
```
rfranzke:
```suggestion
		if err := t.client.Delete(ctx, &pod); client.IgnoreNotFound(err) != nil {
```

pkg/terraformer/terraformer.go
```diff
@@ -517,3 +507,8 @@ func (t *terraformer) cleanupJob(ctx context.Context, jobPodList *corev1.PodList
 	}
 	return nil
 }
+
+func (t *terraformer) computePodName(command string) string {
+	suffix := utils.ComputeSHA256Hex([]byte(time.Now().String()))[:5]
```
rfranzke: Can we remove this on our side? Instead, can we simply use `.metadata.generateName`? We will find the pods anyways via labels.

pkg/terraformer/config.go
```diff
@@ -261,21 +254,41 @@ func (t *terraformer) cleanupConfiguration(ctx context.Context) error {
 	return nil
 }
 
-// ensureCleanedUp deletes the job, pods, and waits until everything has been cleaned up.
+// ensureCleanedUp deletes the Terraformer pods, and waits until everything has been cleaned up.
 func (t *terraformer) ensureCleanedUp() error {
 	ctx := context.TODO()
```
timuthy: I think it makes sense that the `context` is passed by the caller, doesn't it?

ialidzhikov: Yes, it makes sense for sure!

pkg/terraformer/terraformer.go
ialidzhikov: Yes, switched to `.metadata.generateName`. Nice suggestion.

ialidzhikov:
The ci/cd hanged with
```
Backend error: Exit status: 404, message: {"Type":"ProcessNotFoundError","Message":"unknown process: resource","Handle":"","ProcessID":"resource","Binary":""}
```

I will force push to retrigger it.

PR #217, opened 2019-07-26
Worker delegates store machine image information in provider status

rfranzke:
**What this PR does / why we need it**:
The `WorkerDelegate` interface does now have new `GetMachineImages()` interface method. Provider extension that are using the generic worker actuator have to implement this. It should return a provider-specific `WorkerStatus` type that will get written to the `Worker`'s `.status.providerStatus` field. The worker delegate can use this information when computing the machine classes to try to find the provider-specific machine image data for the desired version. This is useful for cases where a version was removed (or updated) from the componentconfig of the provider extension, but old `Worker` resource still use this version. Without this information the worker controller would not be able to reconcile anymore.

**Special notes for your reviewer**:
As discussed/designed with @timuthy.

**Release note**:

```action developer
The `WorkerDelegate` interface does now have new `GetMachineImages()` interface method. Provider extension that are using the generic worker actuator have to implement this. It should return a provider-specific `WorkerStatus` type that will get written to the `Worker`'s `.status.providerStatus` field. The worker delegate can use this information when computing the machine classes to try to find the provider-specific machine image data for the desired version. This is useful for cases where a version was removed (or updated) from the componentconfig of the provider extension, but old `Worker` resource still use this version. Without this information the worker controller would not be able to reconcile anymore.
```

PR #427, opened 2019-11-11
Alicloud: enable customized image sharing

EmoinLanyu:
Alicloud currently does not actively support latest operating system images. To reduce & avoid security risks, we are using customized images for Shoot creation. Customized images need to be first shared with Shoot's alicloud account before it can consume it.
This PR provide automatic sharing mechanism in extension-provider-alicloud.

Fixes #

**Release note**:

```improvement user
provider-alicloud extension now automatically shares customized images to Shoot's alicloud account during infrastructure reconcile.
```


controllers/provider-alicloud/pkg/controller/worker/machines.go
```diff
@@ -40,13 +44,94 @@ func (w *workerDelegate) MachineClassList() runtime.Object {
...
 }
 
+// getSeedCloudProviderCredentials gets Seed's cloud provider's credentials from BackupBucket SecretRef
+func (w *workerDelegate) getSeedCloudProviderCredentials(ctx context.Context) (*alicloud.Credentials, error) {
+	backupBuckets := &extensionsv1alpha1.BackupBucketList{}
+	err := w.client.List(ctx, backupBuckets)
+	if err != nil {
+		return nil, err
+	}
+	credentials, err := alicloud.ReadCredentialsFromSecretRef(ctx, w.client, &backupBuckets.Items[0].Spec.SecretRef)
```
rfranzke: Well, let's not introduce such an implicit, error-prone contract. What if there is no backup configured for this seed?
Instead, if you need the Alicloud credentials, then let the operator add them explicitly to your Alicloud extension controller.

controllers/provider-alicloud/pkg/controller/worker/machines.go
```diff
@@ -40,13 +44,94 @@ func (w *workerDelegate) MachineClassList() runtime.Object {
 	return &machinev1alpha1.AlicloudMachineClassList{}
 }
 
+// getSeedCloudProviderCredentials gets Seed's cloud provider's credentials from BackupBucket SecretRef
+func (w *workerDelegate) getSeedCloudProviderCredentials(ctx context.Context) (*alicloud.Credentials, error) {
```
rfranzke: Why is all this logic here, in the worker delegate? The `Worker` control loops should only manage the worker machines. 
I consider what you are trying to achieve (image sharing) more to be related to infrastructural preparations, so please move your code to your `Infrastructure` control loop.

controllers/provider-alicloud/pkg/controller/worker/machines.go
```diff
@@ -40,13 +44,94 @@ func (w *workerDelegate) MachineClassList() runtime.Object {
...
+// ShareMachineImages checks whether Shoot's Alicloud account has permissions to use the machine images. If it can't
+// access them, these images will be shared with it from Seed's Alicloud account.
+func (w *workerDelegate) ShareMachineImages(ctx context.Context) error {
+	var (
+		shootAlicloudSTSClient      alicloudclient.STS
+		shootCloudProviderAccountID string
+		err                         error
+		logger                      logr.Logger
+	)
+	shootAlicloudSTSClient = nil
```
rfranzke: not needed, implicitly `nil` due to declaration in line 65.

controllers/provider-alicloud/pkg/controller/worker/machines.go
```diff
@@ -40,13 +44,94 @@ func (w *workerDelegate) MachineClassList() runtime.Object {
...
+	if err != nil {
+		return err
+	}
+	logger.Info("Sharing the image with Shoot's alicloud account", "worker", fmt.Sprintf("%s/%s", w.worker.Namespace, w.worker.Name))
+	for _, machineclass := range w.machineClasses {
+		region := machineclass["region"].(string)
+		imageID := machineclass["imageID"].(string)
+
+		if shootAlicloudSTSClient == nil || shootCloudProviderAccountID == "" {
+			shootAlicloudSTSClient, err = alicloudclient.NewSTSClient(ctx, region, shootCloudProviderCredentials.AccessKeyID, shootCloudProviderCredentials.AccessKeySecret)
```
rfranzke: Why is the client instantiated in every iteration of the `for` loop? The `region` is always the same and part of `.spec.region` of all extension resources. No need to iterate.

controllers/provider-alicloud/pkg/controller/worker/machines.go
```diff
@@ -40,13 +44,94 @@ func (w *workerDelegate) MachineClassList() runtime.Object {
...
+			if err != nil {
+				return err
+			}
+			shootCloudProviderAccountID, err = shootAlicloudSTSClient.GetAccountIDFromCallerIdentity(ctx)
+			if err != nil {
+				return err
+			}
+		}
+
+		seedAlicloudECSClient, err := alicloudclient.NewECSClient(ctx, region, seedCloudProviderCredentials.AccessKeyID, seedCloudProviderCredentials.AccessKeySecret)
```
rfranzke: same, move out of the `for` loop. Also, when following above suggestion, you can even instantiate the alicloud client for the seed account only once instead of in every reconciliation loop

controllers/provider-alicloud/pkg/controller/worker/machines.go
```diff
@@ -40,13 +44,94 @@ func (w *workerDelegate) MachineClassList() runtime.Object {
...
+	}
+	logger.Info("Reading Shoot's cloud provider credentials", "worker", fmt.Sprintf("%s/%s", w.worker.Namespace, w.worker.Name))
+	shootCloudProviderCredentials, err := alicloud.ReadCredentialsFromSecretRef(ctx, w.client, &w.worker.Spec.SecretRef)
+	if err != nil {
+		return err
+	}
+	logger.Info("Sharing the image with Shoot's alicloud account", "worker", fmt.Sprintf("%s/%s", w.worker.Namespace, w.worker.Name))
+	for _, machineclass := range w.machineClasses {
+		region := machineclass["region"].(string)
+		imageID := machineclass["imageID"].(string)
```
rfranzke: When moving the code to the `Infrastructure` controller, you can get the `Shoot` resource out of the `Cluster` object, and loop over all `.spec.provider.workers[]` to get the machine image IDs that you need to know in order to share them.

controllers/provider-alicloud/pkg/controller/worker/machines.go
jia-jerry: @rfranzke, so let's add secrets in https://github.wdf.sap.corp/kubernetes/landscape-setup/tree/master/components/gardener-extensions/provider-alicloud/resources.

rfranzke: Why do you post internal links to the public repository? 

> I also tried to add a Secret to https://github.com/gardener/gardener-extensions/tree/master/controllers/provider-alicloud/charts/provider-alicloud/templates, but there is no way for controller registration to config credention information. Credential information should always be stored in Secret.

Yes, but we don't have such mechanism yet, and we probably also don't want to invent one. The idea is to put the credentials as part of the Helm chart values in the `ControllerRegistration`, and advise landscape operators to not only encrypt `Secret`s but also these resources.

jia-jerry: I wiped off the internal link.

controllers/provider-alicloud/pkg/alicloud/client/client.go
```diff
@@ -151,3 +153,82 @@ func (c *storageClient) DeleteBucketIfExists(ctx context.Context, bucketName str
 func ComputeStorageEndpoint(region string) string {
 	return fmt.Sprintf("https://oss-%s.aliyuncs.com/", region)
 }
+
+type clientFactory struct {
```
EmoinLanyu: As can be seen here a new factory is added. This is because the old Factory and its FactoryFunc implementation is not good enough for multiple types of clients. Since we want to enable image sharing asap, we here first temporarily add a new factory interface, next step would be refactor client package code and merge the multiple client factories into one.

controllers/provider-alicloud/pkg/controller/worker/machines.go
EmoinLanyu: done

controllers/provider-alicloud/pkg/controller/worker/machines.go
EmoinLanyu: done

controllers/provider-alicloud/pkg/controller/worker/machines.go
EmoinLanyu: done. Previously we just wanted to share the image when actually deploying the worker. But it indeed make better sense to consider it as a part of infrastructural preparations. Thanks for the advice :)

controllers/provider-alicloud/pkg/alicloud/client/client.go
```diff
@@ -151,3 +153,76 @@ func (c *storageClient) DeleteBucketIfExists(ctx context.Context, bucketName str
...
+func NewECSClient(ctx context.Context, region, accessKeyID, accessKeySecret string) (ECS, error) {
+	client, err := ecs.NewClientWithAccessKey(region, accessKeyID, accessKeySecret)
+	if err != nil {
+		return nil, err
+	}
+
+	ecsClient := &ecsClient{
+		client: client,
+	}
+	return ecsClient, err
```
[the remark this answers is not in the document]
EmoinLanyu: done.

EmoinLanyu:
Hi @rfranzke , I have moved the code to infrastructure controller control loop. Build and tests have passed as well. Could you please help review? Thanks

controllers/provider-alicloud/pkg/controller/infrastructure/actuator.go
```diff
@@ -265,8 +283,96 @@ func computeProviderStatusVSwitches(infrastructure *alicloudv1alpha1.Infrastruct
...
 // Reconcile implements infrastructure.Actuator.
 func (a *actuator) Reconcile(ctx context.Context, infra *extensionsv1alpha1.Infrastructure, cluster *extensioncontroller.Cluster) error {
+	if a.alicloudECSClient == nil {
+		a.logger.Info("Creating Alicloud ECS client for Seed", "infrastructure", infra.Name)
+		regionID := cluster.Shoot.Spec.Region
+		seedCloudProviderCredentials, err := a.getSeedCloudProviderCredentials(ctx)
+		if err != nil {
+			return err
+		}
+		a.alicloudECSClient, err = a.newClientFactory.NewECSClient(ctx, regionID, seedCloudProviderCredentials.AccessKeyID, seedCloudProviderCredentials.AccessKeySecret)
```
rfranzke:
```suggestion
		a.alicloudECSClient, err = a.newClientFactory.NewECSClient(ctx, infra.Spec.Region, seedCloudProviderCredentials.AccessKeyID, seedCloudProviderCredentials.AccessKeySecret)
```

controllers/provider-alicloud/pkg/controller/infrastructure/types.go
```diff
@@ -20,6 +20,12 @@ import (
 )
 
 const (
+	// SeedCloudProviderSecretName is the name of the Secret of Seed's cloud provider credentials
+	SeedCloudProviderSecretName = "vm-image-owner"
```
rfranzke: Can we please remove the hard-coded name/namespace here? Also, who creates this secret? I don't see it in the Alicloud extension Helm chart. 
I suggested to provide the credentials to the Alicloud Helm chart, generate a secret object out of it, and make the name/namespace known to alicloud controller (probably via its component config).

controllers/provider-alicloud/pkg/controller/infrastructure/actuator.go
```diff
@@ -98,6 +109,13 @@ type actuator struct {
 	config *rest.Config
 
 	chartRenderer chartrenderer.Interface
+
+	machineImageMapping []config.MachineImage
+}
+
+func (a *actuator) InjectSeedAlicloudECSClient(alicloudECSClient alicloudclient.ECS) error {
```
rfranzke: I don't see this function being used anywhere. Do I miss something?

controllers/provider-alicloud/pkg/controller/infrastructure/actuator.go
```diff
@@ -265,8 +283,96 @@ func computeProviderStatusVSwitches(infrastructure *alicloudv1alpha1.Infrastruct
 	return vswitchesToReturn, nil
 }
 
+// getSeedCloudProviderCredentials gets Seed's cloud provider's credentials
+func (a *actuator) getSeedCloudProviderCredentials(ctx context.Context) (*alicloud.Credentials, error) {
+	secretRef := &corev1.SecretReference{
+		Name:      SeedCloudProviderSecretName,
+		Namespace: SeedCloudProviderSecretNamespace,
+	}
+	credentials, err := alicloud.ReadCredentialsFromSecretRef(ctx, a.client, secretRef)
```
rfranzke:
```suggestion
	return alicloud.ReadCredentialsFromSecretRef(ctx, a.client, secretRef)
```

controllers/provider-alicloud/pkg/controller/infrastructure/actuator.go
```diff
@@ -265,8 +283,96 @@ func computeProviderStatusVSwitches(infrastructure *alicloudv1alpha1.Infrastruct
...
+// access them, these images will be shared with it from Seed's Alicloud account.
+func (a *actuator) shareCustomizedImages(ctx context.Context, infra *extensionsv1alpha1.Infrastructure, cluster *extensioncontroller.Cluster) error {
+	regionID := cluster.Shoot.Spec.Region
+	if a.alicloudECSClient == nil {
+		a.logger.Info("Creating Alicloud ECS client for Seed", "infrastructure", infra.Name)
+		seedCloudProviderCredentials, err := a.getSeedCloudProviderCredentials(ctx)
+		if err != nil {
+			return err
+		}
+		a.alicloudECSClient, err = a.newClientFactory.NewECSClient(ctx, regionID, seedCloudProviderCredentials.AccessKeyID, seedCloudProviderCredentials.AccessKeySecret)
```
rfranzke: I see the same code below. I suggest to remove it there and keep it here only.

controllers/provider-alicloud/pkg/controller/infrastructure/actuator.go
```diff
@@ -265,8 +283,96 @@ func computeProviderStatusVSwitches(infrastructure *alicloudv1alpha1.Infrastruct
...
+	shootCloudProviderAccountID, err := shootAlicloudSTSClient.GetAccountIDFromCallerIdentity(ctx)
+	if err != nil {
+		return err
+	}
+
+	a.logger.Info("Sharing customized image with Shoot's Alicloud account from Seed", "infrastructure", infra.Name)
+	for _, worker := range workers {
+		imageName := worker.Machine.Image.Name
+		imageVersion := worker.Machine.Image.Version
+		imageID, err := confighelper.FindImageForRegion(a.machineImageMapping, imageName, imageVersion, regionID)
```
rfranzke: I see a problem here: Let's assume a shoot with image `foo` and version `1.0`. Now, somebody deploys a new Alicloud extension with a new configmap that no longer contains the `1.0` version. Then, the infrastructure reconciliation for this shoot will always fail.
We have the same problem with the worker controller. Here, we put the real image ID information into the status:

```yaml
...
status:
  providerStatus:
    apiVersion: alicloud.provider.extensions.gardener.cloud/v1alpha1
    kind: WorkerStatus
    machineImages:
    - id: coreos_2023_4_0_64_30G_alibase_20190319.vhd
      name: coreos-alicloud
      version: 2023.4.0
```

I suggest to do either do the same in the infrastructure controller, or do fall back checking the worker status in case the image ID is not found in the config map. See also https://github.com/gardener/gardener-extensions/pull/217 for more info.

jia-jerry: It is an interesting case. Image sharing is a special step prior to Machine creation. I think we could just `skip sharing` if the image id can't be found according to version and name.  This is because, during whole shoot reconciliation, Machine process loop alway do the same image id searching. 
- If it is found, that means in the infrastructure loop image has already been shared.
- If it is not found, even in the infrasturcture loop the image can be found by some means and shared, it doesn't make any sense. Because machine loop will always fail.

Hope I explained it.

EmoinLanyu: Thanks for advice! However, Imho, this is different from the previous worker controller problem.

1. In worker controller, it needs to find a valid imageID to be able to deploy the machine.
1. In infrastructure controller, it only needs to share the image the extension provided to Shoot's alicloud account when the Shoot doesn't have access to the image. Theoretically, if the Shoot has already been reconciled before, the image should have also already been shared with its Alicloud account, thus, no more sharing is needed. In this case, not finding the imageID is okay.

I do, however, see a potential issue: Assume a Shoot with image `foo` and version `1.0`. As new Alicloud extension is deployed, and the new configmap no longer contains the version `1.0` `foo` image. Since we assume the image has already been shared with its Alicloud account, no sharing will occur. But, somehow, the sharing from Seed's Alicloud account to Shoot's Alicloud account is revoked (which normally shouldn't happen), Shoot worker will thus no longer be able to reconcile.

Do we need to address this problem? If we do, then surely we can store the status in infrastructureStatus (or check workerStatus). But if this is not necessary, then we can simply remove `return err` when imageID is not found by `FindImageForRegion()` (yes, in any case, simply returning err here is not proper)

EmoinLanyu: > I see a problem here: Let's assume a shoot with image `foo` and version `1.0`. Now, somebody deploys a new Alicloud extension with a new configmap that no longer contains the `1.0` version. Then, the infrastructure reconciliation for this shoot will always fail.
> We have the same problem with the worker controller. Here, we put the real image ID information into the status:
> 
> ```yaml
> ...
> status:
>   providerStatus:
>     apiVersion: alicloud.provider.extensions.gardener.cloud/v1alpha1
>     kind: WorkerStatus
>     machineImages:
>     - id: coreos_2023_4_0_64_30G_alibase_20190319.vhd
>       name: coreos-alicloud
>       version: 2023.4.0
> ```
> 
> I suggest to do either do the same in the infrastructure controller, or do fall back checking the worker status in case the image ID is not found in the config map. See also #217 for more info.

Thanks for advice! However, Imho, this is different from the previous worker controller problem.

1. In worker controller, it needs to find a valid imageID to be able to deploy the machine.
1. In infrastructure controller, it only needs to share the image the extension provided to Shoot's alicloud account when the Shoot doesn't have access to the image. Theoretically, if the Shoot has already been reconciled before, the image should have also already been shared with its Alicloud account, thus, no more sharing is needed. In this case, not finding the imageID is okay.

I do, however, see a potential issue: Assume a Shoot with image `foo` and version `1.0`. As new Alicloud extension is deployed, and the new configmap no longer contains the version `1.0` `foo` image. Since we assume the image has already been shared with its Alicloud account, no sharing will occur. But, somehow, the sharing from Seed's Alicloud account to Shoot's Alicloud account is revoked (which normally shouldn't happen), Shoot worker will thus no longer be able to reconcile.

Do we need to address this problem? If we do, then surely we can store the status in infrastructureStatus (or check workerStatus). But if this is not necessary, then we can simply remove `return err` when imageID is not found by `FindImageForRegion()` (yes, in any case, simply returning err here is not proper)

rfranzke: Yes, good points colleagues, thanks. @EmoinLanyu Ideally, you would address this point, yes, but I can also understand that a pragmatic way to start is to follow @jia-jerry's approach - only do the sharing once when the `Infrastructure` is initially created, but then not again. This, however, would not be very much controller-like, where the desired state is always reconciled. If the machine image is revoked from end-user account then manual action is always required, something, that we generally try to avoid when we can. And here we clearly can. But I also understand that this is most likely a rare case. Though, I personally would probably simply store the image versions in the infra status (not looking into the `Worker` object), just to be safe.

controllers/provider-alicloud/pkg/controller/infrastructure/types.go
jia-jerry: We added a secret ref in the extension configuration. However,  there is no automatic way to generate such a secret. Even in ls-setup, we can't create a secret in seed cluster because during ls-setup process, seed clusters may not be ready. For current, I would like to as operators to manually create one. New version of OS is critical for AliCloud to release. So shall we just take this solution for this Takt? In the next Takt, I will firgure out a better way.

controllers/provider-alicloud/pkg/alicloud/client/client.go
```diff
@@ -151,3 +153,82 @@ func (c *storageClient) DeleteBucketIfExists(ctx context.Context, bucketName str
...
+// CheckIfImageExists checks whether given imageID can be accessed by the client
+func (c *ecsClient) CheckIfImageExists(ctx context.Context, imageID string) (bool, error) {
+	request := ecs.CreateDescribeImagesRequest()
+	request.ImageId = imageID
+	request.SetScheme("HTTPS")
+	response, err := c.client.DescribeImages(request)
+	if err != nil {
+		return false, err
+	}
+	if response.TotalCount > 0 {
```
ialidzhikov:
```suggestion
	return response.TotalCount > 0, nil
```

controllers/provider-alicloud/pkg/controller/infrastructure/actuator.go
```diff
@@ -265,6 +287,106 @@ func computeProviderStatusVSwitches(infrastructure *alicloudv1alpha1.Infrastruct
...
+		})
+
+		exists, err := shootAlicloudECSClient.CheckIfImageExists(ctx, imageID)
+		if err != nil {
+			return nil, err
+		}
+		if exists {
+			continue
+		}
+		err = a.alicloudECSClient.ShareImageToAccount(ctx, imageID, shootCloudProviderAccountID)
```
ialidzhikov:
```suggestion
		if err := a.alicloudECSClient.ShareImageToAccount(ctx, imageID, shootCloudProviderAccountID); err != nil {
```

controllers/provider-alicloud/pkg/alicloud/client/client.go
```diff
@@ -151,3 +153,82 @@ func (c *storageClient) DeleteBucketIfExists(ctx context.Context, bucketName str
...
+	return false, nil
+}
+
+// ShareImageToAccount shares the given image to target account from current client
+func (c *ecsClient) ShareImageToAccount(ctx context.Context, imageID, accountID string) error {
+	request := ecs.CreateModifyImageSharePermissionRequest()
+	request.ImageId = imageID
+	request.AddAccount = &[]string{accountID}
+	request.SetScheme("HTTPS")
+	_, err := c.client.ModifyImageSharePermission(request)
```
ialidzhikov:
```suggestion
	_, err := c.client.ModifyImageSharePermission(request)
	return err
```

EmoinLanyu:
> Does it makes sense also to add docs about the image sharing/custom images in `controllers/provider-alicloud/docs/usage-as-end-user.md`/`controllers/provider-alicloud/docs/usage-as-operator.md`?

Yes, good point indeed! Will add it too.

ialidzhikov:
One more question - shouldn't we preserve also the existing public image scenario somehow? Currently I try to pass a worker
```yaml
workers:
      - autoScalerMax: 1
        autoScalerMin: 1
        machineImage:
          name: coreos-alicloud
          version: 2023.4.0
        machineType: ecs.c5.large
```

referencing the public CoreOS image. And the infra fails to reconcile because I didn't specified the `machine-image-owner` secret.

EmoinLanyu:
> One more question - shouldn't we preserve also the existing public image scenario somehow? Currently I try to pass a worker
> 
> ```yaml
> workers:
> - autoScalerMax: 1
>   autoScalerMin: 1
>   machineImage:
>     name: coreos-alicloud
>     version: 2023.4.0
>   machineType: ecs.c5.large
> ```
> 
> referencing the public CoreOS image. And the infra fails to reconcile because I didn't specified the `machine-image-owner` secret.

Sorry for the late reply, I somehow missed your later comments...
Yes, this is because currently Alicloud ECS client is instantiated only once for one extension controller and the initialization is before sharing the image (we are also moving the instantiation to Actuator instantiation as well).
In future setup, which we are working on, this `machine-image-owner` secret should be created in extension's namespace in seeds by default (since only Alicloud provider is needing it, we don't want to simply adding it the way backup bucket did. So it's kind of tricky so we are looking for a good solution).
As a matter of fact, since we have talked Alicloud into providing a new CoreOS version, we have more time reconsider this sharing process and find a better way to pass this secret credentials.

EmoinLanyu:
Hi @ialidzhikov @rfranzke , I have made some refactoring and added related documents for customized image sharing. Could you please do another round of review? Thanks a lot!
So Alicloud ECS Client for Seed now is instantiated only once when the controller is added to controller manager (during APIReader injection). Secret reference will be checked, and only when it exists the client will be created, so landscapes that don't use customized images won't be affected.
As specified in the docs file, administrator needs to explicitly provide access credentials of Seed operator in the controller registration configuration (TODO: encrypt ctrlreg). Related PR is also created for landscape-setup.

rfranzke:
@ialidzhikov could you check again as well?

controllers/provider-alicloud/charts/provider-alicloud/templates/secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
```
jia-jerry: If .Values.config.machineImageOwnerSecret is not set, let's not create the secret.

controllers/provider-alicloud/charts/provider-alicloud/values.yaml
```diff
@@ -62,6 +62,10 @@ config:
       capacity: 25Gi
 #   backup:
 #     schedule: "0 */24 * * *"
+  machineImageOwnerSecret:
+    name: machine-image-owner
+    accessKeyID: ZHVtbXk=
+    accessKeySecret: ZHVtbXk=
```
jia-jerry: Please mark them off. Otherwise, it will be used as an default value.

controllers/provider-alicloud/pkg/controller/infrastructure/add.go
```diff
@@ -32,13 +34,17 @@ type AddOptions struct {
 	Controller controller.Options
 	// IgnoreOperationAnnotation specifies whether to ignore the operation annotation or not.
 	IgnoreOperationAnnotation bool
+	// MachineImages is the default list of machine images.
+	MachineImages []config.MachineImage
+	// MachineImageOwnerSecretRef is the secret reference which contains credential of AliCloud subaccount for customized images.
+	MachineImageOwnerSecretRef corev1.SecretReference
```
jia-jerry: As MachineImageOwnerSecretRef could be a nil, please define it as *corev1.SecretReference

ialidzhikov:
@EmoinLanyu , could you please check @jia-jerry's comments? Thanks!

EmoinLanyu:
> @EmoinLanyu , could you please check @jia-jerry's comments? Thanks!

I've changed them locally and have been testing it, but was occupied by other stuff so didn't create a PR. Anyway, thanks for the reminder, will do asap! :=)

EmoinLanyu:
> @EmoinLanyu , could you please check @jia-jerry's comments? Thanks!

Hi @ialidzhikov , done, please check https://github.com/gardener/gardener-extensions/pull/483, thanks!

PR #432, opened 2019-11-13
Provide option to disable calico-typha

vpnachev:
Provide option to disable calico-typha.
Contributes to  https://github.com/gardener/gardener/issues/1579

**Release note**:

```improvement user
Calico Typha can be disabled for shoot clusters via `Shoot.spec.networking.providerConfig.typha.enabled=false`. Please read the [docs](https://docs.projectcalico.org/v3.9/reference/typha/) before to consider to disable typha.
```


DockToFuture:
I'm wondering why we are deploying  ```calico-typha-horizontal-autoscaler```, ```calico-typha-vertical-autoscaler``` in to the cluster with ```replica=0``` when we disable typha. I would expect that we completely remove all deployments, svc, sa and so on, if we disable the addon.

vpnachev:
@DockToFuture I have no strong opinion about what should be deployed and what not when typha is disabled. I just started with the very minimalistic approach. 
Basically, there is a lot of other resources like (Cluster)Role(Bindings), PodDisruptionBudget, and so on, that I do not want to touch at the first time.
I think to move all typha related resources to dedicated file(s) and they will not be rendered when typha is disabled - wdyt?

rfranzke:
I agree with @DockToFuture, and I like your proposal @vpnachev.

DockToFuture:
I like the idea of dedicated files.

vpnachev:
@DockToFuture @zanetworker , this PR is ready for review.

controllers/networking-calico/charts/internal/calico/templates/rbac.yaml
```diff
@@ -208,89 +210,3 @@ subjects:
 - kind: ServiceAccount
   name: calico-kube-controllers
   namespace: kube-system
----
-apiVersion: {{ include "rbacversion" . }}
```
rfranzke: Where did this cluster role go?

controllers/networking-calico/example/20-network.yaml
```diff
@@ -50,5 +50,6 @@ spec:
     ipam:
       type: host-local
       cidr: usePodCIDR
-#    ipAutodetectionMethod: interface=eth0
-#    ipip: CrossSubnet
\ No newline at end of file
+#   ipAutodetectionMethod: interface=eth0
+#   ipip: CrossSubnet
+#   enableTypha: true
```
rfranzke:
```suggestion
#   typha:
#     enabled: true  
```

controllers/networking-calico/charts/internal/calico/templates/psp/calico-typha-clusterrole.yaml
```diff
@@ -1,4 +1,4 @@
-
+{{- if .Values.config.enableTypha }}
```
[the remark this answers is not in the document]
vpnachev: My idea was to move out `typha` related stuffs to a dedicated file only from `calico.yaml`.
Most of the RBAC resource were in separate files, so I just exclude them when typha is disabled.
I will merge the typha related RBAC files.

> Maybe it makes sense to even split typha/vertical-autoscaler/horizontal-autoscaler into three dedicated files.

I have no strong opinion, so do I have to do it or not?

controllers/networking-calico/charts/internal/calico/templates/rbac.yaml
vpnachev: here https://github.com/gardener/gardener-extensions/pull/432/files#diff-c202b3b53bd963173804776737230878R28

controllers/networking-calico/charts/internal/calico/templates/psp/calico-typha-clusterrole.yaml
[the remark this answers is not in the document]
rfranzke: I think how it is now it's kind of inconsistent. Either put everything together in one file, or separate every resource into a dedicated file and not only RBAC. 

> Maybe it makes sense to even split typha/vertical-autoscaler/horizontal-autoscaler into three dedicated files.

No strong opinion either, I personally would probably do it, but then I would also separate the `calico-kube-controllers` related stuff into a separate file. Then we have four files for the four components. But as said, no strong opinion.

controllers/networking-calico/charts/internal/calico/templates/rbac.yaml
rfranzke: Thanks!

controllers/networking-calico/charts/internal/calico/templates/psp/calico-typha-clusterrole.yaml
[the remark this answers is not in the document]
vpnachev: Now everything is merged into just couple of files:
1. `calico.yaml` - the minimal calico resource that are deployed always
1. `calico-typha.yaml` - resources that are deployed only when typha is enabled
1. Three files containing configmaps used for sha256 sums calculations.

controllers/networking-calico/charts/internal/calico/values.yaml
```diff
@@ -4,7 +4,9 @@ config:
   backend: bird
   ipam:
     type: "host-local"
-#    subnet: "usePodCidr"
+#   subnet: "usePodCidr"
+  typha: {}
+#   enabled: true
```
rfranzke: Can we uncomment this as this should be the default? I know, it's overwritten when actually rendering the chart, though, just to be safe.

vpnachev: done

ialidzhikov:
> What was the reason to put everything in 2 files? According to https://helm.sh/docs/topics/chart_best_practices/templates/:
> 
> > Each resource definition should be in its own template file.
> 
> Currently calico-typha.yaml is 495 lines long, calico.yaml - 861.

We can ignore this comment for now and fix it later.

vpnachev:
Added end-user docs about how to disable Typha.
Thanks @ialidzhikov !

PR #450, opened 2019-11-21
Expose CloudNAT Minimal Ports per VM

### LINKED ISSUE #445: [GCP] Expose configuration for minimum ports per VM instance in CloudNAT
With #417 the shoot worker nodes get outbound network connectivity via CloudNAT instead of external IP per node. By default, CloudNAT has a minimum ports per VM set to 64 which results in dropped packets when more than 64 connection are tried to be established from the same node to the same destination. To resolve this, the minimum ports per VM configuration has to be exposed in the Infrastructure API, as a reasonable default value seems to be 2048 ports(allowing around 30 nodes to be exposed via single NAT IP).

Reference:
* https://cloud.google.com/nat/docs/overview

vpnachev:
Expose CloudNAT Minimal Ports per VM

Fixes #445 

**Release note**:

```improvement operator
An issue with dropped packets by CloudNAT for GCP clusters has been fixed.
```

```noteworthy user
The minimum number of ports allocated to a VM in the CloudNAT resource can be defined in the `gcp.provider.extensions.gardener.cloud/v1alpha1.InfrastructureConfig.networks.cloudNAT.minPortsPerVM`
```


zanetworker (dismissed):
Looks good :)

DockToFuture (dismissed):
/lgtm

zanetworker (approved):

ialidzhikov (approved):
/approve

PR #465, opened 2019-11-26
Split calico chart files

### LINKED ISSUE #439: Clean up kubernetes.io/cluster-service from calico extension
`kubernetes.io/cluster-service` is deprecated label by addon-manager. Now it is used only by kubectl for `cluster-info` command and it is respected only for `kind: Service` - kubernetes/kubernetes#72757 and gardener/gardener#997.
We could clean up this label from the calico chart for all non `kind: Service`s.

/kind flake

### LINKED ISSUE #462: Split calico chart to separate template files
#432 puts different resources in single files which is against to one of the helm chart recommendations that each resource definition should be in its own template file. Ref https://helm.sh/docs/topics/chart_best_practices/templates/.

vpnachev:
Split calico chart files and clean up the label `kubernetes.io/cluster-service`

Fixes #462, fixes #439 

**Release note**:

```improvement operator
NONE
```


zanetworker:
@vpnachev do you think it makes sense to keep a layer of abstraction on top? otherwise with every new resource / change we add another file(s). For example, we can have RBAC, CRDs, Config, Deployment or something like that, wdyt?

vpnachev:
The only option I see is to split this charto it into multiple sub-charts, but the structure will be the same as it is now. So I do not see any pros at the moment.

rfranzke:
+1 for how got done by @vpnachev. It's way more clear to have a dedicated file per resource.

zanetworker:
@rfranzke do we follow this pattern everywhere / should we?

rfranzke:
We don't follow it everywhere, unfortunately, but we can introduce it as "convention" for new charts, at least, and refactor existing ones when we have to touch them anyways.

zanetworker (dismissed):

rfranzke:
Can we first merge #460 and then rebase this one on top?

zanetworker:
sure, then let's hold it for now.

rfranzke:
#460 is merged - can you rebase, please?

rfranzke:
Can we prefix the filenames with the Kubernetes resource kind?

`deployment-<...>`

If you want to group certain resources belonging to a component together use folders:

```
templates
  - calico-node
    - daemonset-calico-node.yaml
    - configmap-<...>
  - calico-typha
    - ...
```

rfranzke:
@zanetworker @DockToFuture ready for review

vpnachev:
Done, I have switched from suffixes to prefixes.

zanetworker (approved):
Yosh, just spun up a cluster, everything looks good.

rfranzke (approved):

PR #474, opened 2019-12-02
Change computation of hash value for machine resources

rfranzke:
In order to prevent undesired rolling updates the hash value for machine resources is now computed only out of 
* machine type
* volume type + size
* worker pool provider config
* kubernetes version
* machine image name + version

/cc @vlerenc 

**Release note**:

```noteworthy user
:warning: In order to prevent undesired rolling updates of machine resources the hash value computation has changed. A rolling update of worker nodes is now only happening if one of the following properties is changed: Kubernetes major/minor version, machine image name or version, machine type, volume type or size, worker pool provider configuration. Deploying this change will, one last time, roll out all nodes of all clusters.
```


rfranzke:
done

vpnachev:
Should we state in the rel note that the roll out will happen in the maintenance time window?

PR #476, opened 2019-12-02
Put Infrastructure state into status as runtime.RawExtension

vlvasilev:
With this PR we copy the terraformer state from the configmap <shoot-name>.infra.tf-state into the Infrastructure's Status under State as runtime.RawExtension. The State is base64 encoded to make the infra resource small as posible. 

**Release note**:

```improvement operator
All `Infrastructure` extension controllers do now copy the Terraformer state into the `Infrastructure`'s `.status.state` field.
```

### DIFF
--- a/controllers/provider-aws/pkg/controller/infrastructure/actuator_reconcile.go
+++ b/controllers/provider-aws/pkg/controller/infrastructure/actuator_reconcile.go
@@ -56,6 +56,11 @@ func (a *actuator) reconcile(ctx context.Context, infrastructure *extensionsv1al
 		return fmt.Errorf("failed to generate Terraform config: %+v", err)
 	}
 
+	terraformState, err := terraformer.UnmarshalRawState(infrastructure.Status.State)
+	if err != nil {
+		return err
+	}
+
 	chartRenderer, err := chartrenderer.NewForConfig(a.restConfig)
 	if err != nil {
 		return fmt.Errorf("could not create chart renderer: %+v", err)
@@ -77,7 +82,8 @@ func (a *actuator) reconcile(ctx context.Context, infrastructure *extensionsv1al
 			a.client,
 			release.FileContent("main.tf"),
 			release.FileContent("variables.tf"),
-			[]byte(release.FileContent("terraform.tfvars"))),
+			[]byte(release.FileContent("terraform.tfvars")),
+			terraformState.Data),
 		).
 		Apply(); err != nil {
 
@@ -180,6 +186,15 @@ func (a *actuator) updateProviderStatus(ctx context.Context, tf terraformer.Terr
 		return err
 	}
 
+	state, err := tf.GetRawState(ctx)
+	if err != nil {
+		return err
+	}
+	stateByte, err := state.Marshal()
+	if err != nil {
+		return err
+	}
+
 	subnets, err := computeProviderStatusSubnets(infrastructureConfig, output)
 	if err != nil {
 		return err
@@ -221,6 +236,7 @@ func (a *actuator) updateProviderStatus(ctx context.Context, tf terraformer.Terr
 				},
 			},
 		}
+		infrastructure.Status.State = &runtime.RawExtension{Raw: stateByte}
 		return nil
 	})
 }
--- a/controllers/provider-azure/pkg/controller/infrastructure/actuator.go
+++ b/controllers/provider-azure/pkg/controller/infrastructure/actuator.go
@@ -16,7 +16,6 @@ package infrastructure
 
 import (
 	"context"
-
 	azurev1alpha1 "github.com/gardener/gardener-extensions/controllers/provider-azure/pkg/apis/azure/v1alpha1"
 	infrainternal "github.com/gardener/gardener-extensions/controllers/provider-azure/pkg/internal/infrastructure"
 	extensionscontroller "github.com/gardener/gardener-extensions/pkg/controller"
@@ -79,8 +78,18 @@ func (a *actuator) updateProviderStatus(
 		return err
 	}
 
+	state, err := tf.GetRawState(ctx)
+	if err != nil {
+		return err
+	}
+	stateByte, err := state.Marshal()
+	if err != nil {
+		return err
+	}
+
 	return extensionscontroller.TryUpdateStatus(ctx, retry.DefaultBackoff, a.client, infra, func() error {
 		infra.Status.ProviderStatus = &runtime.RawExtension{Object: status}
+		infra.Status.State = &runtime.RawExtension{Raw: stateByte}
 		return nil
 	})
 }
--- a/controllers/provider-gcp/pkg/controller/infrastructure/actuator.go
+++ b/controllers/provider-gcp/pkg/controller/infrastructure/actuator.go
@@ -79,8 +79,18 @@ func (a *actuator) updateProviderStatus(
 		return err
 	}
 
+	state, err := tf.GetRawState(ctx)
+	if err != nil {
+		return err
+	}
+	stateByte, err := state.Marshal()
+	if err != nil {
+		return err
+	}
+
 	return extensionscontroller.TryUpdateStatus(ctx, retry.DefaultBackoff, a.client, infra, func() error {
 		infra.Status.ProviderStatus = &runtime.RawExtension{Object: status}
+		infra.Status.State = &runtime.RawExtension{Raw: stateByte}
 		return nil
 	})
 }
--- a/controllers/provider-openstack/pkg/controller/infrastructure/actuator.go
+++ b/controllers/provider-openstack/pkg/controller/infrastructure/actuator.go
@@ -16,6 +16,7 @@ package infrastructure
 
 import (
 	"context"
+
 	"github.com/gardener/gardener/pkg/chartrenderer"
 	"k8s.io/client-go/util/retry"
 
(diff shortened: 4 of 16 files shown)


rfranzke:
Am I correct that the following will happen now: In every `Infrastructure` reconciliation we update the state configmap with the data found in the `.status.state` field of the `Infrastructure resource before we kick off the TF pod? Previously (without this PR) we only rely on the state inside the configmap.

vlvasilev:
@rfranzke Yes. The state in the configmap is copied, encoded and set as state. The state from Infrastructure can be restored only if the configmap is missing. This is the case of a controlplane migration.

rfranzke:
I have the feeling that we should only rely on the `status.state` data in case the Terraform state configmap does not exist (control plane migration case) or is missing/corrupt. If we always rely on it we introduce a new surface for issues, e.g., when the extension controller dies after it has kicked off the Terraform pod. In this case, the pod will do something, update the state configmap, but nothing gets copied back to the `Infrastructure` status. In the next reconciliation the extension would overwrite the (correct) state configmap with an empty or outdated state.

vlvasilev:
@rfranzke  I am very sorry. I misunderstood you earlier. If the configmap exists the extension controller will not override it. The `status.state` will be used only if the configmap is missing.

rfranzke:
Okay, thanks, that's good!

pkg/terraformer/config.go
```diff
@@ -101,6 +101,18 @@ func (t *terraformer) InitializeWith(initializer Initializer) Terraformer {
...
 
+//GetState returns the contenten of terraform state config map
+func (t *terraformer) GetRawState(ctx context.Context) (*RawState, error) {
+	configMap := &corev1.ConfigMap{}
+	if err := t.client.Get(ctx, kutil.Key(t.namespace, t.stateName), configMap); err != nil {
+		return nil, err
+	}
+	return &RawState{
+		Data:   configMap.Data[StateKey],
+		Encode: "none",
```
ialidzhikov: You could introduce constants for `none` and `base64`. Currently they are repeated multiple times.

pkg/terraformer/terraformer.go
```diff
@@ -37,6 +39,8 @@ import (
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
```
ialidzhikov:
```suggestion

```

pkg/terraformer/terraformer.go
```diff
@@ -510,3 +514,84 @@ func (t *terraformer) cleanupJob(ctx context.Context, jobPodList *corev1.PodList
 func (t *terraformer) computePodGenerateName(command string) string {
 	return fmt.Sprintf("%s.%s.tf-%s-", t.name, t.purpose, command)
 }
+
+//*******Terraform Raw State******
```
ialidzhikov:
```suggestion

```

pkg/terraformer/terraformer.go
```diff
@@ -510,3 +514,84 @@ func (t *terraformer) cleanupJob(ctx context.Context, jobPodList *corev1.PodList
...
+
+// Marshal transform RawState to []byte representation. It encodes the raw state data
+func (trs *RawState) Marshal() ([]byte, error) {
```
ialidzhikov: I guess you could split rawState related funcs to `pkg/terraformer/raw_state.go`.

pkg/terraformer/types.go
```diff
@@ -63,6 +64,12 @@ type terraformer struct {
 	deadlinePod      time.Duration
 }
 
+//RawState represent the terraformer state's raw data
```
ialidzhikov:
```suggestion
// RawState represent the terraformer state's raw data
```

pkg/terraformer/config.go
```diff
@@ -101,6 +101,18 @@ func (t *terraformer) InitializeWith(initializer Initializer) Terraformer {
 	return t
 }
 
+//GetState returns the contenten of terraform state config map
```
ialidzhikov:
```suggestion
// GetStateConfigMapAsRawState returns the content of terraform state config map.
```

I think `GetStateConfigMapAsRawState ` is a better self-explaining name.

pkg/terraformer/config.go
ialidzhikov:
```suggestion
		Encode: NoneEncoding,
```

pkg/terraformer/raw_state.go
```diff
@@ -0,0 +1,102 @@
...
+			rawData = v.Raw
+		}
+	case []byte:
+		rawData = v
+	case string:
+		rawData = []byte(v)
+	case nil:
+		rawData = nil
+	default:
+		return nil, fmt.Errorf("unsuported type \"%T\" for unmarshaling Raw Terraform State", rawState)
```
ialidzhikov:
```suggestion
		return nil, fmt.Errorf("unsuported type '%T' for unmarshaling raw Terraform state", rawState)
```

controllers/provider-aws/pkg/controller/infrastructure/actuator_reconcile.go
```diff
@@ -180,6 +186,15 @@ func (a *actuator) updateProviderStatus(ctx context.Context, tf terraformer.Terr
 		return err
 	}
 
+	state, err := tf.GetRawState(context.TODO())
```
ialidzhikov:
```suggestion
	state, err := tf.GetRawState(ctx)
```

controllers/provider-azure/pkg/controller/infrastructure/actuator.go
```diff
@@ -79,8 +78,18 @@ func (a *actuator) updateProviderStatus(
 		return err
 	}
 
+	state, err := tf.GetRawState(context.TODO())
```
ialidzhikov:
```suggestion
	state, err := tf.GetRawState(ctx)
```

controllers/provider-gcp/pkg/controller/infrastructure/actuator.go
```diff
@@ -79,8 +79,18 @@ func (a *actuator) updateProviderStatus(
 		return err
 	}
 
+	state, err := tf.GetRawState(context.TODO())
```
ialidzhikov:
```suggestion
	state, err := tf.GetRawState(ctx)
```

controllers/provider-openstack/pkg/controller/infrastructure/actuator.go
```diff
@@ -101,8 +102,18 @@ func (a *actuator) updateProviderStatus(
 		return err
 	}
 
+	state, err := tf.GetRawState(context.TODO())
```
ialidzhikov:
```suggestion
	state, err := tf.GetRawState(ctx)
```

controllers/provider-packet/pkg/controller/infrastructure/actuator_reconcile.go
```diff
@@ -102,6 +108,15 @@ func (a *actuator) updateProviderStatus(ctx context.Context, tf terraformer.Terr
 		return err
 	}
 
+	state, err := tf.GetRawState(context.TODO())
```
ialidzhikov:
```suggestion
	state, err := tf.GetRawState(ctx)
```

pkg/terraformer/types.go
```diff
@@ -63,6 +64,12 @@ type terraformer struct {
 	deadlinePod      time.Duration
 }
 
+// RawState represent the terraformer state's raw data
+type RawState struct {
+	Data   string `json:"data"`
+	Encode string `json:"encode"`
```
ialidzhikov: `encode` is a verb, can we change to `encoding`? (as the coreos-cloudinit, the have similar `encoding` field)

```suggestion
	Encoding string `json:"encoding"`
```

pkg/terraformer/config.go
vlvasilev: At the moment we are using config map but if we change it to secret we have to change the name of the method. No matter what kind is the k8s resource in which we store the state we get the same RawState object.

pkg/terraformer/raw_state.go
```diff
@@ -0,0 +1,102 @@
...
+			rawData = v.Raw
+		}
+	case []byte:
+		rawData = v
+	case string:
+		rawData = []byte(v)
+	case nil:
+		rawData = nil
+	default:
+		return nil, fmt.Errorf("unsuported type '%T' for unmarshaling Raw Terraform State", rawState)
```
rfranzke:
```suggestion
		return nil, fmt.Errorf("unsupported type '%T' for unmarshaling raw Terraform state", rawState)
```

pkg/terraformer/config.go
```diff
@@ -101,6 +101,18 @@ func (t *terraformer) InitializeWith(initializer Initializer) Terraformer {
 	return t
 }
 
+//GetRawState returns the contenten of terraform state config map
```
plkokanov:
```suggestion
//GetRawState returns the content of terraform state config map
```

pkg/terraformer/raw_state.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	if err := json.Unmarshal(terraformRawState, trs); err != nil {
+		return nil, err
+	}
+	return trs, nil
+}
+
+// encodeBase64 encode the RawState.Data if it is not already base64 encoded
+func (trs *RawState) encodeBase64() *RawState {
```
plkokanov: The return value of encodeBase64 is never used. And since you don't copy an object and encode it I don't see the point of the return value.

vlvasilev: @plkokanov  What if I change func (trs *RawState) Marshal() ([]byte, error)
```suggestion
func (trs *RawState) Marshal() ([]byte, error) {
	return json.Marshal(trs.encodeBase64())
}
```

PR #492, opened 2019-12-09
configure shoot worker node calico port mtu

jiangytcn:
let user configure the mtu for shoot worker node
Fixes #

**Release note**:
NONE

```improvement operator
MTU value can now be configured as part of the NetworkConfig
```

```noteworthy user
If you wish to configure the MTU value, please make sure that all the MTUs in your environment, failure to do so will result in performance penalties and packet loss.
```


vpnachev:
@jiangytcn thanks for the PR, but this is not sufficient to make mtu configurable.
You have to extend the internal structure here https://github.com/gardener/gardener-extensions/blob/3b288ffe55333be7d980ccc20e5e7a9b1b9ca354/controllers/networking-calico/pkg/apis/calico/v1alpha1/types_network.go#L68-L95 and to parse the value here https://github.com/gardener/gardener-extensions/blob/master/controllers/networking-calico/pkg/charts/utils.go

jiangytcn:
thank you @vpnachev , working on it

jiangytcn:
hi @vpnachev  is this sufficient to configure mtu ?

vpnachev:
Not exactly, you need to add it in the internal version, too. https://github.com/gardener/gardener-extensions/blob/3b288ffe55333be7d980ccc20e5e7a9b1b9ca354/controllers/networking-calico/pkg/apis/calico/types_network.go#L64-L85

As an example, you can see how I added the option typha to be disabled here https://github.com/gardener/gardener-extensions/pull/432

Also, run `make generate` to generate the conversion and defaulting code.


PS. Do not merge the master into your branch but fetch and rebase, otherwise I am not sure whether we will be able to merge it.

jiangytcn:
can you please give me some guidance on  how to run the changes into a live gardener cluster ? 
build the image locally and replace  the one ran in networking-calico-xxx namespace  ?

vpnachev:
You should have resource ControllerRegistration `networking-calico` in your garden cluster(the cluster where you can get shoots,seeds).

So, you can provide your own image repository and tag in the controller registration https://github.com/gardener/gardener-extensions/blob/4d3ae875d03c74184cef35eaed195cabb9980592/controllers/networking-calico/example/controller-registration.yaml#L15-L16

The default image repository is `eu.gcr.io/gardener-project/gardener/gardener-extension-hyper` , but you do not have permissions to push images there, so you should switch to your own image registry, at the end the image section should  look like this:
```yaml
image:
  repository: <your-image-registry>
  tag: <your-image-tag>
```

jiangytcn:
created a new one in https://github.com/gardener/gardener-extensions/pull/502

PR #498, opened 2019-12-13
Azure: Add storage class for Azure Disk Standard SSD LRS

dkistner:
Add a Azure storage class for the standard ssd volume type

```noteworthy users
Azure Storage Classes for azure-disk Standards LRS Sku are now available for all Azure cluster >= v1.13.
```


controllers/provider-azure/charts/internal/shoot-storageclasses/templates/storageclasses.yaml
```diff
@@ -21,6 +21,15 @@ parameters:
 ---
 apiVersion: {{ include "storageclassversion" . }}
 kind: StorageClass
+metadata:
+  name: managed-standard-ssd
+provisioner: kubernetes.io/azure-disk
+parameters:
+  storageaccounttype: StandardSSD_LRS
```
ialidzhikov: This storageclass is only available for Kubernetes >= 1.13 - ref https://github.com/kubernetes/kubernetes/issues/73260#issuecomment-457252026.
Can we pass the Shoot Kubernetes version as a value and do not apply this storage class when the version is < 1.13?

```
$ k describe pvc azure-managed-disk
Events:
  Type       Reason              Age               From                         Message
  ----       ------              ----              ----                         -------
  Warning    ProvisioningFailed  6s (x3 over 27s)  persistentvolume-controller  Failed to provision volume with StorageClass "managed-standard-ssd": azureDisk - StandardSSD_LRS is not supported sku/storageaccounttype. Supported values are [Premium_LRS Standard_GRS Standard_LRS Standard_RAGRS Standard_ZRS]
Mounted By:  <none>
```

dkistner:
Thanks for the feedback. I have addressed it and updated the pr. I will run another test tomorrow.
/hold

dkistner:
I have updated and tested the pr once more. The storage classes are only deployed if the corresponding k8s version support it. I found two additional things:

1. Azure file with premium storage work only if the user request at least 100 Gi storage. This is an potential source of errors by users. I'm therefore think about not to include a storage class for `azure-file-premium` out of the box. @rfranzke WDYT?
2. Dynamic provisioning of Azure file shares require to allow the provisioner (service account `persistent-volume-binder`) to `read` and `create` secrets, which where not yet deployed by Gardener and therefore no out of the box use of Azure file in general was possible.

rfranzke:
Makes sense, let's not include it. Who needs it can easily create the `StorageClass` on his own.

dkistner:
I have removed the storageclass for `azure-file` with premium storage from this PR.

controllers/provider-azure/pkg/controller/controlplane/valuesprovider.go
```diff
@@ -289,3 +289,10 @@ func getInfraNames(infraStatus *apisazure.InfrastructureStatus) (string, string,
 
 	return nodesSubnet.Name, nodesRouteTable.Name, nodesSecurityGroup.Name, nil
 }
+
+// GetStorageClassesChartValues returns the values for the storage classes chart applied by this actuator.
+func (vp *valuesProvider) GetStorageClassesChartValues(ctx context.Context, cp *extensionsv1alpha1.ControlPlane, cluster *extensionscontroller.Cluster) (map[string]interface{}, error) {
```
rfranzke: As discussed and raised by @vpnachev  you probably can remove this code and check in the chart for `.Capabilities.Kube.Version` instead of `.Values.kubernetesVersion`.

vpnachev: Yes, it is already used in the very same chart https://github.com/gardener/gardener-extensions/blob/e76686203ce1dd0b1fd8b4ef92a33ebe8dae5914/controllers/provider-azure/charts/internal/shoot-storageclasses/templates/_helpers.tpl#L2

PR #499, opened 2019-12-13
Azure: Optimize infrastructure creation/deletion

### LINKED ISSUE #373: New Azure clusters often fail during initial creation
New Azure clusters often fail during initial creation because the Terraform infrastructure job returns errors that are confusing for end-users. After the next reconciliation/Terraform run it succeeds, however, following errors are shown:

```
-> Pod 'example.infra.tf-job-v8zct' reported:
* azurerm_availability_set.workers: 1 error occurred:
	
* azurerm_availability_set.workers: compute.AvailabilitySetsClient#CreateOrUpdate: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="ResourceGroupNotFound" Message="Resource group 'shoot--project--example' could not be found."
	
* azurerm_route_table.workers: 1 error occurred:
	
* azurerm_route_table.workers: Error Creating/Updating Route Table "worker_route_table" (Resource Group "shoot--project--example"): network.RouteTablesClient#CreateOrUpdate: Failure sending request: StatusCode=404 -- Original Error: Code="ResourceGroupNotFound" Message="Resource group 'shoot--project--example' could not be found."
* azurerm_virtual_network.vnet: 1 error occurred:
	
* azurerm_virtual_network.vnet: Error Creating/Updating Virtual Network "shoot--project--example" (Resource Group "shoot--project--example"): network.VirtualNetworksClient#CreateOrUpdate: Failure sending request: StatusCode=404 -- Original Error: Code="ResourceGroupNotFound" Message="Resource group 'shoot--project--example' could not be found."]
```

dkistner:
Azure infrastructure creation fail frequently because Terraform can't calculate the dependency graph correctly. As of today it will still work after some retries, but this will be improved with this PR. 

The correct resource group references are now passed to the respective resources and should therefore prevent infra failures due to `ResourceGroupNotFound` errors.

Fixes #373

**Release note**:

```improvement operator
Infrastructure creation and deletion are optimized for Azure.
```

### DIFF
--- a/controllers/provider-azure/charts/internal/azure-infra/templates/main.tf
+++ b/controllers/provider-azure/charts/internal/azure-infra/templates/main.tf
@@ -10,6 +10,10 @@ resource "azurerm_resource_group" "rg" {
   name     = "{{ required "resourceGroup.name is required" .Values.resourceGroup.name }}"
   location = "{{ required "azure.region is required" .Values.azure.region }}"
 }
+{{- else -}}
+data "azurerm_resource_group" "rg" {
+  name     = "{{ required "resourceGroup.name is required" .Values.resourceGroup.name }}"
+}
 {{- end}}
 
 #=====================================================================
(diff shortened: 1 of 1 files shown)


controllers/provider-azure/charts/internal/azure-infra/templates/main.tf
```diff
@@ -70,9 +94,15 @@ resource "azurerm_availability_set" "workers" {
 //= Output variables
 //=====================================================================
 
+{{ if .Values.create.resourceGroup -}}
```
rfranzke:
```suggestion
output "{{ .Values.outputKeys.resourceGroupName }}" {
{{ if .Values.create.resourceGroup -}}
  value = "${azurerm_resource_group.rg.name}"
{{ else -}}
  value = "${data.azurerm_resource_group.rg.name}"
{{ end -}}
}
```

dkistner: Nice catch. Thanks.

rfranzke:
@dkistner what is different now compared to last time and what would help fixing @ialidzhikov's observations?

dkistner:
@ialidzhikov Yes I was able to reproduce it. It happens sporadically to me. It seems to be also related to some missing resource references to the `azurerm_virtual_network` in the `azurerm_subnet` resource. 

Beside that I found also an issue for the existing vnet case which could lead also to `ResourceNotFound` issues. This should be now avoided by introducing a datasource for the existing vnet similar as we did it for existing resource groups.

Could you have another try?

PR #500, opened 2019-12-13
Use MTU of 1460 for all SUSE nodes

### LINKED ISSUE #497: Kubernetes clusters using SUSE JeOS on Azure do not become ready
**What happened**:

Note: this bug report is to investigate and document the findings for https://github.com/projectcalico/calico/issues/2973. It is as of now unclear which component is to blame. 

Problem is that the readiness check for felix fails (container in the calico-node pod) on nodes where calico typha is not running. Felix appears to work fine on nodes where typha is running locally. The problem does not appear to show up when typha is not used.

What is strange is that networking in the cluster appears to work fine and all the routes (at least in a small cluster) are created correctly by felix

The readiness check for felix succeeds.

Create a 2 node cluster on Azure with SuSE JeOS as operating system and calico typha. The readiness check will fail on the calico-node pod on the host where typha is not running.

**Anything else we need to know?**:

**Environment**:

- SUSE Linux Enterprise Server 15 SP1   4.12.14-197.21-default
- Azure

vpnachev:
Use MTU of 1460 for all SUSE nodes

Fixes #497 

Thanks to @marwinski  and @zanetworker .

**Release note**:

```improvement operator
SuSE nodes are now using mtu of 1460 on all infrastructures.
```

```improvement user
An issue where calico typha was not working with SuSE nodes has been fixed. Now SuSE based cluster can enable calico typha again and benefit from it.
```

### DIFF
--- a/controllers/os-suse-jeos/pkg/generator/templates/cloud-init/suse-jeos.template
+++ b/controllers/os-suse-jeos/pkg/generator/templates/cloud-init/suse-jeos.template
@@ -28,6 +28,8 @@ write_files:
 runcmd:
 - "until zypper -q install -y docker wget socat jq nfs-client; [ $? -ne 7 ]; do sleep 1; done"
 - systemctl daemon-reload
+- ip link set dev eth0 mtu 1460
+- grep -q '^MTU' /etc/sysconfig/network/ifcfg-eth0 && sed -i 's/^MTU.*/MTU=1460/' /etc/sysconfig/network/ifcfg-eth0 || echo 'MTU=1460' >> /etc/sysconfig/network/ifcfg-eth0
 {{ if .Bootstrap -}}
 - ln -s /usr/bin/docker /bin/docker
 - ln -s /bin/ip /usr/bin/ip
--- a/controllers/os-suse-jeos/pkg/generator/templates/script/suse-jeos.template
+++ b/controllers/os-suse-jeos/pkg/generator/templates/script/suse-jeos.template
@@ -26,6 +26,8 @@ mkdir -p '{{ $unit.DropIns.Path }}'
 
 until zypper -q install -y docker wget socat jq nfs-client; [ $? -ne 7 ]; do sleep 1; done
 systemctl daemon-reload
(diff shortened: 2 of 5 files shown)


controllers/os-suse-jeos/pkg/generator/templates/script/suse-jeos.template
```diff
@@ -26,6 +26,7 @@ mkdir -p '{{ $unit.DropIns.Path }}'
 
 until zypper -q install -y docker wget socat jq nfs-client; [ $? -ne 7 ]; do sleep 1; done
 systemctl daemon-reload
+ip link set dev eth0 mtu 1460
```
zanetworker: do we need it here as well?

vpnachev: Yes, we needed.
This template is used when the controller is running in `script` mode.
The other one is for the `cloud-init` mode.
Both should be the same semantically, but different syntactically(one is `bash` script, the other one is `cloud-init`).

controllers/os-suse-jeos/pkg/generator/templates/cloud-init/suse-jeos.template
```diff
@@ -28,6 +28,7 @@ write_files:
 runcmd:
 - "until zypper -q install -y docker wget socat jq nfs-client; [ $? -ne 7 ]; do sleep 1; done"
 - systemctl daemon-reload
+- ip link set dev eth0 mtu 1460
```
aojea: do this survive reboots? how can this change be permanent @rossella?

adding add/change MTU='1460' in /etc/sysconfig/network/ifcfg-eth0?

vpnachev: > do this survive reboots?

Gardener never reboot the VMs, but of course it is possible to happen. 
I've added a command to set the MTU in `/etc/sysconfig/network/ifcfg-eth0` file.

PR #503, opened 2019-12-15
Bumped OpenStack CCM to v1.17.0

afritzler:
Bumped OpenStack CCM to v1.17.0

**Release note**:
```improvement operator
Added support for OpenStack CCM v1.17.0
```


rfranzke:
Is the CCM backwards compatible with 1.15 and 1.16?

afritzler:
Yes according to the community this should be fine. The CCM version is actually not really tight to the Kubernetes version. Unfortunately they mostly release the new CCM versions together with new Kubernetes versions.

PR #505, opened 2019-12-17
Adapt extensions to be able to work with all versions of `core.gardener.cloud` API group

rfranzke:
Adapt extensions to be able to work with all versions of `core.gardener.cloud` API group

Required as of https://github.com/gardener/gardener/pull/1681

**Release note**:

```noteworthy operator
All extension controllers are now able to work with all versions of Gardener's `core.gardener.cloud` API group.
```


rfranzke:
The problem is that there are "cyclic" dependencies now for backwards-compatibility. g/g requires g/gext to convert the old API to the new one, and g/gext requires g/g for its core and extension APIs. Code-wise the cycle does not exist (luckily), so the program compiles and works as desired, but our check scripts, especially those for generating and checking don't like the situation... working on it.

rfranzke:
OK, everything fixed now. /ready-for-review @gardener/gardener-maintainers

PR #509, opened 2019-12-19
Add hack script for extensions and webhook local developement

zanetworker:
Currently testing webhooks and extensions end-to-end takes a lot of time due to building, code generation, linting, and so on.

This PR adds a hack script to allow for running webhooks and extensions locally through tunneling requests from the seed clusters to the provider extensions running on the local development machine.

**Release note**:

```noteworthy developer
Developers need to set the `ignoreResources` variable of the extension chart to `true` in their extension controller-registration.
```

```noteworthy developer
Developers need to point their local `kubeconfig` to a target development seed cluster.
```


Makefile
```diff
@@ -216,6 +217,86 @@ start-provider-packet:
 		--webhook-config-mode=url \
 		--webhook-config-url=$(WEBHOOK_CONFIG_URL)
 
+
+.PHONY: start-provider-aws-local
+start-provider-aws-local:
```
timuthy: Instead of introducing new targets you could also implement something like this:
```
.PHONY: start-provider-aws
WEBHOOK_PARAM := --webhook-config-url=$(WEBHOOK_CONFIG_URL)
ifeq ($(WEBHOOK_CONFIG_MODE), service)
  WEBHOOK_PARAM := --webhook-config-namespace=$(EXTENSION_NAMESPACE)
endif

start-provider-aws:
	@LEADER_ELECTION_NAMESPACE=garden GO111MODULE=on go run \
		-mod=vendor \
		-ldflags $(LD_FLAGS) \
		./controllers/provider-aws/cmd/gardener-extension-provider-aws \
		--config-file=./controllers/provider-aws/example/00-componentconfig.yaml \
		--ignore-operation-annotation=$(IGNORE_OPERATION_ANNOTATION) \
		--leader-election=$(LEADER_ELECTION) \
		--webhook-config-server-host=0.0.0.0 \
		--webhook-config-server-port=8443 \
		--webhook-config-mode=$(WEBHOOK_CONFIG_MODE) \
		$(WEBHOOK_PARAM)
```
WDYT?

hack/hook-me.sh
```diff
@@ -0,0 +1,260 @@
...
+  [[ -z $providerName ]] && echo "Please specify the provider name (aws,gcp,azure,..etc.)!" && exit 1
+
+  webhookServerPort=${3:-}
+  [[ -z $webhookServerPort ]] && echo "Please specify webhook server port" && exit 1
+
+
+  trap 'cleanUP $namespace' SIGINT SIGTERM
+
+  echo "[STEP 0] Ignore Updates for the webhook service and the deployment"
+  kubectl -n garden scale deploy gardener-resource-manager --replicas=0
```
timuthy: AFAIK, the `gardener-resource-manager` is scaled up again after some time by the `seed` control loop and reverts the changes you made to the seed's resources. Doesn't that break the `inlets` setup?

Furthermore, this `gardener-resource-manager` instance is used to deploy control plane components for shoot cluster. For example, the `extension-shoot-cert-service` makes the `gardener-resource-manager` take care about the following resources: https://github.com/gardener/gardener-extensions/tree/master/controllers/extension-shoot-cert-service/charts/internal/shoot-cert-management-seed/templates

This also breaks the end-to-end test scenario, doesn't it?

zanetworker: Depends on how you use it, my use-case so far is to reconcile and introduce changes and let the new (local extension) handle the reconciliation and see it reconcile successfully, of course there are other use-cases as you mention where you want to start reconciling a shoot from scratch using the local-extension. In that case, we should not switch-off the resource manager. 

Another solution is to modify the charts to make use of the managed resource annotation. Let's discuss what would most make sense here.

hack/hook-me.sh
```diff
@@ -0,0 +1,260 @@
+
+#!/usr/bin/env bash
+
+checkPrereqs() {
```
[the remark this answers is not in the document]
zanetworker: Yes which is already checked at the shell script entry-level (if namespace is not set the script would error). `EXTENSION_NAMESPACE` is at the level of the make file in that case only (in that case the check needs to happen in the makefile)

Makefile
zanetworker: Will try it out.

hack/hook-me.sh
```diff
@@ -0,0 +1,260 @@
...
+usage(){
+  echo "===================================PRE-REQs========================================="
+  echo "\`host\` commands for DNS"
+  echo "\`inlets\` command. For mac, simply use \`brew install inlets\`, for linux \`curl -sLS https://get.inlets.dev | sudo sh\`"
+  echo "===================================================================================="
+
+  echo ""
+
+  echo "========================================================USAGE======================================================================"
+  echo "> ./hack/hook-me.sh <extension namespace e.g. extension-provider-aws-fpr6w> <provider e.g., aws>  <webhookserver port e.g., 8443>"
```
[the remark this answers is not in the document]
zanetworker: Sure.

hack/hook-me.sh
timuthy: > Another solution is to modify the charts to make use of the managed resource ignore annotation. Let's discuss what would most make sense here.

As discussed, this may be the only viable solution when end-to-end testing is desired. It'd be nice to see a documentation explaining those prerequisites because it's not trivial to understand the rationale behind it.

rfranzke: Can't we simply expose a new Helm chart value that, if set to `true`, adds the ignore annotation to the service behind the webhook pod? Then, if you want to run the webhook locally, you just need to start this script and change the mentioned flag in the respective `ControllerRegistration`. When you are done testing/running the webhook locally then you can easily go back to the standard deployment by removing the mentioned flag from the `ControllerRegistration` again.
No need to scale down the GRM.

Makefile
```diff
@@ -22,7 +22,14 @@ LD_FLAGS                    := "-w -X github.com/gardener/gardener-extensions/pk
 VERIFY                      := true
 LEADER_ELECTION             := false
 IGNORE_OPERATION_ANNOTATION := true
+WEBHOOK_CONFIG_MODE 		:= service
```
rfranzke:
```suggestion
WEBHOOK_CONFIG_MODE 		    := service
```

Makefile
```diff
@@ -22,7 +22,14 @@ LD_FLAGS                    := "-w -X github.com/gardener/gardener-extensions/pk
...
 WEBHOOK_CONFIG_URL          := docker.for.mac.localhost
+EXTENSION_NAMESPACE 		:= "extension-provider-aws-xxxx"
```
rfranzke:
```suggestion
EXTENSION_NAMESPACE 	    	:= "extension-provider-aws-xxxx"
```

rfranzke: Why is `aws-xxx` hard-coded? I would leave this variable empty.

controllers/provider-gcp/charts/provider-gcp/templates/deployment.yaml
```diff
@@ -3,6 +3,10 @@ kind: Deployment
 metadata:
   name: {{ include "name" . }}
   namespace: {{ .Release.Namespace }}
+{{-  if .Values.ignoreResources }}
```
rfranzke: Can you add this to the `values.yaml`?

hack/hook-me.sh
```diff
@@ -0,0 +1,291 @@
...
+    do
+      sleep 2s
+    done
+}
+
+cleanUP() {
+   namespace=${1:-}
+   [[ -z $namespace ]] && echo "Please specify the extension namespace!" && exit 1
+
+   kubectl -n $namespace wait --for=delete svc/inlets-lb --timeout=60s
```
rfranzke: Shouldn't we also delete the `Pod`, and if possible, revert the changes in the `Service` objects?

rfranzke:
Rebase needed because of new year :)

Makefile
```diff
@@ -22,7 +22,15 @@ LD_FLAGS                    := "-w -X github.com/gardener/gardener-extensions/pk
 VERIFY                      := true
 LEADER_ELECTION             := false
 IGNORE_OPERATION_ANNOTATION := true
-WEBHOOK_CONFIG_URL          := docker.for.mac.localhost
+
+WEBHOOK_CONFIG_MODE			:= service
```
rfranzke: Can you please format this better?

rfranzke:
Can you rebase again? The CI problems should be fixed as of #534

rfranzke:
@timuthy can you please check this PR again?

PR #510, opened 2019-12-19
Added Github Docker build and push action

afritzler:
This PR adds a GitHub action which builds (and potentially tests) and pushes a Docker image on each commit into a users own package repository on GH. Here it doesn't matter if it is on master or a specific branch.

In case of a master branch push the `latest` tag is updated. For a branch commit the corresponding `sha` is used as a tag.

**Why?** 
This GH action can conveniently be used on forks and publishes an image into a users local package repository.

E.g. for a commit into a users fork branch `feature/added_docker_action` the resulting image will be located here:
```shell
docker pull docker.pkg.github.com/${USER/ORG}/gardener-extensions/gardener-extension-hyper:4ed66ee9bcd5871f71a7f0350b3eb334920d33f8
```

This allows an easy use of the `gardener-extension` in a forked testing environment as those org/user images can be used in your own deployment manifests.

**Release note**:
```improvement developer
Added GitHub Docker build and push action
```

### DIFF
--- a/.github/workflows/docker_push.yaml
+++ b/.github/workflows/docker_push.yaml
@@ -1,0 +1,72 @@
+name: Docker
+
+on:
+  push:
+    # Publish `master` as Docker `latest` image.
+    branches:
+      - master
+      - '**'
+
+    # Publish `v1.2.3` tags as releases.
+    tags:
+      - v*
+
+  # Run tests for any PRs.
+  pull_request:
+
+env:
+  IMAGE_NAME: gardener-extension-hyper
+
+jobs:
+  # Run tests.
+  # See also https://docs.docker.com/docker-hub/builds/automated-testing/
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Run tests
+        run: |
+          echo "test run goes here"
+
+  # Push image to GitHub Package Registry.
+  # See also https://docs.docker.com/docker-hub/builds/
+  push:
+    # Ensure test job passes before pushing image.
+    # needs: test
+
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Build image
+        run: docker build . --build-arg VERIFY=false --file Dockerfile --tag image
+
+      - name: Log into registry
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login docker.pkg.github.com -u ${{ github.actor }} --password-stdin
+
+      - name: Push image
+        run: |
+          IMAGE_ID=docker.pkg.github.com/${{ github.repository }}/$IMAGE_NAME
+
+          # Strip git ref prefix from version
+          VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
+
+          # Strip "v" prefix from tag name
+          [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
+
+          # Use Docker `latest` tag convention
+          if [ "$VERSION" == "master" ]; then
+            VERSION=latest
+          else
+            VERSION="${{ github.sha }}"
+          fi
+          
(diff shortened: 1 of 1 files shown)


ialidzhikov:
Each PR (also master and release branch head updates) are built and published to our public GCR. You can find them under `eu.gcr.io/gardener-project/gardener/gardener-extension-hyper:<version>-<commit-hash>`.
For example

```
docker pull eu.gcr.io/gardener-project/gardener/gardener-extension-hyper:v1.1.0-dev-76ed93f24a3147e16547825076ef40bb56265991
```

And the same applies for g/g.

afritzler:
This does not happen on forks though. And that is the main usecase of this PR.

afritzler:
In my local dev/testing environment I have now a gazillion manually pushed tags in my private dockerhub. Would be great if that happens automatically while developing a feature in my fork.

afritzler:
If there is no need for such a thing we can close this PR here. Just wanted to get a feel if ppl would use something like that.

vpnachev:
But the forks can setup this on their own?
Should it be enabled by default for any fork?

afritzler:
Just wanted to get your input on that. If not needed we can close this PR.

vpnachev:
Lets keep it open for some time to get more feedback.

Generally I agree - it is useful feature that will ease the development. We can add it as an optional step for local development setup, at least.

ialidzhikov:
Ok, I got your point. I guess we all have dockerhubs for development with many pushed tags. It could be useful to use this github action when you want to build your work and you don’t want to do it locally, or you have some limitations in the local env.

afritzler:
Another use case I can think of are people who want to integrate a new infra (or other bigger changes) and don't want to open a PR for all the repos and wait for our CI to publish dev builds (or want to keep their work in the fork until things are properly working).

rfranzke:
Generally a nice idea, but given the plan to split up this repo, eventually it won't have any Dockerfile anymore. So, this would be deleted/removed again. :(

afritzler:
I know that things will be disassembled - removing it or moving this stuff to a new repo is not an issue.

rfranzke:
Right. /lgtm then :) thank you!

afritzler:
Before merging we probably should consider this:

>  To avoid breaking projects that may depend on your packages, you cannot delete an entire public package or specific versions of a public package. 

(https://help.github.com/en/github/managing-packages-with-github-packages/deleting-a-package)

rfranzke:
@afritzler what do you suggest now? Should we merge it or wait for the split?

.github/workflows/docker_push.yaml
```diff
@@ -0,0 +1,72 @@
...
+  # Run tests for any PRs.
+  pull_request:
+
+env:
+  IMAGE_NAME: gardener-extension-hyper
+
+jobs:
+  # Run tests.
+  # See also https://docs.docker.com/docker-hub/builds/automated-testing/
+  test:
```
ialidzhikov: Currently the test step does nothing. Can we remove it?

.github/workflows/docker_push.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: Push image
+        run: |
+          IMAGE_ID=docker.pkg.github.com/${{ github.repository }}/$IMAGE_NAME
+
+          # Strip git ref prefix from version
+          VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
+
+          # Strip "v" prefix from tag name
+          [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
```
ialidzhikov: Do we need this line as we fallback to `latest` or the commit sha?

vpnachev:
> I don't think it is a hard requirement for g/gext repo. We could avoid in the upstream repo something that builds "packages" (that cannot be deleted?).
> It is useful for forks and I rather see it as a guide under /docs about how to setup docker_push with github actions to automate and improve the development process.

I completely share this PoV.

rfranzke:
+1, agreed as well. Then let's close the PR and rather focus on a documentation PR that explains to developers how such a build and push action can be achieved for their forks.

PR #520, opened 2019-12-30
Allow to disable AWS ECR access via infrastructure config

### LINKED ISSUE #238: Disable Default ECR Access
**What would you like to be added**:
Either means to switch off ECR access for the worker nodes or influence the instance profile for the worker nodes.

**Why is this needed**:
There are use cases such as CI/CD where the cluster creates and loads images to/from a registry, but is itself offered as multi-tenant app/service in one cluster. Now, running custom code is always dangerous and even so PSP and NP help to secure a cluster, as an additional security layer, it would make sense to not generally allow the worker nodes to access ECR which these apps/services utilise. However, Gardener generally enabled ECR (historic setting), which is a problem for these use cases:
```
"Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "ec2:DescribeInstances"
            ],
            "Resource": [
                "*"
            ]
        },
        {
            "Effect": "Allow",
            "Action": [
                "ecr:GetAuthorizationToken",
                "ecr:BatchCheckLayerAvailability",
                "ecr:GetDownloadUrlForLayer",
                "ecr:GetRepositoryPolicy",
                "ecr:DescribeRepositories",
                "ecr:ListImages",
                "ecr:BatchGetImage"
            ],
            "Resource": [
                "*"
            ]
        }
    ]
}
```
That made kind of sense in the early days of Gardener, since Gardener focuses on bring-your-own-account, but it is problematic for use cases such as the above or even where multi-tenancy is implemented by means of separate clusters within one account all using ECR, but wanting to isolate access.

**Implementation proposals**:
- We could add another "switch" to the shoot spec (deprecated from the start) and that, if set to true, enables ECR access. However, if omitted/not set, ECR access would be disabled. This way we could phase out this permission over time. Owners of use cases such as the above could then use clean `imagePullSecret`s to enforce the behaviour they want. This however would make it ugly for other stakeholders that also use ECR, but not in a multi-tenant way.
- Same as above, but not deprecated. Then the default can/should maybe rather be enabled.
- When extensibility is ready, it would be possible to extend the shoot spec with custom configuration that could be evaluated by a custom controller that would then make the necessary modifications, e.g. imprinting a custom instance profile.

rfranzke:
Added a new `enableECRAccess` flag to `aws.provider.extensions.gardener.cloud/v1alpha1.InfrastructureConfig` that allows to control whether the permissions shall be added to the AWS IAM role policy attached to all worker nodes.

Fixes #238

**Release note**:

```noteworthy user
It is now possible to remove the ECR permission that is part of the AWS IAM role policy attached to all shoot worker nodes by specifying `aws.provider.extensions.gardener.cloud/v1alpha1.InfrastructureConfig.enableECRAccess=false`. If the field is not provided then it is defaulted to `true`, preserving the old behaviour where ECR access is always granted.
```

### DIFF
--- a/controllers/provider-aws/charts/internal/aws-infra/templates/main.tf
+++ b/controllers/provider-aws/charts/internal/aws-infra/templates/main.tf
@@ -312,7 +312,7 @@ resource "aws_iam_role_policy" "nodes" {
       "Resource": [
         "*"
       ]
-    },
+    }{{ if .Values.enableECRAccess }},
     {
       "Effect": "Allow",
       "Action": [
@@ -327,7 +327,7 @@ resource "aws_iam_role_policy" "nodes" {
       "Resource": [
         "*"
       ]
-    }
+    }{{ end }}
   ]
 }
 EOF
--- a/controllers/provider-aws/charts/internal/aws-infra/values.yaml
+++ b/controllers/provider-aws/charts/internal/aws-infra/values.yaml
@@ -7,6 +7,7 @@ create:
 sshPublicKey: sshkey-12345
 
 clusterName: test-namespace
+enableECRAccess: true
 
 vpc:
   id: ${aws_vpc.vpc.id}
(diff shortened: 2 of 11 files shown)


rfranzke:
ping

PR #528, opened 2020-01-07
[WIP]add worker state retention

vlvasilev:
With this PR we add second controller which works side by side with the worker controller.|
This new controller watch for changes in the MachineSets and the Machines and update the state of the Worker state. The Worker State consist of the actual state of the MachineSets and the Machines.
In case of a control plane migration this resources will be extracted from the Worker State and deployed instead of letting MCM to create a new ones.
 
**Release note**:

```improvement operator
Add worker state updating controller
```

### DIFF
--- a/pkg/controller/worker/genericactuator/actuator_reconcile.go
+++ b/pkg/controller/worker/genericactuator/actuator_reconcile.go
@@ -22,7 +22,7 @@ import (
 	"github.com/gardener/gardener-extensions/pkg/controller"
 	extensionscontroller "github.com/gardener/gardener-extensions/pkg/controller"
 	workerhealthcheck "github.com/gardener/gardener-extensions/pkg/controller/healthcheck/worker"
-	"github.com/gardener/gardener-extensions/pkg/controller/worker"
+	workercontroller "github.com/gardener/gardener-extensions/pkg/controller/worker"
 	"github.com/gardener/gardener-extensions/pkg/util"
 
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
@@ -49,31 +49,22 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 		return errors.Wrapf(err, "could not instantiate actuator context")
 	}
 
-	// If the shoot is hibernated then we want to scale down the machine-controller-manager. However, we want to first allow it to delete
-	// all remaining worker nodes. Hence, we cannot set the replicas=0 here (otherwise it would be offline and not able to delete the nodes).
-	var replicaFunc = func() (int32, error) {
-		if extensionscontroller.IsHibernated(cluster) {
-			deployment := &appsv1.Deployment{}
-			if err := a.client.Get(ctx, kutil.Key(worker.Namespace, a.mcmName), deployment); err != nil && !apierrors.IsNotFound(err) {
-				return 0, err
-			}
-			if replicas := deployment.Spec.Replicas; replicas != nil {
-				return *replicas, nil
-			}
-		}
-		return 1, nil
-	}
-
-	// Deploy the machine-controller-manager into the cluster.
-	a.logger.Info("Deploying the machine-controller-manager", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
-	if err := a.deployMachineControllerManager(ctx, worker, cluster, workerDelegate, replicaFunc); err != nil {
-		return err
-	}
-
 	// Generate the desired machine deployments.
 	wantedMachineDeployments, err := workerDelegate.GenerateMachineDeployments(ctx)
 	if err != nil {
 		return errors.Wrapf(err, "failed to generate the machine deployments")
+	}
+
+	// Get the list of all existing machine deployments.
+	existingMachineDeployments := &machinev1alpha1.MachineDeploymentList{}
+	if err := a.client.List(ctx, existingMachineDeployments, client.InNamespace(worker.Namespace)); err != nil {
+		return err
+	}
+
+	// If we have control plane migration then we have to restore the machineSet and the machines
+	// saved in the Worker state
+	if err := a.restore(ctx, worker, existingMachineDeployments, wantedMachineDeployments); err != nil {
+		return err
 	}
 
 	// Get list of existing machine class names and list of used machine class secrets.
@@ -129,16 +120,32 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 		return errors.Wrapf(err, "failed to update the machine images in worker status")
 	}
 
-	// Get the list of all existing machine deployments.
-	existingMachineDeployments := &machinev1alpha1.MachineDeploymentList{}
-	if err := a.client.List(ctx, existingMachineDeployments, client.InNamespace(worker.Namespace)); err != nil {
-		return err
-	}
-
 	// Generate machine deployment configuration based on previously computed list of deployments and deploy them.
 	a.logger.Info("Deploying the machine deployments", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
 	if err := a.deployMachineDeployments(ctx, cluster, worker, existingMachineDeployments, wantedMachineDeployments, workerDelegate.MachineClassKind(), clusterAutoscalerUsed); err != nil {
 		return errors.Wrapf(err, "failed to generate the machine deployment config")
+	}
+
+	// If the shoot is hibernated then we want to scale down the machine-controller-manager. However, we want to first allow it to delete
+	// all remaining worker nodes. Hence, we cannot set the replicas=0 here (otherwise it would be offline and not able to delete the nodes).
+	var replicaFunc = func() (int32, error) {
+		if extensionscontroller.IsHibernated(cluster) {
+			deployment := &appsv1.Deployment{}
+			if err := a.client.Get(ctx, kutil.Key(worker.Namespace, a.mcmName), deployment); err != nil && !apierrors.IsNotFound(err) {
+				return 0, err
+			}
+			if replicas := deployment.Spec.Replicas; replicas != nil {
+				return *replicas, nil
+			}
+		}
+		return 1, nil
+	}
+
+	// Deploy the machine-controller-manager into the cluster.
+	// It is crucial the deployment of MCM to be done after restore, otherwise the behavior is undefined
+	a.logger.Info("Deploying the machine-controller-manager", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
+	if err := a.deployMachineControllerManager(ctx, worker, cluster, workerDelegate, replicaFunc); err != nil {
+		return err
 	}
 
 	// Wait until all generated machine deployments are healthy/available.
@@ -200,7 +207,7 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 	return nil
 }
 
-func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster *controller.Cluster, worker *extensionsv1alpha1.Worker, existingMachineDeployments *machinev1alpha1.MachineDeploymentList, wantedMachineDeployments worker.MachineDeployments, classKind string, clusterAutoscalerUsed bool) error {
+func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster *controller.Cluster, worker *extensionsv1alpha1.Worker, existingMachineDeployments *machinev1alpha1.MachineDeploymentList, wantedMachineDeployments workercontroller.MachineDeployments, classKind string, clusterAutoscalerUsed bool) error {
 	for _, deployment := range wantedMachineDeployments {
 		var (
 			labels                    = map[string]string{"name": deployment.Name}
@@ -217,9 +224,15 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster 
 		case !clusterAutoscalerUsed:
 			replicas = deployment.Minimum
 		// If the machine deployment does not yet exist we set replicas to min so that the cluster
-		// autoscaler can scale them as required.
+		// autoscaler can scale them as required. But if the machine deployment has some state records
+		// it means that we either have control plane migration or someone  intentionally has deleted
+		// the machine deployment
 		case existingMachineDeployment == nil:
-			replicas = deployment.Minimum
+			if deployment.State == nil {
+				replicas = deployment.Minimum
+			} else {
+				replicas = len(deployment.State.Machines)
+			}
 		// If the Shoot was hibernated and is now woken up we set replicas to min so that the cluster
 		// autoscaler can scale them as required.
 		case shootIsAwake(controller.IsHibernated(cluster), existingMachineDeployments):
@@ -296,7 +309,7 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster 
 
 // waitUntilMachineDeploymentsAvailable waits for a maximum of 30 minutes until all the desired <machineDeployments>
 // were marked as healthy/available by the machine-controller-manager. It polls the status every 5 seconds.
-func (a *genericActuator) waitUntilMachineDeploymentsAvailable(ctx context.Context, cluster *controller.Cluster, worker *extensionsv1alpha1.Worker, wantedMachineDeployments worker.MachineDeployments) error {
+func (a *genericActuator) waitUntilMachineDeploymentsAvailable(ctx context.Context, cluster *controller.Cluster, worker *extensionsv1alpha1.Worker, wantedMachineDeployments workercontroller.MachineDeployments) error {
 	return wait.PollUntil(5*time.Second, func() (bool, error) {
 		var numHealthyDeployments, numUpdated, numDesired, numberOfAwakeMachines int32
 
@@ -351,7 +364,7 @@ func (a *genericActuator) waitUntilMachineDeploymentsAvailable(ctx context.Conte
 	}, ctx.Done())
 }
 
-func (a *genericActuator) updateWorkerStatusMachineDeployments(ctx context.Context, worker *extensionsv1alpha1.Worker, machineDeployments worker.MachineDeployments) error {
+func (a *genericActuator) updateWorkerStatusMachineDeployments(ctx context.Context, worker *extensionsv1alpha1.Worker, machineDeployments workercontroller.MachineDeployments) error {
 	var statusMachineDeployments []extensionsv1alpha1.MachineDeployment
 
 	for _, machineDeployment := range machineDeployments {
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -26,6 +26,7 @@ import (
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	"github.com/gardener/gardener/pkg/utils"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 )
@@ -51,10 +52,18 @@ type MachineDeployment struct {
 	Labels         map[string]string
 	Annotations    map[string]string
 	Taints         []corev1.Taint
+	State          *MachineDeploymentState
 }
 
 // MachineDeployments is a list of machine deployments.
 type MachineDeployments []MachineDeployment
+
+// MachineDeploymentState stores the last versions of the machine sets and machine which
+// the machine deployment corresponds
+type MachineDeploymentState struct {
+	MachineSet *runtime.RawExtension  `json:"machineSet,omitempty"`
+	Machines   []runtime.RawExtension `json:"machines,omitempty"`
+}
 
 // HasDeployment checks whether the <name> is part of the <machineDeployments>
 // list, i.e. whether there is an entry whose 'Name' attribute matches <name>. It returns true or false.
--- a/pkg/controller/worker/state_actuator.go
+++ b/pkg/controller/worker/state_actuator.go
@@ -1,0 +1,197 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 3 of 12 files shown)


plkokanov:
/assign
I'll look at it a bit later today.

pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -192,10 +203,83 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 		return errors.Wrapf(err, "failed to update the machine deployments in worker status")
 	}
 
+	// if err := a.updateWorkerState(ctx, worker); err != nil {
```
plkokanov: If this is unnecessary remove the comment

pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -346,9 +436,10 @@ func (a *genericActuator) waitUntilMachineDeploymentsAvailable(ctx context.Conte
 	}, ctx.Done())
 }
 
-func (a *genericActuator) updateWorkerStatusMachineDeployments(ctx context.Context, worker *extensionsv1alpha1.Worker, machineDeployments worker.MachineDeployments) error {
+func (a *genericActuator) updateWorkerStatusMachineDeployments(ctx context.Context, worker *extensionsv1alpha1.Worker, machineDeployments workercontroller.MachineDeployments) error {
 	var statusMachineDeployments []extensionsv1alpha1.MachineDeployment
 
+	a.logger.V(4).Info("")
```
plkokanov: Empty log

pkg/controller/worker/machines.go
```diff
@@ -51,11 +52,19 @@ type MachineDeployment struct {
...
 	Annotations    map[string]string
 	Taints         []corev1.Taint
+	State          *MachineDeploymentState
 }
 
 // MachineDeployments is a list of machine deployments.
 type MachineDeployments []MachineDeployment
 
+// MachineDeploymentState stores the last versions of the machine sets and machine which
+// the machine ployment corresponds
```
plkokanov:
```suggestion
// the machine deployment corresponds
```

pkg/predicate/predicate.go
```diff
@@ -202,3 +205,197 @@ func AddTypePredicate(extensionType string, predicates []predicate.Predicate) []
...
+				return true
+			}
+		}
+		return false
+	}
+
+	return statusChanged(statusHasChanged)
+}
+
+// MachineSetStatusHasChanged is a predicate deciding wether the status of a MCM's MachineSet has been changed.
```
plkokanov:
```suggestion
// MachineSetStatusHasChanged is a predicate deciding whether the status of a MCM's MachineSet has been changed.
 ```

**Review by plkokanov — CHANGES_REQUESTED** [review]
I still have to play around with it locally and take a closer look.
In the meantime can you fix the failing checks during the build process?

File: pkg/predicate/predicate.go [hunk 5]
```diff
@@ -202,3 +205,197 @@ func AddTypePredicate(extensionType string, predicates []predicate.Predicate) []
...
+		}
+
+		//Check the LastOperation
+		oldLastOperation := oldStatus.LastOperation
+		newLastOperation := newStatus.LastOperation
+
+		if oldLastOperation.Type != newLastOperation.Type ||
+			oldLastOperation.State != newLastOperation.State ||
+			oldLastOperation.Description != oldLastOperation.Description ||
+			newLastOperation.LastUpdateTime.After(newLastOperation.LastUpdateTime.Time) {
```
plkokanov:
```suggestion
			newLastOperation.LastUpdateTime.After(oldLastOperation.LastUpdateTime.Time) {
```

You can use equality.Semantic.DeepEqual(oldLastOperation, newLastOperation)

pkg/predicate/predicate.go
```diff
@@ -202,3 +205,197 @@ func AddTypePredicate(extensionType string, predicates []predicate.Predicate) []
...
+			return true
+		}
+
+		//Check the Conditions
+		oldConditions := oldStatus.Conditions
+		newConditions := newStatus.Conditions
```
plkokanov: You can simply use equaility.Semantic.DeepEquals here, again to avoid having to change this code in case the NodeCondition definition changes.

pkg/controller/worker/machines.go
```diff
@@ -51,11 +52,19 @@ type MachineDeployment struct {
...
+	State          *MachineDeploymentState
 }
 
 // MachineDeployments is a list of machine deployments.
 type MachineDeployments []MachineDeployment
 
+// MachineDeploymentState stores the last versions of the machine sets and machine which
+// the machine deployment corresponds
+type MachineDeploymentState struct {
+	MachineSet *runtime.RawExtension  `json:"machine_set,omitempty"`
```
rfranzke:
```suggestion
	MachineSet *runtime.RawExtension  `json:"machineSet,omitempty"`
```

pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -195,7 +206,76 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
...
+			}
+			if err := a.client.Create(ctx, machine); err != nil && !apierrors.IsAlreadyExists(err) {
+				return err
+			}
+			if err := a.client.Status().Update(ctx, machine); err != nil {
+				return err
+			}
+		}
+	}
+
```
rfranzke: Don't we need to handle the owner reference UIDs here in the `restore` case? Or does MCM do this automatically itself correctly?

pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -195,7 +206,76 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
...
+		return nil
+	}
+	workerState := make(map[string]*workercontroller.MachineDeploymentState)
+
+	if err := json.Unmarshal(workerCopy.Status.State.Raw, &workerState); err != nil {
+		return err
+	}
+
+	for index, wantedMachineDeployment := range wantedMachineDeployments {
+		wantedMachineDeployments[index].State = workerState[wantedMachineDeployment.Name]
```
rfranzke: Is my understanding correct that the only reason for this function being here is to preserve the state of already existing `MachineDeployment`s?

pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -212,9 +292,15 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster
 		case !clusterAutoscalerUsed:
 			replicas = deployment.Minimum
 		// If the machine deployment does not yet exist we set replicas to min so that the cluster
-		// autoscaler can scale them as required.
+		// autoscaler can scale them as required. But if the machine deployment has some state records
+		// it means that we aither have controll plane migration or someone  intentionally has deleted
```
rfranzke:
```suggestion
		// it means that we either have control plane migration or someone intentionally has deleted
```

pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -212,9 +292,15 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster
...
+		// the machine deployment
 		case existingMachineDeployment == nil:
-			replicas = deployment.Minimum
+			if deployment.State == nil {
+				replicas = deployment.Minimum
+			} else {
+				replicas = len(deployment.State.Machines)
```
rfranzke: I don't understand this code - can you elaborate?

pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -195,7 +206,76 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
...
+	}
+
+	for index, wantedMachineDeployment := range wantedMachineDeployments {
+		wantedMachineDeployments[index].State = workerState[wantedMachineDeployment.Name]
+	}
+	return nil
+}
+
+func (a *genericActuator) restore(ctx context.Context, cluster *controller.Cluster, worker *extensionsv1alpha1.Worker, existingMachineDeployments *machinev1alpha1.MachineDeploymentList, wantedMachineDeployments workercontroller.MachineDeployments) error {
+	existingMachineDeploymentNameSet := make(map[string]bool)
```
rfranzke:
```suggestion
	existingMachineDeploymentNames := make(map[string]struct{})
```

pkg/predicate/predicate.go
```diff
@@ -202,3 +205,114 @@ func AddTypePredicate(extensionType string, predicates []predicate.Predicate) []
...
+
+		if !equality.Semantic.DeepEqual(oldStatus.LastOperation, newStatus.LastOperation) {
+			return true
+		}
+
+		//Check the Conditions
+		if !equality.Semantic.DeepEqual(oldStatus.Conditions, newStatus.Conditions) {
+			return true
+		}
+
```
rfranzke: Do we really need to react when the conditions change? What data in the machine status is dynamic that we have to keep for the migration? Isn't it only the `node` field?

pkg/predicate/predicate.go
```diff
@@ -202,3 +205,114 @@ func AddTypePredicate(extensionType string, predicates []predicate.Predicate) []
...
+		//Check the MachineSetConditions
+		if !equality.Semantic.DeepEqual(newStatus.Conditions, newStatus.Conditions) {
+			return true
+		}
+
+		//Check MachineSummary
+		if !equality.Semantic.DeepEqual(oldStatus.FailedMachines, newStatus.FailedMachines) {
+			return true
+		}
+
```
rfranzke: Same question like above - which data here is dynamic that we have to keep it for the migration?

pkg/controller/worker/state_reconciler.go
```diff
@@ -0,0 +1,140 @@
...
+		//Something which is not created does not need state update for the moment so we leave it for later
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	if err := r.updateStatusProcessing(r.ctx, worker, operationType, "Updateting the worker state"); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if err := r.actuator.Reconcile(r.ctx, worker); err != nil {
+		msg := "Error updateting worker state"
```
rfranzke:
```suggestion
		msg := "Error updating worker state"
```

pkg/controller/worker/state_reconciler.go
```diff
@@ -0,0 +1,140 @@
...
+		return reconcile.Result{}, err
+	}
+
+	operationType := gardencorev1beta1helper.ComputeOperationType(worker.ObjectMeta, worker.Status.LastOperation)
+	if operationType == gardencorev1beta1.LastOperationTypeCreate {
+		//Something which is not created does not need state update for the moment so we leave it for later
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	if err := r.updateStatusProcessing(r.ctx, worker, operationType, "Updateting the worker state"); err != nil {
```
rfranzke:
```suggestion
	if err := r.updateStatusProcessing(r.ctx, worker, operationType, "Updating the worker state"); err != nil {
```

rfranzke: But wait a sec, why do you need to update the `lastOperation` here in the first place? The `lastOperation` should only be touched by the "real"/main controller. The state updater should only update the state.

pkg/controller/worker/state_reconciler.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (r *stateReconciler) updateStatusError(ctx context.Context, err error, worker *extensionsv1alpha1.Worker, lastOperationType gardencorev1beta1.LastOperationType, description string) error {
+	return extensionscontroller.TryUpdateStatus(ctx, retry.DefaultBackoff, r.client, worker, func() error {
+		worker.Status.ObservedGeneration = worker.Generation
+		worker.Status.LastOperation, worker.Status.LastError = extensionscontroller.ReconcileError(lastOperationType, gardencorev1beta1helper.FormatLastErrDescription(fmt.Errorf("%s: %v", description, err)), 50, gardencorev1beta1helper.ExtractErrorCodes(err)...)
+		return nil
+	})
+}
+
+func (r *stateReconciler) updateStatusSuccess(ctx context.Context, worker *extensionsv1alpha1.Worker, lastOperationType gardencorev1beta1.LastOperationType, description string) error {
```
rfranzke: See above, I think we can remove all these update functions to prevent conflicts between the real and the state updater.

pkg/controller/worker/state_actuator.go
```diff
@@ -0,0 +1,154 @@
...
+	return nil
+}
+
+func (a *genericStateActuator) updateWorkerState(ctx context.Context, worker *extensionsv1alpha1.Worker) error {
+	state, err := a.getWorkerState(ctx, worker)
+	if err != nil {
+		return err
+	}
+
+	err = extensionscontroller.TryUpdateStatus(ctx, retry.DefaultBackoff, a.client, worker, func() error {
```
rfranzke:
```suggestion
	return extensionscontroller.TryUpdateStatus(ctx, retry.DefaultBackoff, a.client, worker, func() error {
```

pkg/controller/worker/state_actuator.go
```diff
@@ -0,0 +1,154 @@
...
+			Annotations: machineSet.Annotations,
+			Labels:      machineSet.Labels,
+		}
+		machineSet.OwnerReferences = nil
+
+		machineDeploymentState.MachineSet = &runtime.RawExtension{Object: machineSet}
+
+		currentMachines := machines[machineSet.Name]
+		if len(currentMachines) <= 0 {
+			continue
```
rfranzke: This check is not needed. If the length is zero then your `for` loop will simply do nothing. Actually, `continue`ing here is even a bug, or? The line `150` would be skipped which is undesired, or?

rfranzke (changes requested):
Nice PR, I put a few questions and suggestions inline.

pkg/controller/worker/genericactuator/actuator_reconcile.go
vlvasilev: At the moment MCM has a adoption mechanism and it assigns the right resource as owners.

rfranzke: Great, thanks for clarifying and nice implementation @hardikdr @prashanth26 @ggaurav10

pkg/controller/worker/genericactuator/actuator_reconcile.go
vlvasilev: This function is a part of restoration of the state of the desired `MachineDeployment`.
We do not restore state of something that already exist because this interferes the work of the MCM and gives out of date state.
The purpose of this function is to parse the Worker state to separate `MachineDeployment` states and assigns them to the appropriate `MachineDeployment`.
I know that this function can be optimised by setting the state to those `MachineDeployments` that does not exist but we have few `MachineDeployments` and we are assigning only pointers.

pkg/controller/worker/genericactuator/actuator_reconcile.go
vlvasilev: Case hit question: `existingMachineDeployment == nil`
Does the wanted deployment we are about to deploy already exists. 
If not, we have to deploy it. But how many replicas have to have this deployment?
If this deployment was deleted(not by the worker controller) or we are migrating to another seed then we have a certain number of replicas. So we have to set the exact number or we are risking to let MCM delete some of the existing `Machines`. 
Does the wantedDeployment has state entry?: `deployment.State == nil`
If no, this means that this deployment have never existed before and we use the old approach to set replicas to `deployment.Minimum`. The ClusterAutoScaler will take care of the rest.
But If  the wantedDeployment has state entry, this mean that this deployment was existing recently and it is either manual deleted or we are migrating the control plane. 
How much Machine does it have? ` len(deployment.State.Machines)`

pkg/controller/worker/state_reconciler.go
vlvasilev: I was wondering does this auxiliary controller has to left trace of his work in the Worker state and if so, where?

pkg/controller/worker/state_reconciler.go
vlvasilev: agree

pkg/controller/worker/state_actuator.go
vlvasilev: Do I have to save state of a `MachineDeployment` which has no machines?
How to tell json.Marshal(workerState) to ignore keys which have empty `MachineDeploymentState`?

pkg/predicate/predicate.go
vlvasilev: You are right, it is only the node which we have to keep.
What about the correctness of the Machines and MachineSet states.
Aren't they important for other components or we just leave this to the reconciliation done by MCM?
The predicates will be unusable by future functionalities which are going to depend of this changes in the state.
Do I have to react on changes int the MachineSet at all, then?

pkg/predicate/predicate.go
vlvasilev: Same comment as above

pkg/controller/worker/state_actuator.go
vlvasilev: I was using the error for some debug logs and forgot the reverse it. Thanks for pointing it.

pkg/controller/worker/machines.go
vlvasilev: ok

pkg/controller/worker/genericactuator/actuator_reconcile.go
vlvasilev: I forgot my head home when I wrote this comments

pkg/controller/worker/genericactuator/actuator_reconcile.go
vlvasilev: Didn't know that this is more cost efficient

pkg/controller/worker/state_reconciler.go
vlvasilev: I don't know which I need more, glasses or English lessons. Thanks

pkg/controller/worker/genericactuator/actuator_reconcile.go
rfranzke: Hm, I'm still confused, so this is already the code for restoring `MachineDeployment`s after a migration? Didn't the design doc tell that this is only triggered via a special "restore" annotation (cc @plkokanov)?

pkg/controller/worker/genericactuator/actuator_reconcile.go
rfranzke: Ah, okay, I see, thanks for the clarification, it makes sense after reading it again. :)

pkg/controller/worker/state_reconciler.go
rfranzke: Hm, what about sending `Events`?

pkg/controller/worker/state_actuator.go
rfranzke: Hm, okay, I understand. No, you don't have to save the state, of course, and if you have problems with `json.Marshal` then it's fine to `continue` here. Thanks.

pkg/predicate/predicate.go
rfranzke: Hm, good point, especially the `conditions` might trigger remediation if they are `False` for a too long time (and during migration the timestamp might be quite old). So, yes, if you "simply" apply the whole status again after migration to the new `Machine` object then it's probably safer to at least react on `condition` changes. 
Another option could be to only store `.status.node` (also motivated by the fact of reducing the stored data volume) because we know that this is the only field that is needed for the new MCM to adopt the existing node (@hardikdr @prashanth26 @ggaurav10 is this assumption correct)?

For `MachineSet`s, you are probably right and it does not make sense to store any status here because it's not needed for the adoption of the new MCM (again correct @hardikdr @prashanth26 @ggaurav10)?

pkg/controller/worker/genericactuator/actuator_reconcile.go
vlvasilev: I don't know what to say. Annotation operation is removed by the rapper we are using.
Unless we use some other kind of annotation I don't see how we can use the standard Reconcile flow and explicitly knowing that we are in migration.

pkg/controller/worker/state_reconciler.go
vlvasilev: This is good Idea. I will implement it.

pkg/controller/worker/genericactuator/actuator_reconcile.go
rfranzke: https://github.com/gardener/gardener/blob/master/docs/proposals/07-shoot-control-plane-migration.md states that `gardener.cloud/operation=restore` will be added in case Gardener wants the extensions to restore the state. It should not happen automatically "just because a machinedeployment is missing". This annotation should only be removed after the restoration succeeded. If the wrapper we are using removes this annotation then it should be adapted to onl remove the `gardener.cloud/operation=reconcile` and keep `gardener.cloud/operation=restore` until it can be safely removed.

plkokanov: Sorry for the late reply. Indeed it should be done as it is stated in the GEP. I’m still testing if the current implementation works in all cases, however we cannot guarantee that it will be safe in the future (if something in the nvm changes) . Also what worries me and what I’m testing right now is that we have both the worker controller and mcm creating machines/machinesets. With the restore annotation the logic will be much simpler.

plkokanov: An idea for the restore annotation: add a check for it to the OperationAnnotationWrapper.Reconcile and handle it just as the reconcile operation. Create a new class “inheriting” OperationAnnotationWrapper (which will be used by controllers perserving state) where the restore annotation is removed after successful reconciliation so that it can be used by the actuator to check whether to perform restoration.

pkg/predicate/predicate.go
hardikdr: Sorry for the delay, I missed this thread somehow.
Regarding the status of the Machine object, @plkokanov [if I dint forget] did few experiments and it worked fine keeping the only `status.node`,  MCM could reconcile the rest of the status-information. @plkokanov can you please confirm?

Regarding Machineset status, I guess it can also be safely reconciled and we don't need to copy it while migration. Although it would be nice if it could be tested once, I haven't explicitly tested this. 
cc @prashanth26

plkokanov: Yep, I experimented with keeping only the status.node field and no status at all for the machinesets. However I only did it with one node and after everything was working (although I doubt that it would be any different otherwise)

vlvasilev:
@plkokanov @vpnachev @swilen-iwanow @ialidzhikov @mvladev 
Please, make a review. Thanks in advance

plkokanov (changes requested):
Generally looks good, but I have some questions :).

PR #530, opened 2020-01-08
AWS validator requires a nodes CIDR in shoots

rfranzke:
After gardener/gardener#1782 the `nodes` CIDR in `.spec.networking` is optional to support more use-cases and infrastructures (see gardener/gardener#1582 for the motivation). The provider-specific validation webhooks now have to check that a `nodes` CIDR is given if it is required for their infrastructure.
This PR realizes this for AWS. Please note that the validation whether the given CIDR is correct and that it is disjoint with the other networks is still done by the Gardener API server itself.

Depends on gardener/gardener#1782 which should be merged first.

**Release note**:

```improvement operator
The AWS validation webhook now requires a specified `.spec.networking.nodes` field for AWS shotos.
```


rfranzke:
Rebasing and merging after the dependent https://github.com/gardener/gardener/pull/1782 is merged now.

PR #532, opened 2020-01-09
Azure-CCM: Bundle cloudProviderRateLimitQPS[Write] to the maxNodes

vpnachev:
Azure-CCM: Bundle cloudProviderRateLimitQPS[Write] to the maxNodes

Fixes #

https://github.com/kubernetes-sigs/cloud-provider-azure/issues/155#issuecomment-488523167

**Release note**:

```improvement operator
An issue where cloud controller manager was self-rate-limited for azure shoot clusters with more than 10 nodes has been fixed.  
```

### DIFF
--- a/controllers/provider-azure/charts/internal/cloud-provider-config/templates/cloud-provider-config.tpl
+++ b/controllers/provider-azure/charts/internal/cloud-provider-config/templates/cloud-provider-config.tpl
@@ -21,9 +21,9 @@ cloudProviderBackoffExponent: 1.5
 cloudProviderBackoffDuration: 5
 cloudProviderBackoffJitter: 1.0
 cloudProviderRateLimit: true
-cloudProviderRateLimitQPS: 10.0
+cloudProviderRateLimitQPS: {{ ( max .Values.maxNodes 10 ) }}
 cloudProviderRateLimitBucket: 100
-cloudProviderRateLimitQPSWrite: 10.0
+cloudProviderRateLimitQPSWrite: {{ ( max .Values.maxNodes 10 ) }}
 cloudProviderRateLimitBucketWrite: 100
 {{- if semverCompare ">= 1.14" .Values.kubernetesVersion }}
 cloudProviderBackoffMode: v2
--- a/controllers/provider-azure/pkg/controller/controlplane/valuesprovider.go
+++ b/controllers/provider-azure/pkg/controller/controlplane/valuesprovider.go
@@ -196,6 +196,11 @@ func getConfigChartValues(
 		return nil, errors.Wrapf(err, "could not determine subnet, availability set, route table or security group name from infrastructureStatus of controlplane '%s'", util.ObjectName(cp))
 	}
 
+	var maxNodes int32
+	for _, worker := range cluster.Shoot.Spec.Provider.Workers {
+		maxNodes = maxNodes + worker.Maximum
+	}
+
 	// Collect config chart values.
 	values := map[string]interface{}{
 		"kubernetesVersion": cluster.Shoot.Spec.Kubernetes.Version,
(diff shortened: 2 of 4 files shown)


controllers/provider-azure/charts/internal/cloud-provider-config/templates/cloud-provider-config.tpl
```diff
@@ -21,9 +21,9 @@ cloudProviderBackoffExponent: 1.5
 cloudProviderBackoffDuration: 5
 cloudProviderBackoffJitter: 1.0
 cloudProviderRateLimit: true
-cloudProviderRateLimitQPS: 10.0
+cloudProviderRateLimitQPS: {{ printf "%d%s" ( add .Values.maxNodes 10 ) ".0" }}
 cloudProviderRateLimitBucket: 100
-cloudProviderRateLimitQPSWrite: 10.0
+cloudProviderRateLimitQPSWrite: {{ printf "%d%s" ( add .Values.maxNodes 10 ) ".0" }}
```
ialidzhikov: Hmm, do you always add 10 to maxNode? Isn't better to use maxNodes if maxNodes >= 10, else - to use 10 as minimum.

controllers/provider-azure/pkg/controller/controlplane/valuesprovider.go
```diff
@@ -196,6 +197,13 @@ func getConfigChartValues(
 		return nil, errors.Wrapf(err, "could not determine subnet, availability set, route table or security group name from infrastructureStatus of controlplane '%s'", util.ObjectName(cp))
 	}
 
+	var maxNodesCount int32
+	maxNodesCount = 0
```
ialidzhikov: The default value of int32 is already 0. I guess there is no need from this statement.

controllers/provider-azure/pkg/controller/controlplane/valuesprovider.go
```diff
@@ -209,6 +217,7 @@ func getConfigChartValues(
 		"routeTableName":    routeTableName,
 		"securityGroupName": securityGroupName,
 		"region":            cp.Spec.Region,
+		"maxNodes":          maxNodes,
```
ialidzhikov:
```suggestion
		"maxNodes":          fmt.Sprintf("%d", maxNodesCount)
```

rfranzke: Why is `fmt.Sprintf` needed? Also, why do we need two variables `maxNodes` and `maxNodesCount`?

controllers/provider-azure/charts/internal/cloud-provider-config/templates/cloud-provider-config.tpl
ialidzhikov: See https://github.com/kubernetes-sigs/cloud-provider-azure/issues/155#issuecomment-488523167.

rfranzke: Nice reference 👍 Thanks, makes sense. I agree with @ialidzhikov suggestion (`max{10, maxNodes}`).

vpnachev: Done.
I also removed trailing `.0` as it seems it is not required. See this example [here](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/fb507c32b206a110f4c8bbf4316c7417101ce3ef/examples/azure.json#L23).

controllers/provider-azure/pkg/controller/controlplane/valuesprovider.go
vpnachev: Sorry, it was a dummy approach to fix unit tests failures.
They were failing because `int(0)!=int32(0)`, which I got too late.

PR #533, opened 2020-01-09
Allow Azure cloud-controller to create service events

### LINKED ISSUE #531: Azure cloud-controller-manager RBAC for serviceaccount "system:serviceaccount:kube-system:azure-cloud-provider"
Currently the azure cloud-controller-manager has multiple error logs :
```
'events "kube-apiserver.15e7ebd2126cb9f8" is forbidden: User "system:serviceaccount:kube-system:azure-cloud-provider" cannot patch resource "events" in API group "" in the namespace "shoot--foo--bar"' (will not retry!)
```

According to kubernetes/kubernetes#68212 the ClusterRole `system:azure-cloud-provider` and ClusterRoleBinding `system:azure-cloud-provider` needs to be created.

ialidzhikov:
Allow Azure cloud-controller to create Service events.
Azure Services can now have Events also from source `azure-cloud-provider` (in addition to the existing source `service-controller`).
For example:
```
$ k describe svc -n kube-system vpn-shoot
# ...
Events:
  Type     Reason                    Age                  From                  Message
  ----     ------                    ----                 ----                  -------
  Normal   EnsuringLoadBalancer      53m (x5 over 54m)    service-controller    Ensuring load balancer
  Warning  SyncLoadBalancerFailed    53m (x5 over 54m)    service-controller    Error syncing load balancer: failed to ensure load balancer: autorest/azure: Service returned an error. Status=<nil> Code="ScopeLocked" Message="The scope 'foo' cannot perform write operation because following scope(s) are locked: 'bar'. Please remove the lock and try again."
  Warning  CreateOrUpdateInterface   11m                  azure-cloud-provider  network.InterfacesClient#CreateOrUpdate: Failure sending request: StatusCode=429 -- Original Error: Code="RetryableError" Message="A retryable error occurred." Details=[{"code":"ReferencedResourceNotProvisioned","message":"Cannot proceed with operation because resource foo is not in Succeeded state. Resource is in Updating state and the last operation that updated/is updating the resource is InternalOperation."}]
  Warning  CreateOrUpdateInterface   11m                  azure-cloud-provider  network.InterfacesClient#CreateOrUpdate: Failure sending request: StatusCode=429 -- Original Error: Code="RetryableError" Message="A retryable error occurred." Details=[{"code":"ReferencedResourceNotProvisioned","message":"Cannot proceed with operation because resource foo is not in Succeeded state. Resource is in Updating state and the last operation that updated/is updating the resource is InternalOperation."}]
```

Fixes #531

**Release note**:

```improvement operator
Azure cloud-controller-manager is now able to create Service events which are helpful to easily identify the underground errors of Azure API.
```

### DIFF
--- a/controllers/provider-azure/charts/internal/cloud-controller-manager-shoot/templates/rbac-cloud-controller.yaml
+++ b/controllers/provider-azure/charts/internal/cloud-controller-manager-shoot/templates/rbac-cloud-controller.yaml
@@ -1,0 +1,27 @@
+{{- if semverCompare ">= 1.12" .Capabilities.KubeVersion.GitVersion }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:azure-cloud-provider
+rules:
+- apiGroups: [""]
+  resources: ["events"]
+  verbs:
+  - create
+  - patch
+  - update
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: system:azure-cloud-provider
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:azure-cloud-provider
+subjects:
+- kind: ServiceAccount
+  name: azure-cloud-provider
+  namespace: kube-system
+{{- end }}


rfranzke:
Hm, can you explain 

a) why we need to have a `ClusterRole` with the fixed name `system:azure-cloud-provider` (theoretically, we could also call it differently, or?)
b) why you bind this role to a `azure-cloud-provider` service account in the `kube-system` namespace? We don't use this service account for the CCM, instead, we [generate a client cert](https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-azure/pkg/controller/controlplane/valuesprovider.go#L59-L65) with `cluster-admin` privileges for it. Why is all of this needed?

ialidzhikov:
> a) why we need to have a ClusterRole with the fixed name system:azure-cloud-provider (theoretically, we could also call it differently, or?)

Yep, probably. We want to to bind the `azure-cloud-provider` ServiceAccount to a ClusterRole that allows creating events.

> b) why you bind this role to a azure-cloud-provider service account in the kube-system namespace?

Because the issue I am trying to resolve is stating
```
'events "kube-apiserver.15e7ebd2126cb9f8" is forbidden: User "system:serviceaccount:kube-system:azure-cloud-provider" cannot patch resource "events" in API group "" in the namespace "shoot--foo--bar"' (will not retry!)
```

and for me the `azure-cloud-provider` ServiceAccount is being used to patch the events. You think that the cloud-controller-manager should not use this ServiceAccount?
Isn't it hard-coded in the source with
https://github.com/kubernetes/kubernetes/blob/5f799a487b70aea5e298e5f5f1e3bac904b54ef6/pkg/cloudprovider/providers/azure/azure.go#L414-L419

> Why is all of this needed?

See https://github.com/kubernetes/kubernetes/pull/68212. To allow cloud-controller-manager to be able to create Service events which are helpful to easily identify the underground errors of Azure API.

rfranzke:
Thanks for the thorough explanations, but isn't this a bug in the Azure CCM? Why doesn't it use the kubeconfig that it got assigned to via command line flags?

Also, I tried to follow [this code](https://github.com/kubernetes/kubernetes/blob/5f799a487b70aea5e298e5f5f1e3bac904b54ef6/pkg/cloudprovider/providers/azure/azure.go#L415) to understand where the username `system:serviceaccount:kube-system:azure-cloud-provider` is coming from. Can you help out?

/lgtm anyways for now, thanks for the digging and the fix!

dkistner:
Hmm I'm also a little bit confused why they use hard coded the `azure-cloud-provider` service account instead of the passed kubeconfig, which should have the required privileges. As far as I can oversee this they use the sa only for the event broadcasting.

Should we open a ticket to clarify this? I would do this.
I think for now we should go on with this approach. Thanks!
/lgtm

rfranzke:
Yes, please open a ticket.

PR #545, opened 2020-01-14
React on operation annotation when value is "restore" and delete it before reconciliation

vlvasilev:
Whit this PR the predicate HasOperationAnnotation will return true when the operation is "restore" and the OperationAnnotationWrapper will remove the annotation before reconciliation.
The Wrapper will react on the "restore" the same way as on "reconcile". 

**Release note**:

```noteworthy developer
 The predicate HasOperationAnnotation will react on "resore" value returning true.
 The OperationAnnotationWrapper will remove operation annotation with value "restore" before reconciliation.
```

### DIFF
--- a/pkg/controller/reconciler.go
+++ b/pkg/controller/reconciler.go
@@ -68,19 +68,27 @@ func (o *operationAnnotationWrapper) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, err
 	}
 
-	acc, err := meta.Accessor(obj)
-	if err != nil {
+	if err := o.removeOperationAnnotation(obj, v1beta1constants.GardenerOperation, v1beta1constants.GardenerOperationReconcile, v1beta1constants.GardenerOperationRestore); err != nil {
 		return reconcile.Result{}, err
 	}
 
+	return o.Reconciler.Reconcile(request)
+}
+
+func (o *operationAnnotationWrapper) removeOperationAnnotation(obj runtime.Object, key string, values ...string) error {
+	acc, err := meta.Accessor(obj)
+	if err != nil {
+		return err
+	}
+
 	annotations := acc.GetAnnotations()
-	if annotations[v1beta1constants.GardenerOperation] == v1beta1constants.GardenerOperationReconcile {
-		withOpAnnotation := obj.DeepCopyObject()
-		delete(annotations, v1beta1constants.GardenerOperation)
-		acc.SetAnnotations(annotations)
-		if err := o.client.Patch(o.ctx, obj, client.MergeFrom(withOpAnnotation)); err != nil {
-			return reconcile.Result{}, err
+	for _, value := range values {
+		if annotations[key] == value {
+			objWithOpAnnotation := obj.DeepCopyObject()
+			delete(annotations, key)
+			acc.SetAnnotations(annotations)
+			return o.client.Patch(o.ctx, obj, client.MergeFrom(objWithOpAnnotation))
 		}
 	}
-	return o.Reconciler.Reconcile(request)
+	return nil
 }
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -158,7 +158,8 @@ func HasName(name string) predicate.Predicate {
 // HasOperationAnnotation is a predicate for the operation annotation.
 func HasOperationAnnotation() predicate.Predicate {
 	return FromMapper(MapperFunc(func(e event.GenericEvent) bool {
-		return e.Meta.GetAnnotations()[v1beta1constants.GardenerOperation] == v1beta1constants.GardenerOperationReconcile
+		return e.Meta.GetAnnotations()[v1beta1constants.GardenerOperation] == v1beta1constants.GardenerOperationReconcile ||
+			e.Meta.GetAnnotations()[v1beta1constants.GardenerOperation] == v1beta1constants.GardenerOperationRestore
 	}), CreateTrigger, UpdateNewTrigger, GenericTrigger)
 }
 


pkg/controller/reconciler.go
```diff
@@ -68,19 +68,36 @@ func (o *operationAnnotationWrapper) Reconcile(request reconcile.Request) (recon
...
 	}
 
+	if ok, err := o.removeOperationAnnotation(obj, v1beta1constants.GardenerOperation, v1beta1constants.GardenerOperationReconcile); !ok {
+		return reconcile.Result{}, err
+	}
+
+	result, err := o.Reconciler.Reconcile(request)
+
+	if err == nil {
+		if ok, err := o.removeOperationAnnotation(obj, v1beta1constants.GardenerOperation, v1beta1constants.GardenerOperationMigrate); !ok {
```
timuthy: I think removing the `migrate` annotation after the reconciliation doesn't work with the current implementation. Let's say you start the reconciliation (`o.Reconciler.Reconcile(request)`) which also updates the resource a bunch of times. While sending the updates, the resource still carries the `reconcile` operation. As a consequence, the triggered update enqueues the resource again because `HasOperationAnnotation()` returns true, i.e. the resource is infinitely re-queued.

Have you tested this scenario?

timuthy (changes requested):

pkg/controller/reconciler.go
plkokanov: Generally for the OperationAnnotationWrapper, which is used by all controllers, the `restore` annotation should be handled exactly the same as the `reconcile` annotation

pkg/predicate/predicate.go
```diff
@@ -158,7 +158,8 @@ func HasName(name string) predicate.Predicate {
 // HasOperationAnnotation is a predicate for the operation annotation.
 func HasOperationAnnotation() predicate.Predicate {
 	return FromMapper(MapperFunc(func(e event.GenericEvent) bool {
-		return e.Meta.GetAnnotations()[v1beta1constants.GardenerOperation] == v1beta1constants.GardenerOperationReconcile
+		return e.Meta.GetAnnotations()[v1beta1constants.GardenerOperation] == v1beta1constants.GardenerOperationReconcile ||
+			e.Meta.GetAnnotations()[v1beta1constants.GardenerOperation] == v1beta1constants.GardenerOperationMigrate
```
plkokanov: As we talked: `GardenerOperationMigrate` should be `GardenerOperationRestore`

timuthy: > Generally for the OperationAnnotationWrapper, which is used by all controllers, the restore annotation should be handled exactly the same as the reconcile annotation

Why using `restore` at all then?

plkokanov: Meant to say "all controllers which do not have to restore their state"

pkg/controller/reconciler.go
timuthy: > Meant to say "all controllers which do not have to restore their state". Basically during migration Gardener doesn't know whether an extension controller has to restore its state or simply perform reconciliation on the new seed, so it sets operation=restore on all extension resources.
Extension controllers which do not have to restore their state, simply have to remove the annotation and perform reconciliation. Whereas controllers which have to restore their state need to trigger restoration instead of reconciliation.

So what will be done is that a new struct similar to OperationAnnotationWrapper will be created which should handle the restore and reconcile operations differently.

Sorry I commented the wrong conversation.

Okay, makes sense what you say with an additional wrapping functionality.

vlvasilev: We shall move the Restore handling in another Wrapper.

pkg/predicate/predicate.go
vlvasilev: Migration is replaced by Restore

pkg/controller/reconciler.go
```diff
@@ -68,19 +68,35 @@ func (o *operationAnnotationWrapper) Reconcile(request reconcile.Request) (recon
...
 	}
 
+	if ok, err := o.removeOperationAnnotation(obj, v1beta1constants.GardenerOperation, v1beta1constants.GardenerOperationReconcile, v1beta1constants.GardenerOperationRestore); !ok {
+		return reconcile.Result{}, err
+	}
+
+	return o.Reconciler.Reconcile(request)
+}
+
+func (o *operationAnnotationWrapper) removeOperationAnnotation(obj runtime.Object, key string, values ...string) (bool, error) {
```
plkokanov: Not sure that it is necessary to return (bool, err) here. Anytime err != nil is returned, false is also returned, and when err == nil, trrue is returned. You could just reuse the err.

pkg/controller/reconciler.go
```diff
@@ -68,19 +68,35 @@ func (o *operationAnnotationWrapper) Reconcile(request reconcile.Request) (recon
...
-		delete(annotations, v1beta1constants.GardenerOperation)
-		acc.SetAnnotations(annotations)
-		if err := o.client.Patch(o.ctx, obj, client.MergeFrom(withOpAnnotation)); err != nil {
-			return reconcile.Result{}, err
+	objWithOpAnnotation := obj.DeepCopyObject()
+	for _, value := range values {
+		if annotations[key] == value {
+			delete(annotations, key)
+			acc.SetAnnotations(annotations)
+			opAnnotationRemoved = true
```
plkokanov: nit: you can return from the loop here: 
move the patch command here and then you can get rid of `opAnnotationRemoved`

plkokanov (changes requested):

pkg/controller/reconciler.go
vlvasilev: You are right.

pkg/controller/reconciler.go
vlvasilev: You are right. Thanks

plkokanov: The function still returns err and bool :)

plkokanov (changes requested):

pkg/controller/reconciler.go
vlvasilev: Not enymore

pkg/controller/reconciler.go
```diff
@@ -68,19 +68,31 @@ func (o *operationAnnotationWrapper) Reconcile(request reconcile.Request) (recon
...
-		delete(annotations, v1beta1constants.GardenerOperation)
-		acc.SetAnnotations(annotations)
-		if err := o.client.Patch(o.ctx, obj, client.MergeFrom(withOpAnnotation)); err != nil {
-			return reconcile.Result{}, err
+	for _, value := range values {
+		if annotations[key] == value {
+			objWithOpAnnotation := obj.DeepCopyObject()
+			delete(annotations, key)
+			acc.SetAnnotations(annotations)
+			if err := o.client.Patch(o.ctx, obj, client.MergeFrom(objWithOpAnnotation)); err != nil {
```
plkokanov:
```suggestion
			return o.client.Patch(o.ctx, obj, client.MergeFrom(objWithOpAnnotation))
```

plkokanov (changes requested):

pkg/controller/reconciler.go
vlvasilev: thanks

plkokanov:
@timuthy @rfranzke could you have another look? 

The additional wrapper which will handle `restore` annotations separately from `reconcile` will be added with https://github.com/gardener/gardener-extensions/pull/528

timuthy:
> The additional wrapper which will handle restore annotations separately from reconcile will be added with #528

To be honst, I don't get the motivation for this PR if you at the same time change the behavior for `restore` in another PR. I would either see this additional wrapper in this PR or the introduction of `restore` as part of #528. WDYT?

plkokanov:
The idea was that in a second PR we will add a `RestoreAnnotationWrapper` which will handle restore differently and replace the OperationAnnotationWrapper for the worker controller.  The OperationAnnotationWrapper implementation would remain as it is in this PR.

But tbh I don't care that much whether there are 2 PRs or just #528

pkg/controller/reconciler.go
```diff
@@ -68,19 +68,27 @@ func (o *operationAnnotationWrapper) Reconcile(request reconcile.Request) (recon
...
 	}
 
+	if err := o.removeOperationAnnotation(obj, v1beta1constants.GardenerOperation, v1beta1constants.GardenerOperationReconcile, v1beta1constants.GardenerOperationRestore); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	return o.Reconciler.Reconcile(request)
+}
+
+func (o *operationAnnotationWrapper) removeOperationAnnotation(obj runtime.Object, key string, values ...string) error {
```
timuthy: If you keep this functionality in #528, I would design the function more generally:

```
func RemoveAnnotationContainingValues(ctx context.Context, client client.Client, obj runtime.Object, key string, values ...string) error
```

PR #546, opened 2020-01-14
fixing healthcheck wrong shoot client injection

danielfoehrKn:

Wrong shoot client injection in the healthcheck library causes.

Fixes #

**Release note**:

```improvement operator

```


pkg/controller/healthcheck/actuator.go
```diff
@@ -92,6 +92,8 @@ type HealthCheck interface {
 	InjectShootClient(client.Client)
 	// SetLoggerSuffix injects the logger
 	SetLoggerSuffix(string, string)
+	// Clone clones the healthCheck
+	Clone() HealthCheck
```
rfranzke: Call it DeepCopy like in the apimachinery? :)

PR #552, opened 2020-01-15
Use  instead of  until old API group is removed

rfranzke:
Until the deprecated `garden.sapcloud.io` API group is removed we have to keep using the old `worker` field instead of `workers`, otherwise Gardener API server will reject:

```
time="2020-01-14T16:18:16Z" level=debug msg="unable to create shoot tm-v39-ujw: Shoot.core.gardener.cloud \"tm-v39-ujw\" is invalid: spec.cloud.alicloud.networks.workers[0]: Invalid value: \"\": invalid CIDR address: "
```

/cc @schrodit 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/controllers/provider-alicloud/test/tm/generator.go
+++ b/controllers/provider-alicloud/test/tm/generator.go
@@ -59,8 +59,9 @@ func main() {
 			},
 			Zones: []v1alpha1.Zone{
 				{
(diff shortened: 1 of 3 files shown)


schrodit (approved):
lgtm
tested with all cloudproviders

PR #554, opened 2020-01-15
Update golangci/golangci-lint to v1.20.1

ialidzhikov:
Currently we experience high memory usage for golangci-lint runs which most often breaks the ci/cd run.
With the current master branch and a local container with 5GB limit the `golangci-lint` is getting OOMKilled trying to consume almost all of the memory.

Several improvements:
- Update to golangci/golangci-lint to v1.20.1 which according to release notes has improved memory usage (ref https://github.com/golangci/golangci-lint/releases/tag/v1.20.0). (We don't update to the latest one because of the flakes we hit with it - ref #534).
- Disable the `unused` linter.

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,6 @@
+run:
+  deadline: 7m
+
+linters:
+  disable:
+  - unused
--- a/.dockerignore
+++ b/.dockerignore
@@ -10,6 +10,7 @@
 !VERSION
 !Makefile
 !.gitignore
+!.golangci.yaml
 !go.mod
 !go.sum
 !test/
--- a/hack/check.sh
+++ b/hack/check.sh
@@ -24,7 +24,7 @@ echo "Executing check-generate"
 "$DIRNAME"/check-generate.sh
 
 echo "Executing golangci-lint"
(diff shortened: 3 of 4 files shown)


.golangci.yaml
```diff
@@ -0,0 +1,6 @@
+run:
+  deadline: 7m
+
+linters:
+  disable:
+  - unused
```
rfranzke: why is this disabled?

ialidzhikov: With `unused` linter enabled, in a local container with `4.8GiB` memory limit at some point of `golangci-lint` execution it gets all memory used (at least this is output from `docker stats`) and the `golangci-lint` process gets killed with exit code 137.
With `unused` linter disabled, the memory usage is much better and the execution is ok.

rfranzke: Does this mean we won't find any unused variables anymore?

ialidzhikov: Variable which is declared but not used should be compilation error. 
The linter doc
```
unused (megacheck): Checks Go code for unused constants, variables, functions and types [fast: false, auto-fix: false]
```

PR #558, opened 2020-01-17
Cleanup Alicloud SLBs before destroying infrastructure

### LINKED ISSUE #539: SLBs are left after some round of e2e test in Alicloud
We found issues that SLBs (Service load balancer) are left after belonging shoot cluster is deleted. This often ocurrs when e2e are run in the cluster.

EmoinLanyu:
Currently, Alicloud CCM does not use finalizer to ensure that SLBs are deleted when deleting services. In general cases, there is no problem (before we delete CCM, we will wait for some time until all the resources, SLBs not included though, are deleted). But from time to time, during e2e tests, we will see some left-over SLBs. This might be caused by CCM not being able to cleanup the resources in time before CCM itself is destroyed.

Provider-aws has similar logic. And of course, the best way to do this would be making the CCM itself use finalizer. However, to do this we have to pull PR in alicloud-cloud-controller-manager repo (or ask Alicloud people to do it). So we implement this workaround instead as a quick fix to avoid having to manually clean up these nameless SLBs (it is such a pain because currently it is NOT possible to identify where they come from, and you don't want to accidentally delete a SLB someone is using).

Since Alicloud CCM also doesn't support customizable naming of the SLB or tagging, we use an alternative to change the name of the VServer which SLB is associated with, and use it to identify whether the SLB is related to the Shoot.

Fixes #539 

**Release note**:

```improvement operator

```


EmoinLanyu:
> Code-wise it /lgtm, thanks for the PR. I haven't tested it, but I assume you have already done, of course. :)

Yes, I tested it in our local environment and it works fine. It was kind of tricky because we can't reproduce the issue in our local landscapes since we can't run e2e tests here. So I tried to delete CCM the moment when control plane is deployed & woken up again during deletion flow and let our infrastructure controller to do the job :-).

rfranzke:
Can you please edit the release note and explain what is coming in with this PR?

PR #559, opened 2020-01-17
Remove support for public IPs in GCP

### LINKED ISSUE #408: Remove external IPs from GCP worker nodes
With https://github.com/gardener/gardener-extensions/pull/379 and https://github.com/gardener/gardener-extensions/pull/398 Cloud NATs have been introduced for GCP shoots in order to remove the external IPs from the shoot worker nodes.

However, these changes were reverted with https://github.com/gardener/gardener-extensions/pull/405 because of instabilities.
More specifically, the main problem was that shoots which are deployed into the same VPC/network get one router and cloud NAT each. However, GCP has a hard quota limit here:

From https://cloud.google.com/router/quotas:

> Cloud Routers per project    Quotas    Regardless of quota, each network is limited to five Cloud Routers per region. See limits below.
The following limits for Cloud Router apply to VPC networks. Unless otherwise stated, these limits cannot be increased.

Hence, the implementation with #379 and #398 does not work.

What can we do to circumvent the problem and to get rid of the external IPs for GCP worker nodes?

zanetworker:
This PR completely removes support for public IPs in GCP. All egress traffic should be routed via CloudNAT. 

Fixes #408 

**Release note**:

```noteworthy user
Support for public IPs is now completely removed and new machines will get no public IPs.
```
```action user
It is now mandatory to specify a CloudRouter configuration for GCP shoots (i.e., name) if a VPC is re-used. The Gardener landscape operator should ensure that this is done for all GCP shoots before updating to this version.
```

### DIFF
--- a/controllers/provider-gcp/pkg/controller/worker/machines.go
+++ b/controllers/provider-gcp/pkg/controller/worker/machines.go
@@ -133,10 +133,6 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 		}
 
 		for zoneIndex, zone := range pool.Zones {
-			var disableExternalIP = false
-			if infrastructureStatus.Networks.VPC.CloudRouter != nil {
-				disableExternalIP = true
-			}
 			machineClassSpec := map[string]interface{}{
 				"region":             w.worker.Spec.Region,
 				"zone":               zone,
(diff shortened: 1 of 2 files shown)


vpnachev:
@zanetworker please open another PR and adapt the documentation here https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-gcp/docs/usage-as-end-user.md

PR #563, opened 2020-01-17
adding integration test healthchecks for all providers

### LINKED ISSUE #466: Integration Tests for Health Checks
For the health check controllers [check controllers](https://github.com/gardener/gardener-extensions/issues/406) we need integration tests for each extension controller.

This issue is to track the progress.

- [x] New Test Framework according to [Issue](https://github.com/gardener/gardener/issues/1565)
- [ ] Integration Tests for each provider using the Test framework

danielfoehrKn:

Implements the integration tests for the healthcheck for all extensions using the new [gardener test framework ](https://github.com/gardener/gardener/pull/1802).

Do not merge before  before [PR](https://github.com/gardener/gardener/pull/1802) is merged and go.mod points to the commit in the gardener.

Fixes #466 

The package test/integration/operation contains code commonly used by all health check integration test. It should be moved to the gardener repository or test framework repository before splitting up the gardener/gardener-extensions repo into provider specific repos. 

**Release note**:

```improvement operator
 added integration tests for health checks for all extensions.
```


danielfoehrKn:
Fixing CI issues when PR in Gardener is merged and I updated the go.mod

controllers/provider-aws/test/integration/healthcheck/test.go
```diff
@@ -0,0 +1,142 @@
...
+	timeout               = 10 * time.Minute
+	nodeRecreationTimeout = 20 * time.Minute
+	setupContextTimeout   = 2 * time.Minute
+)
+
+var _ = ginkgo.Describe("Provider-aws integration test: health checks", func() {
+	f := createShootFramework()
+
+	ginkgo.Context("Setup Check", func() {
+		f.Serial().CIt(fmt.Sprintf("Should be running on '%s'", aws.Type), func(ctx context.Context) {
```
schrodit: this test not do anything expect for printing out the provider.
I think we should check the shoots provider before every test and skip if the types do not match?

I suggest to add a `BeforeEach` that checks whether the shoot is applicable or not and skip the test if necessary.

test/integration/healthcheck/healthcheck_operation.go
```diff
@@ -0,0 +1,185 @@
...
+		healthConditionType)
+}
+
+func testHealthCheckWithManagedResource(ctx context.Context, timeout time.Duration, f *framework.ShootFramework, extensionKind schema.GroupVersionKind, extensionName string, managedResourceName string, healthConditionType gardencorev1beta1.ConditionType) error {
+	var (
+		err                                              error
+		resourceManagerDeploymentReplicasBeforeScaledown *int32
+	)
+	zero := int32(0)
+	resourceManagerDeploymentReplicasBeforeScaledown, err = operation.ScaleGardenerResourceManager(timeout, f.ShootSeedNamespace(), f.SeedClient.Client(), &zero)
```
schrodit:
```suggestion
	resourceManagerDeploymentReplicasBeforeScaledown, err = operation.ScaleGardenerResourceManager(timeout, f.ShootSeedNamespace(), f.SeedClient.Client(), pointer.Int32Ptr(0))
```

test/integration/healthcheck/healthcheck_operation.go
```diff
@@ -0,0 +1,185 @@
...
+	if len(machineList.Items) == 0 {
+		return fmt.Errorf("trying to delete machine as part of health check test from a cluster with no nodes (seed: %s, namespace: %s)", f.Seed.Name, f.ShootSeedNamespace())
+	}
+
+	machine := machineList.Items[0]
+	err = f.SeedClient.Client().Delete(ctx, &machine)
+	if err != nil {
+		return err
+	}
+	defer func() {
```
schrodit: the error handling is mixed between returning an error and panic on error (ginkgo Expect no error).
I think we should stay with one implementation so returning an error at any time? wdyt?

rfranzke:
Can you please rebase now that gardener/gardener#1802 is merged?

test/integration/healthcheck/healthcheck_operation.go
danielfoehrKn: The reason is, that the cleanup functionality is in the defer function to make sur eit is being executed.   I could either return a []error that includes potential cleanup errors, or let it ginko fail. I think it is appropriate to let the test suite fail of the cleanup is not successful. IDK what kind of error handling the caller of this function could do if the cleanup failed / the test def. should not continue.

controllers/provider-aws/test/integration/healthcheck/test.go
danielfoehrKn: Why would you have that in a foreach? As far as I understand I am acting on the same shoot, retrieved by the Framework (that I created before the execution) - so only one preliminary check is necessary.
In the long-run I would like to see that kind of functionality in a common section for all providers e.g in the gardener - so that each extension-provider does not have to implement that check anymore.
WDYT

danielfoehrKn:
PTAL again

rfranzke:
@danielfoehrKn please rebase

controllers/provider-aws/test/integration/healthcheck/test.go
schrodit: I like the idea of this being part of the framework to verify such things there 👍 

Yes, you are only acting the one shoot created beforehand.
My point here is that this only gives an info that the tests are skipped (or should be skipped).
But it does not actually skip the tests if the shoot is of a different type, or am I missing something?
So for example if I would run this test on a GCP shoot, it would print the message that the provider is not aws but still execute all the tests.

test/integration/healthcheck/healthcheck_operation.go
schrodit: If I understand the test correctly:
The machine is deleted and the test waits for the worker extensions to be unhealthy.
Then the defer function will be called that validates that the mcm correctly reconciles the deleted machine and the worker extensions gets healthy again.

Wouldn't it be sufficient to check the successful healthiness after the unsuccessful. 
Why do we want to check if the worker is healthy when the unhealthy checks fails?

So just:
```go

err := operation.WaitForExtensionCondition(ctx, f.Logger, f.SeedClient.Client(), extensionsv1alpha1.SchemeGroupVersion.WithKind(extensionsv1alpha1.WorkerResource), types.NamespacedName{
	Namespace: f.ShootSeedNamespace(),
	Name:      f.Shoot.GetName(),
}, gardencorev1beta1.ShootEveryNodeReady, gardencorev1beta1.ConditionFalse, healthcheck.HealthCheckUnsuccessful)
framework.ExpectNoError(err) // alternatively return err

err := operation.WaitForExtensionCondition(ctx, f.Logger, f.SeedClient.Client(), extensionsv1alpha1.SchemeGroupVersion.WithKind(extensionsv1alpha1.WorkerResource), types.NamespacedName{
	Namespace: f.ShootSeedNamespace(),
	Name:      f.Shoot.GetName(),
}, gardencorev1beta1.ShootEveryNodeReady, gardencorev1beta1.ConditionTrue, healthcheck.HealthCheckSuccessful)
framework.ExpectNoError(err) // alternatively return err
```

controllers/provider-aws/test/integration/healthcheck/test.go
danielfoehrKn: As discussed: removing provider checks as this repo is being split so that each provider has its own repository anyways.

test/integration/healthcheck/healthcheck_operation.go
danielfoehrKn: Thanks!

schrodit (dismissed):

test/integration/healthcheck/healthcheck_operation.go
```diff
@@ -0,0 +1,186 @@
+package healthcheck
```
rfranzke: license header

test/integration/operation/shoot_operation.go
```diff
@@ -0,0 +1,84 @@
+package operation
```
rfranzke: license header

rfranzke (changes requested):

rfranzke (approved):

PR #568, opened 2020-01-22
Add missing resources for Calico components

zanetworker:
This PR adds resources to Calico typha, kube-controllers, and cvpa. 

**Release note**:

```improvement operator
None
```

### DIFF
--- a/controllers/networking-calico/charts/internal/calico/templates/typha-cpva/deployment-calico-typha-vertical-autoscaler.yaml
+++ b/controllers/networking-calico/charts/internal/calico/templates/typha-cpva/deployment-calico-typha-vertical-autoscaler.yaml
@@ -34,6 +34,13 @@ spec:
             - --poll-period-seconds=30
             - --v=2
             - --config-file=/etc/config/typha-autoscaler
+          resources:
+            requests:
+              cpu: 10m
+              memory: 50Mi
+            limits:
+              cpu: 10m
+              memory: 50Mi
           volumeMounts:
             - name: config
               mountPath: /etc/config
--- a/controllers/networking-calico/charts/internal/calico/templates/kube-controller/deployment-calico-kube-controllers.yaml
+++ b/controllers/networking-calico/charts/internal/calico/templates/kube-controller/deployment-calico-kube-controllers.yaml
@@ -50,6 +50,13 @@ spec:
               value: node
             - name: DATASTORE_TYPE
               value: kubernetes
+          resources:
+            requests:
+              cpu: 10m
+              memory: 50Mi
+            limits:
+              cpu: 50m
+              memory: 100Mi
           readinessProbe:
             exec:
               command:
--- a/controllers/networking-calico/charts/internal/calico/templates/typha-cpha/deployment-calico-typha-horizontal-autoscaler.yaml
+++ b/controllers/networking-calico/charts/internal/calico/templates/typha-cpha/deployment-calico-typha-horizontal-autoscaler.yaml
@@ -38,7 +38,9 @@ spec:
           resources:
             requests:
               cpu: 10m
+              memory: 50Mi
             limits:
               cpu: 10m
+              memory: 50Mi
       serviceAccountName: typha-cpha
 {{- end }}
--- a/controllers/networking-calico/charts/internal/calico/templates/typha/deployment-calico-typha.yaml
+++ b/controllers/networking-calico/charts/internal/calico/templates/typha/deployment-calico-typha.yaml
@@ -92,6 +92,16 @@ spec:
 #            value: "true"
 #          - name: TYPHA_PROMETHEUSMETRICSPORT
 #            value: "9093"
+        resources:
+          requests:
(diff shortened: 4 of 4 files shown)


zanetworker:
@vpnachev as discussed, I have no good estimate for limits for some components, so I added requests only.

rfranzke:
Can we still add some limits based on some observations in one of our large clusters?

zanetworker:
Sure, thats where my requests estimates are coming from, but I can't see data on Grafana for these components so I have no historical reference just a point-in-time reference. Furthermore, on GKE I saw no limits set for `typha`, and no requests at all for the cvpa.

rfranzke:
I don’t know if it is a good idea to not specify anything. At least requests we should have for everything, and ideally limits as well.

zanetworker:
Requests are already covered, I was talking about limits more. In anycase, I will add a limit for the kube-controllers that is a little bit more generous than the one I see in our landscapes.

controllers/networking-calico/charts/internal/calico/templates/typha-cpva/deployment-calico-typha-vertical-autoscaler.yaml
```diff
@@ -34,6 +34,11 @@ spec:
             - --poll-period-seconds=30
             - --v=2
             - --config-file=/etc/config/typha-autoscaler
+          resources:
+            requests:
+              cpu: 10m
+            limits:
```
rfranzke: Shouldn’t the limits be somewhat higher than the requests here? Also, why is there no memory definition?

zanetworker: pelase have a look at the GKE clusters, I used them as reference for this specific component.

zanetworker: Also: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/calico-policy-controller/typha-horizontal-autoscaler-deployment.yaml

rfranzke: I don't understand the ping-pong. Do we now blindly copy stuff from others without understanding the rationales behind it? I'm fine with the current state of the PR if you can explain why there is no need to specify any limits/requests and why the limits should not be at least slightly higher than the requests to accommodate larger clusters, especially for a compressible resource like CPU.

vpnachev: the CPU is simply throttled when the limit is hit, so it is not critical. 
Also, when requests==limits the QoS is [`Guaranteed`](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed). So, @zanetworker I think you can even set the memory requests to be the same limits.

zanetworker: You didn't explain your rational behind having the limits more than the requests, but let me explain why it is set like this in the `Kuberentes` repo as far as I understood it. If a components is not meant to consume more than 1m or 2m CPU why would you give it more limits which would only occur if there is a bug in the code, the idea here is fail fast in case such bug exists for us to notice (at least thats what I understood). AND thats why I used this reference when I added this component. 

That said, you don't have to use the royal "we". And no "I" did not blindly copy anything. If you have a tangible reference for value please provide them, along with where you had that reference from other than a PIT guess.

zanetworker: @vpnachev thanks for the constructive feedback.

rfranzke: > You didn't explain your rational behind having the limits more than the requests

I did, see https://github.com/gardener/gardener-extensions/pull/568#discussion_r369435290
>> to accommodate larger clusters, especially for a compressible resource like CPU.

If it will never happen that `10m` CPU are reached for this component, even for large clusters, then everything is good. I just don't want to run into a situation where we are hit by a problem just because our limit was not even slightly higher than the requested value. This has happened sometimes in the past and always required manual intervention and adaption on our end which should be avoided.

> Also: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/calico-policy-controller/typha-horizontal-autoscaler-deployment.yaml

These manifests are mostly used for the e2e testing of Kubernetes which does not mean that they fit for all use-cases. GKE is a good reference, but we don't know what they do in the background. So far, we have always specified `cpu`/`memory` values for both `requests`/`limits` to avoid exactly those situations that triggered this PR.

Anyways, now it's specified

PR #572, opened 2020-01-22
Support for slice of similar type of operatingsystemconfigs

### LINKED ISSUE #318: Make OS controllers capable to support different linux distro families
Currently there is OS controllers for the following distros:
* CoreOS Linux
* Ubuntu
* Suse

These OSes are part of Distro families that shares a lot, e.g. Ubuntu and Debian, openSuse and Suse,  CoreOS and FlatCar.

For example, the current coreos controller successfully bootstraps FlatCar Linux nodes, but it cannot be easily configured, because the coreos controller is acting only on `OperatingSystemConfigs` of type `coreos` and this is hard-coded in the extension controller https://github.com/gardener/gardener-extensions/blob/644b5245ea0ef543385db2313a900dc40b71a06a/controllers/os-coreos/pkg/coreos/add.go#L25

A possible solution would be to make the controller configurable.  At the moment `ControllerRegistration` specifies on what kind of resource the controller should act on
https://github.com/gardener/gardener-extensions/blob/644b5245ea0ef543385db2313a900dc40b71a06a/controllers/os-coreos/example/controller-registration.yaml#L8-L9
This configuration should be propagated to the controller.
So, with the following controller registration
```yaml
---
apiVersion: core.gardener.cloud/v1alpha1
kind: ControllerRegistration
metadata:
  name: os-coreos
spec:
  resources:
  - kind: OperatingSystemConfig
    type: coreos
  - kind: OperatingSystemConfig
    type: flatcar
...
```
the os-coreos controller should be able to reconcile `coreos` and `flatcar` at the same time.

majst01:
For OS Extensions which have to handle semantically identical OS derivates like ubuntu/debian, or coreos/flatcar, two almost identical controllers have to be written.

Modify the AddArgs to support slice of types which are or´ed afterwards.

All affected os-controllers are modified to match.

Closes  #318

```noteworthy developer
All extension controllers can now provide multiple extension types instead of only one. This allows to bundle multiple responsibilities into the same extension controller.
```

### DIFF
--- a/controllers/os-coreos/pkg/coreos/add.go
+++ b/controllers/os-coreos/pkg/coreos/add.go
@@ -21,12 +21,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
-// Type is the type of OperatingSystemConfigs the coreos actuator / predicate are built for.
-const Type = "coreos"
-
 var (
 	// DefaultAddOptions are the default controller.Options for AddToManager.
 	DefaultAddOptions = AddOptions{}
+	// Types is the type of OperatingSystemConfigs the coreos actuator / predicate are built for.
+	Types = []string{"coreos"}
 )
 
 // AddOptions are the options for adding the controller to the manager.
@@ -44,7 +43,7 @@ func AddToManagerWithOptions(mgr manager.Manager, opts AddOptions) error {
 		Actuator:          NewActuator(),
 		ControllerOptions: opts.Controller,
 		Predicates:        operatingsystemconfig.DefaultPredicates(opts.IgnoreOperationAnnotation),
-		Type:              Type,
+		Types:             Types,
 	})
 }
 
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -199,10 +199,23 @@ func IsDeleting() predicate.Predicate {
 }
 
 // AddTypePredicate returns a new slice which contains a type predicate and the given `predicates`.
-func AddTypePredicate(extensionType string, predicates []predicate.Predicate) []predicate.Predicate {
-	preds := make([]predicate.Predicate, 0, len(predicates)+1)
-	preds = append(preds, HasType(extensionType))
-	return append(preds, predicates...)
+// if more than one extensionTypes is given all given types are or combined
+func AddTypePredicate(predicates []predicate.Predicate, extensionTypes ...string) []predicate.Predicate {
+	resultPredicates := make([]predicate.Predicate, 0, len(predicates)+1)
+	resultPredicates = append(resultPredicates, predicates...)
+
+	if len(extensionTypes) == 1 {
+		resultPredicates = append(resultPredicates, HasType(extensionTypes[0]))
+		return resultPredicates
+	}
+
+	orPreds := make([]predicate.Predicate, 0, len(extensionTypes))
+	for _, extensionType := range extensionTypes {
+		orPreds = append(orPreds, HasType(extensionType))
+	}
+	orPred := Or(orPreds...)
+
+	return append(resultPredicates, orPred)
 }
 
 // HasPurpose filters the incoming Controlplanes  for the given spec.purpose
--- a/controllers/os-coreos-alicloud/pkg/coreos-alicloud/add.go
+++ b/controllers/os-coreos-alicloud/pkg/coreos-alicloud/add.go
@@ -21,12 +21,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
-// Type is the type of operating system configs the CoreOS Alicloud controller monitors.
-const Type = "coreos-alicloud"
-
 var (
 	// DefaultAddOptions are the default controller.Options for AddToManager.
 	DefaultAddOptions = AddOptions{}
+	// Types is the type of operating system configs the CoreOS Alicloud controller monitors.
+	Types = []string{"coreos-alicloud"}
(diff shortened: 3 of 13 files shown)


pkg/predicate/predicate.go
```diff
@@ -205,6 +205,19 @@ func AddTypePredicate(extensionType string, predicates []predicate.Predicate) []
 	return append(preds, predicates...)
 }
 
+// AddTypesPredicate returns a new slice which contains a types predicate and the given `predicates`.
+// all given types are or combined
+func AddTypesPredicate(extensionTypes []string, predicates []predicate.Predicate) []predicate.Predicate {
```
rfranzke: Does it make sense to re-use the `AddTypePredicate` function instead and change its signature from

`func AddTypePredicate(extensionType string, predicates []predicate.Predicate) []predicate.Predicate`

to

`func AddTypePredicate(predicates []predicate.Predicate, extensionType ...string) []predicate.Predicate`

?

majst01: sounds legit, gimme a second

pkg/predicate/predicate.go
```diff
@@ -199,10 +199,23 @@ func IsDeleting() predicate.Predicate {
 }
 
 // AddTypePredicate returns a new slice which contains a type predicate and the given `predicates`.
-func AddTypePredicate(extensionType string, predicates []predicate.Predicate) []predicate.Predicate {
-	preds := make([]predicate.Predicate, 0, len(predicates)+1)
-	preds = append(preds, HasType(extensionType))
-	return append(preds, predicates...)
+// if more than one extensionTypes is given all given types are or combined
+func AddTypePredicate(predicates []predicate.Predicate, extensionTypes ...string) []predicate.Predicate {
```
rfranzke: Looks great! But I think you have to adapt all the controllers in `pkg/controller` now, or?

majst01: Damn, right, one more second ...

rfranzke (dismissed):

majst01:
If someone pastes the errors from the broken CI runs, i´m happy to fix.

rfranzke:
@ccwienk @AndreasBurger @RaphaelVogel why do external contributors still have no access to their CI builds?

rfranzke:
@majst01 the issue is
```
Executing golangci-lint
pkg/predicate/predicate.go:209:10: SA4021: x = append(y) is equivalent to x = y (staticcheck)
		return append(resultPredicates)
		       ^
make: *** [Makefile:51: check] Error 1
```

ialidzhikov:
Could you please squash to a single commit according to https://github.com/gardener/documentation/blob/master/CONTRIBUTING.md?
You may also want to state that this PR is aimed to fix https://github.com/gardener/gardener-extensions/issues/318 and add appropriate release note about it.

ialidzhikov (approved):
Nice. I was able to test the coreos/flatcar types with a minor change on top of this PR.

```
$ k get nodes -o wide
NAME                                         STATUS   ROLES    AGE   VERSION   INTERNAL-IP    EXTERNAL-IP   OS-IMAGE                                                 KERNEL-VERSION    CONTAINER-RUNTIME
ip-10-222-1-242.eu-west-1.compute.internal   Ready    <none>   19m   v1.15.9   10.222.1.242   <none>        Container Linux by CoreOS 2303.3.0 (Rhyolite)            4.19.86-coreos    docker://18.6.3
ip-10-222-18-13.eu-west-1.compute.internal   Ready    <none>   69s   v1.15.9   10.222.18.13   <none>        Flatcar Container Linux by Kinvolk 2303.3.1 (Rhyolite)   4.19.86-flatcar   docker://18.6.3
```

We could add the types and add few words as docs with other PR(s).

/lgtm

majst01:
@ialidzhikov should i add these additional types `coreos` and `flatcar` to the coreos controller as well ? I don´t know your images and how they compare, so this PR is only the groundwork for more simplifications, right ?

controllers/os-coreos/pkg/coreos/add.go
```diff
@@ -21,12 +21,11 @@ import (
...
 )
 
-// Type is the type of OperatingSystemConfigs the coreos actuator / predicate are built for.
-const Type = "coreos"
-
 var (
 	// DefaultAddOptions are the default controller.Options for AddToManager.
 	DefaultAddOptions = AddOptions{}
+	// Types is the type of OperatingSystemConfigs the coreos actuator / predicate are built for.
+	Types = []string{"coreos"}
```
ialidzhikov:
```suggestion
	Types = []string{"coreos", "flatcar"}
```

ialidzhikov (commented):
> @ialidzhikov should i add these additional types coreos and flatcar to the coreos controller as well ?

We can add it to have 1 scenario working end-to-end.
If you say it is ok as it is, we can do it with other PRs.

controllers/os-coreos/pkg/coreos/add.go
rfranzke: (FYI: What @ialidzhikov suggests should be done via https://github.com/gardener/gardener-extensions/blob/master/controllers/os-coreos/charts/os-coreos/doc.go#L15 and `go generate`)

majst01: I would create other PR´s which can then take care of further simplifications on the controller side. 
I´m fine merging that now.

rfranzke: Fair enough. 👍

PR #574, opened 2020-01-23
provider-vsphere: fixing etcd storage provider and parameters

MartinWeindel:
wrong storage class for etcd-main

**Release note**:

```improvement operator

```


rfranzke:
Can you also change https://github.com/gardener/gardener-extensions/blob/master/controllers/provider-vsphere/example/00-componentconfig.yaml#L13-L16 ?

PR #580, opened 2020-01-24
Allow os-coreos controller to handle Flatcar distro

ialidzhikov:
Allow os-coreos controller to handle Flatcar distro.

Fixes #

**Release note**:

```improvement operator
`os-coreos` extension controller is now able to handle `OperatingSystemConfig` with `.spec.type=flatcar`.
```


majst01:
Nice, this tells me also what i was missing in our ubuntu/debian controller:

https://github.com/metal-pod/os-metal-extension/pull/2

timuthy:
Now we should think about renaming this thing, imo.

rfranzke:
What do you want to rename? The controller itself?

majst01:
We have the same discussion for our controller, we support `ubuntu` and `debian`, but in our flavour which creates `ignition` rather than `cloud-init` payloads. You guys also have a `ubuntu` controller which would overlap then.

I think this topic misses a additional property in the OperatingSystemConfig which allows a distinction of the provider. Should be a additional field along with version i think.

what about:
```yaml
kind: OperatingSystemConfig
ostype: ubuntu         # can be coreos,flatcar,ubuntu, debian etc.
userdatatype: ignition # ignition or cloud-init
provider: gcp          # one of the supported cloud providers.
```
And the os-extension watches on these for properties.

Just to start the conversation.

rfranzke:
Where would the information which value Gardener shall use for the `userdatatype` come from when it creates the `OperatingSystemConfig`? Exposing it in the `Shoot` resource feels strange, end-users don't care about these implementation details?

majst01:
I think it must come from the images in the CloudProfile:

```yaml
apiVersion: core.gardener.cloud/v1beta1                                                                                                                                                                                                                                                           
kind: CloudProfile                                                                                                                                                                                                                                                                                
  machineImages:
  - name: ubuntu
    userdatatype: ignition
    versions:
    - version: "19.04"
    - version: "19.10"
  - name: debian
    userdatatype: ignition
    versions:
    - version: "10"
```
provider is obvious, it is already part of the cloudprofile. This way many different ubuntu flavours may exist without fighting each other.

Name of the controller can be something like: `os-ignition-gcp`

rfranzke:
Hm, but what if I have two `debian`s, one with `userdatatype=ignition` and another one with `userdatatype=cloudinit`?

majst01:
This would be a drawback of this, is this a use case we can think about ? Same provider with two forms of ubuntu images ?

We already discussed this topic here as well and we where not able to create a data model which will cover all of the aspects one can think of. It might be that the current approach leads us to the wrong direction. WDYT ?

On the other side we already see the need for improvement here because we already clash, and naming images like `ubuntu-metal` is not the right thing to do IMHO. This is also misleading for a end user.
 
/cc @Gerrit91

rfranzke:
Hm, well, although I don't have a concrete use-case for the scenario of registering two extensions listening to the same type in a productive system, I think it should be made possible.
Though, I would not add a `userdatatype` field (tomorrow there is again a clash and we need yet another distinction). Instead, I would indeed add logical names to the controllers, e.g.,

```yaml
apiVersion: core.gardener.cloud/v1beta1                                                                                                                                                                                                                                                           
kind: CloudProfile                                                                                                                                                                                                                                                                                
  machineImages:
  - name: ubuntu1
    type: ubuntu
    versions:
    - version: "19.04"
    - version: "19.10"
  - name: ubuntu2
    type: ubuntu
    versions:
    - version: "10"
  - name: debian
    versions:
    - version: "10"
```

The `type` field could be optional and, if not specified, Gardener would assume `type=name` to preserve backwards compatibility. WDYT?

majst01:
Yes, in order to keep over engineered complexity away, this is a good approach solve current requirements.

PR #591, opened 2020-01-31
Remove `install` command from Makefile

### LINKED ISSUE #590: Running `make install` returns an error
**What happened**:
`make install` returns the following error when run locally

image

Running `make install` should not return an error

1. Clone the master branch
2. Run `make install`

plkokanov:
Removes the `install` command from the Makefile

Fixes #590 

None

**Release note**:

```noteworthy developer
Removed unnecessary `install` command from Makefile
```


plkokanov:
Agree let’s remove it

PR #600, opened 2020-02-13
implement restore and migrate worker

vlvasilev:
With this PR we expand the worker actuator interface with Restore and Migrate functions.

The restoration restore previous state stored under worker.status.state which consist of machineDeloyments, machineSets and Machenes. Thus wen we migrate a shoot control plane, we keep the existing VM representing the shoot cluster nodes.

The migration makes shallow deletion of MachineClass, MachineSecrets, MachineDeployment, MachineSets and Machines. This shallow deletion remove Machine-Control-Manager from the shoot control plane, then remove the finalizers and then delete the resources. Thus we don't allow the MCM to delete the shoot nodes.

With this PR we also add a sidecar controller which handles any changes related to the Machine resources and if needed update the worker.status.state subresource.  
Part of  [#1631](https://github.com/gardener/gardener/issues/1631)

**Release note**:

```improvement operator
- Add Separate sidecar controller to save worker state under Worker.Status.State
- Add Worker.Status.State restoration functionality 
- Add migration functionality
```

### DIFF
--- a/pkg/controller/utils_test.go
+++ b/pkg/controller/utils_test.go
@@ -28,6 +28,9 @@ import (
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -102,5 +105,97 @@ var _ = Describe("Utils", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(secret).To(BeNil())
 		})
+	})
+
+	Describe("#DeleteAllFinalizers", func() {
+		It("should delete all finalizers", func() {
+			creationTimestamp := v1.Now()
+			deletionTimestamp := v1.Now()
+			labels := make(map[string]string)
+			labels["test-label-key"] = "test-label-value"
+			annotation := make(map[string]string)
+			annotation["test-annotation-key"] = "test-annotation-value"
+			owner := []metav1.OwnerReference{
+				{
+					APIVersion:         "test-api",
+					Kind:               "test-owner-kind",
+					Name:               "test-owner",
+					UID:                types.UID("test-owner-UID"),
+					Controller:         pointer.BoolPtr(true),
+					BlockOwnerDeletion: pointer.BoolPtr(true),
+				},
+			}
+
+			testFinalizer1 := "test-finalizer1"
+			testFinalizer2 := "test-finalizer2"
+			testFinalizer3 := "test-finalizer4"
+
+			finalizers := []string{
+				testFinalizer1,
+				testFinalizer2,
+				testFinalizer3,
+			}
+
+			secretRef := corev1.SecretReference{
+				Name:      "test-secret",
+				Namespace: "test-namespace",
+			}
+
+			worker := &extensionsv1alpha1.Worker{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "Worker",
+					APIVersion: "TestApi",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:                       "test-worker",
+					Namespace:                  "test-namespace",
+					GenerateName:               "test-generate-name",
+					SelfLink:                   "test-self-link",
+					UID:                        types.UID("test-UID"),
+					ResourceVersion:            "test-resource-version",
+					Generation:                 int64(1),
+					CreationTimestamp:          creationTimestamp,
+					DeletionTimestamp:          &deletionTimestamp,
+					DeletionGracePeriodSeconds: pointer.Int64Ptr(10),
+					Labels:                     labels,
+					Annotations:                annotation,
+					OwnerReferences:            owner,
+					Finalizers:                 finalizers,
+					ClusterName:                "test-cluster-name",
+				},
+				Spec: extensionsv1alpha1.WorkerSpec{
+					DefaultSpec: extensionsv1alpha1.DefaultSpec{
+						Type: "",
+					},
+					Region:    "test-region",
+					SecretRef: secretRef,
+				},
+			}
+			ctx := context.TODO()
+			key, err := client.ObjectKeyFromObject(worker)
+			Expect(err).NotTo(HaveOccurred())
+
+			gomock.InOrder(
+				c.EXPECT().
+					Get(ctx, key, worker).
+					DoAndReturn(func(_ context.Context, _ client.ObjectKey, worker *extensionsv1alpha1.Worker) error {
+						if len(worker.Finalizers) < 1 {
+							return fmt.Errorf("Worker %s has no finalizers", worker.Name)
+						}
+						for _, finalizer := range worker.Finalizers {
+							if finalizer != testFinalizer1 && finalizer != testFinalizer2 && finalizer != testFinalizer3 {
+								return fmt.Errorf("Finalizer %s not found for worker %s", finalizer, worker.Name)
+							}
+						}
+						return nil
+					}),
+
+				c.EXPECT().Update(ctx, worker),
+			)
+
+			Expect(controller.DeleteAllFinalizers(ctx, c, worker)).To(Succeed())
+			Expect(len(worker.Finalizers)).To(Equal(0))
+		})
+
 	})
 })
--- a/pkg/controller/worker/actuator.go
+++ b/pkg/controller/worker/actuator.go
@@ -28,4 +28,15 @@ type Actuator interface {
 	Reconcile(context.Context, *extensionsv1alpha1.Worker, *extensionscontroller.Cluster) error
 	// Delete deletes the Worker.
 	Delete(context.Context, *extensionsv1alpha1.Worker, *extensionscontroller.Cluster) error
+	// Restore reads from the worker.status.state field and deploys the machines and machineSet
+	Restore(context.Context, *extensionsv1alpha1.Worker, *extensionscontroller.Cluster) error
+	// Migrate deletes the MCM, machineDeployments, machineClasses, machineClassSecrets,
+	// machineSets and the machines. The underlying VMs representing the Shoot nodes are not deleted
+	Migrate(context.Context, *extensionsv1alpha1.Worker, *extensionscontroller.Cluster) error
+}
+
+// StateActuator acts upon Worker's State resources.
+type StateActuator interface {
+	// Reconcile reconciles the Worker State.
+	Reconcile(context.Context, *extensionsv1alpha1.Worker) error
 }
--- a/pkg/controller/worker/controller.go
+++ b/pkg/controller/worker/controller.go
@@ -17,10 +17,12 @@ package worker
 import (
 	extensionshandler "github.com/gardener/gardener-extensions/pkg/handler"
 	extensionspredicate "github.com/gardener/gardener-extensions/pkg/predicate"
-
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	machinev1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
+
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/source"
@@ -31,6 +33,8 @@ const (
 	FinalizerName = "extensions.gardener.cloud/worker"
 	// ControllerName is the name of the controller.
 	ControllerName = "worker_controller"
+	// StateUpdatingControllerName is the name of the controller responsible for updating the worker's state.
+	StateUpdatingControllerName = "worker_state_controller"
 )
 
 // AddArgs are arguments for adding an worker controller to a manager.
@@ -75,7 +79,11 @@ func DefaultPredicates(ignoreOperationAnnotation bool) []predicate.Predicate {
 func Add(mgr manager.Manager, args AddArgs) error {
 	args.ControllerOptions.Reconciler = NewReconciler(mgr, args.Actuator)
 	predicates := extensionspredicate.AddTypePredicate(args.Predicates, args.Type)
-	return add(mgr, args.ControllerOptions, predicates)
+	if err := add(mgr, args.ControllerOptions, predicates); err != nil {
+		return err
+	}
+
+	return addStateUpdatingController(mgr, args.ControllerOptions)
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
@@ -92,4 +100,34 @@ func add(mgr manager.Manager, options controller.Options, predicates []predicate
 	return ctrl.Watch(&source.Kind{Type: &extensionsv1alpha1.Cluster{}}, &extensionshandler.EnqueueRequestsFromMapFunc{
 		ToRequests: extensionshandler.SimpleMapper(ClusterToWorkerMapper(predicates), extensionshandler.UpdateWithNew),
 	})
+}
+
+func addStateUpdatingController(mgr manager.Manager, options controller.Options) error {
+	stateActuator := NewStateActuator(log.Log.WithName("worker-state-actuator"))
+	stateReconciler := NewStateReconciler(mgr, stateActuator)
+	addStateUpdatingControllerOptions := controller.Options{
+		MaxConcurrentReconciles: options.MaxConcurrentReconciles,
+		Reconciler:              stateReconciler,
+	}
+	predicates := []predicate.Predicate{
+		extensionspredicate.Or(
+			MachineStatusHasChanged(),
+			predicate.GenerationChangedPredicate{},
+		),
+	}
+
+	ctrl, err := controller.New(StateUpdatingControllerName, mgr, addStateUpdatingControllerOptions)
+	if err != nil {
+		return err
+	}
+
+	if err := ctrl.Watch(&source.Kind{Type: &machinev1alpha1.MachineSet{}}, &extensionshandler.EnqueueRequestsFromMapFunc{
+		ToRequests: extensionshandler.SimpleMapper(MachineSetToWorkerMapper(nil), extensionshandler.UpdateWithNew),
+	}, predicates...); err != nil {
+		return err
+	}
+
+	return ctrl.Watch(&source.Kind{Type: &machinev1alpha1.Machine{}}, &extensionshandler.EnqueueRequestsFromMapFunc{
+		ToRequests: extensionshandler.SimpleMapper(MachineToWorkerMapper(nil), extensionshandler.UpdateWithNew),
+	}, predicates...)
 }
--- a/pkg/controller/worker/genericactuator/actuator.go
+++ b/pkg/controller/worker/genericactuator/actuator.go
@@ -31,6 +31,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -52,6 +53,7 @@ type genericActuator struct {
 
 	client               client.Client
 	clientset            kubernetes.Interface
+	decoder              runtime.Decoder
 	gardenerClientset    gardenerkubernetes.Interface
 	chartApplier         gardenerkubernetes.ChartApplier
 	chartRendererFactory extensionscontroller.ChartRendererFactory
@@ -79,6 +81,11 @@ func (a *genericActuator) InjectFunc(f inject.Func) error {
 
 func (a *genericActuator) InjectClient(client client.Client) error {
 	a.client = client
+	return nil
+}
+
+func (a *genericActuator) InjectScheme(scheme *runtime.Scheme) error {
+	a.decoder = serializer.NewCodecFactory(scheme).UniversalDecoder()
 	return nil
 }
 
@@ -213,7 +220,29 @@ func (a *genericActuator) cleanupMachineClassSecrets(ctx context.Context, namesp
 	return nil
 }
 
-// cleanupMachineSets deletes MachineSets having number of desired and actual replicas equaling 0
+// shallowDeleteMachineClassSecrets deletes all unused machine class secrets (i.e., those which are not part
+// of the provided list <usedSecrets>) without waiting for MCM to do this.
+func (a *genericActuator) shallowDeleteMachineClassSecrets(ctx context.Context, namespace string, wantedMachineDeployments worker.MachineDeployments) error {
+	secretList, err := a.listMachineClassSecrets(ctx, namespace)
+	if err != nil {
+		return err
+	}
+	// Delete the finalizers to all secrets which were used for machine classes that do not exist anymore.
+	for _, secret := range secretList.Items {
+		if !wantedMachineDeployments.HasSecret(secret.Name) {
+			if err := extensionscontroller.DeleteAllFinalizers(ctx, a.client, &secret); err != nil {
+				return errors.Wrapf(err, "Error removing finalizer from MachineClassSecret: %s/%s", secret.Namespace, secret.Name)
+			}
+			if err := a.client.Delete(ctx, &secret); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// cleanupMachineClassSecrets deletes MachineSets having number of desired and actual replicas equaling 0
 func (a *genericActuator) cleanupMachineSets(ctx context.Context, namespace string) error {
 	machineSetList := &machinev1alpha1.MachineSetList{}
 	if err := a.client.List(ctx, machineSetList, client.InNamespace(namespace)); err != nil {
@@ -232,4 +261,21 @@ func (a *genericActuator) cleanupMachineSets(ctx context.Context, namespace stri
 		}
 	}
 	return nil
+}
+
+func (a *genericActuator) shallowDeleteAllObjects(ctx context.Context, namespace string, objectList runtime.Object) error {
+	if err := a.client.List(ctx, objectList, client.InNamespace(namespace)); err != nil {
+		return err
+	}
+
+	return meta.EachListItem(objectList, func(obj runtime.Object) error {
+		object := obj.DeepCopyObject()
+		if err := extensionscontroller.DeleteAllFinalizers(ctx, a.client, object); err != nil {
+			return err
+		}
+		if err := a.client.Delete(ctx, object); client.IgnoreNotFound(err) != nil {
+			return err
+		}
+		return nil
+	})
 }
--- a/pkg/controller/worker/genericactuator/actuator_migrate.go
+++ b/pkg/controller/worker/genericactuator/actuator_migrate.go
@@ -1,0 +1,95 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package genericactuator
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/gardener/gardener-extensions/pkg/controller"
+	machinev1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
+	appsv1 "k8s.io/api/apps/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	"github.com/pkg/errors"
+)
+
+// Migrate removes all machine related resources (e.g. MachineDeployments, MachineClasses, MachineClassSecrets, MachineSets and Machines)
+// without waiting for machine-controller-manager to do that. Before removal it ensures that the MCM is deleted.
+func (a *genericActuator) Migrate(ctx context.Context, worker *extensionsv1alpha1.Worker, cluster *controller.Cluster) error {
+	workerDelegate, err := a.delegateFactory.WorkerDelegate(ctx, worker, cluster)
+	if err != nil {
+		return errors.Wrap(err, "could not instantiate actuator context")
+	}
+
+	// Make sure machine-controller-manager is deleted before deleting the machines.
+	if err := a.deleteMachineControllerManager(ctx, worker); err != nil {
+		return errors.Wrap(err, "failed deleting machine-controller-manager")
+	}
+
+	if err := a.waitUntilMachineControllerManagerIsDeleted(ctx, worker.Namespace); err != nil {
+		return errors.Wrap(err, "failed deleting machine-controller-manager manager")
+	}
+
+	if err := a.shallowDeleteAllObjects(ctx, worker.Namespace, &machinev1alpha1.MachineList{}); err != nil {
+		return errors.Wrap(err, "shallow Deletion of all machine failed")
+	}
+
+	if err := a.shallowDeleteAllObjects(ctx, worker.Namespace, &machinev1alpha1.MachineSetList{}); err != nil {
+		return errors.Wrap(err, "shallow Deletion of all machineSets failed")
+	}
+
+	if err := a.shallowDeleteAllObjects(ctx, worker.Namespace, &machinev1alpha1.MachineDeploymentList{}); err != nil {
+		return errors.Wrap(err, "shallow Deletion of all machineDeployments failed")
+	}
+
+	if err := a.shallowDeleteAllObjects(ctx, worker.Namespace, workerDelegate.MachineClassList()); err != nil {
+		return errors.Wrap(err, "cleaning up machine classes failed")
+	}
+
+	if err := a.shallowDeleteMachineClassSecrets(ctx, worker.Namespace, nil); err != nil {
+		return errors.Wrap(err, "cleaning up machine class secrets failed")
+	}
+
+	// Wait until all machine resources have been properly deleted.
+	timeoutCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+
+	if err := a.waitUntilMachineResourcesDeleted(timeoutCtx, worker, workerDelegate); err != nil {
+		return gardencorev1beta1helper.DetermineError(err, fmt.Sprintf("Failed while waiting for all machine resources to be deleted: '%s'", err.Error()))
+	}
+
+	return nil
+}
+
+func (a *genericActuator) waitUntilMachineControllerManagerIsDeleted(ctx context.Context, namespace string) error {
+	return wait.PollUntil(5*time.Second, func() (bool, error) {
+		machineControllerManagerDeployment := &appsv1.Deployment{}
+		if err := a.client.Get(ctx, client.ObjectKey{Namespace: namespace, Name: McmDeploymentName}, machineControllerManagerDeployment); err != nil {
+			if apierrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, err
+		}
+
+		return false, nil
+	}, ctx.Done())
+}
--- a/pkg/controller/worker/genericactuator/actuator_reconcile.go
+++ b/pkg/controller/worker/genericactuator/actuator_reconcile.go
@@ -232,7 +232,13 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster 
 		// If the machine deployment does not yet exist we set replicas to min so that the cluster
 		// autoscaler can scale them as required.
 		case existingMachineDeployment == nil:
-			replicas = deployment.Minimum
+			if deployment.State != nil {
+				// During restoration the actual replica count is in the State.Replicas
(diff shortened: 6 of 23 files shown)


pkg/controller/worker/actuator.go
```diff
@@ -28,4 +28,15 @@ type Actuator interface {
 	Reconcile(context.Context, *extensionsv1alpha1.Worker, *extensionscontroller.Cluster) error
 	// Delete deletes the Worker.
 	Delete(context.Context, *extensionsv1alpha1.Worker, *extensionscontroller.Cluster) error
+	// Restore read from worker.status.state field and deploy the machines and machineSet
```
plkokanov:
```suggestion
	// Restore reads from the worker.status.state field and deploys the machines and machineSet
```

pkg/controller/worker/actuator.go
```diff
@@ -28,4 +28,15 @@ type Actuator interface {
...
+	Restore(context.Context, *extensionsv1alpha1.Worker, *extensionscontroller.Cluster) error
+	// Migrate delete the MCM, machineDeployments, mahicneClasees, machineClassSecrets,
```
plkokanov:
```suggestion
	// Migrate deletes the MCM, machineDeployments, mahicneClasees, machineClassSecrets,
```

pkg/controller/worker/actuator.go
```diff
@@ -28,4 +28,15 @@ type Actuator interface {
...
+	// machineSets and the machines, then deletes the worker
```
plkokanov: Maybe we should add additional information that actual VMs should not be deleted.

pkg/controller/worker/genericactuator/actuator.go
```diff
@@ -114,6 +122,21 @@ func (a *genericActuator) cleanupMachineDeployments(ctx context.Context, existin
...
 
+func (a *genericActuator) shallowDeleteMachineDeployments(ctx context.Context, existingMachineDeployments *machinev1alpha1.MachineDeploymentList, wantedMachineDeployments worker.MachineDeployments) error {
+	if err := a.cleanupMachineDeployments(ctx, existingMachineDeployments, wantedMachineDeployments); err != nil {
+		return err
+	}
+
+	for _, existingMachineDeployment := range existingMachineDeployments.Items {
+		if !wantedMachineDeployments.HasDeployment(existingMachineDeployment.Name) {
+			if err := extensionscontroller.DeleteAllFinalizers(ctx, a.client, &existingMachineDeployment); err != nil {
+				return errors.Wrap(err, fmt.Sprintf("Error removing finalizer from MachineDeployment: %s/%s", existingMachineDeployment.Namespace, existingMachineDeployment.Name))
```
plkokanov: You can directly use errors..Wrapf here

pkg/controller/worker/genericactuator/actuator.go
```diff
@@ -157,6 +180,30 @@ func (a *genericActuator) cleanupMachineClasses(ctx context.Context, namespace s
...
+
+	return meta.EachListItem(machineClassList, func(machineClass runtime.Object) error {
+		accessor, err := meta.Accessor(machineClass)
+		if err != nil {
+			return err
+		}
+
+		if !wantedMachineDeployments.HasClass(accessor.GetName()) {
+			if err := extensionscontroller.DeleteAllFinalizers(ctx, a.client, machineClass); err != nil {
+				return errors.Wrap(err, fmt.Sprintf("Error removing finalizer from MachineClass: %s/%s", accessor.GetNamespace(), accessor.GetName()))
```
plkokanov: errors.Wrapf

pkg/controller/worker/genericactuator/actuator.go
```diff
@@ -191,7 +238,29 @@ func (a *genericActuator) cleanupMachineClassSecrets(ctx context.Context, namesp
...
+	}
+	secretList, err := a.listMachineClassSecrets(ctx, namespace)
+	if err != nil {
+		return err
+	}
+	// Delete the finalizaers to all secrets which were used for machine classes that do not exist anymore.
+	for _, secret := range secretList.Items {
+		if !wantedMachineDeployments.HasSecret(secret.Name) {
+			if err := extensionscontroller.DeleteAllFinalizers(ctx, a.client, &secret); err != nil {
+				return errors.Wrap(err, fmt.Sprintf("Error removing finalizer from MachineClassSecret: %s/%s", secret.Namespace, secret.Name))
```
plkokanov: errors.Wrapf

pkg/controller/worker/genericactuator/actuator_migrate.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/gardener/gardener-extensions/pkg/controller"
+	machinev1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	"github.com/pkg/errors"
+)
+
+// Migrate remove all machine related resources (e.g. MachineDeployments, MachineClasses, MachineClassSecrets, MachineSets and Machines)
```
plkokanov:
```suggestion
// Migrate removes all machine related resources (e.g. MachineDeployments, MachineClasses, MachineClassSecrets, MachineSets and Machines)
```

pkg/controller/worker/genericactuator/actuator_migrate.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// Make sure machine-controller-manager is deleted before deleting the machines.
+	if err := a.deleteMachineControllerManager(ctx, worker); err != nil {
+		return errors.Wrapf(err, "failed deleting machine-controller-manager")
+	}
+
+	// Delete all existing machines without deleting the VM corresponding to them.
+	a.logger.Info("Shallow Deleting all machines", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
+	if err := a.shallowDeleteAllExistingMachines(ctx, worker.Namespace); err != nil {
+		return errors.Wrapf(err, "Shallow Deletion of all machines failed")
```
plkokanov: Here you can use `errors.Wrap` :)

pkg/controller/worker/genericactuator/actuator_migrate.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	// Get the list of all existing machine deployments.
+	existingMachineDeployments := &machinev1alpha1.MachineDeploymentList{}
+	if err := a.client.List(ctx, existingMachineDeployments, client.InNamespace(worker.Namespace)); err != nil {
+		return err
+	}
+
+	// Shallow delete all machine deployments.
+	if err := a.shallowDeleteMachineDeployments(ctx, existingMachineDeployments, nil); err != nil {
```
plkokanov: What is the point of the third argument in the function? It's only called with nil. Same question for `shallowDeleteMachineClasses` and `shallowDeleteMachineSecrets`

pkg/controller/worker/genericactuator/actuator_resore.go
```diff
@@ -0,0 +1,172 @@
...
+	workercontroller "github.com/gardener/gardener-extensions/pkg/controller/worker"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	machinev1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
```
plkokanov: Missing comment

pkg/controller/worker/genericactuator/actuator_resore.go
```diff
@@ -0,0 +1,172 @@
...
+	// Parse the worker state to a separete machineDeployment states and attach them to
+	// the corresponding machineDeployments which are to be deployed later
+	a.logger.Info("Extracting the worker status.state", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
+	if err := a.addStateToMachineDeployment(ctx, worker, wantedMachineDeployments); err != nil {
+		return err
+	}
+
+	wantedMachineDeployments = removeWantedDeploymentWithoutState(wantedMachineDeployments)
+
+	// Delete the machine-controller-manager. During restoration MCM must not exists
```
plkokanov: @hardikdr @prashanth26 @amshuman-kr can you say if this step is necessary if we perform the steps in the following order -> create machines -> create machineset -> create machinedeployment.
At this point we still do not have machineclasses so it's most likely necessary.

pkg/controller/worker/genericactuator/actuator_resore.go
```diff
@@ -0,0 +1,172 @@
...
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// alreadyExists must be returned from from apierrors.IsAlreadyExists(err) where err commes from client.Create.
+// This function is needed because wen you call client.Create and it successed if you immediately call TryUpdateStatus
```
plkokanov:
```suggestion
// This function is needed because when you call client.Create and it successed if you immediately call TryUpdateStatus
```

pkg/controller/worker/genericactuator/actuator_resore.go
```diff
@@ -0,0 +1,172 @@
...
+// it will return an error stating that the Object does not exists.
+// This bug comes form the fact that exponentialBackoff in TryUpdateStatus does not retry if client.Get returns error
+// so the function exits immediately with error
```
plkokanov: @rfranzke should we update TrypUpdateStatus to respect the exponentialBackoff when the initial `Get` returns an error (I'd imagine that that would be the desired behaviour actually)

pkg/controller/worker/genericactuator/actuator_resore.go
```diff
@@ -0,0 +1,172 @@
...
+func (a *genericActuator) updateStatus(ctx context.Context, obj runtime.Object, alreadyExists bool, transform func() error) error {
+	if alreadyExists {
+		return extensionscontroller.TryUpdateStatus(ctx, retry.DefaultBackoff, a.client, obj, transform)
+	}
+	return a.client.Status().Update(ctx, obj)
+}
+
+func removeWantedDeploymentWithoutState(wantedMachineDeployments workercontroller.MachineDeployments) workercontroller.MachineDeployments {
```
plkokanov: Why do we need this function. It's not clear to me on first viewing.

pkg/controller/worker/reconciler.go
```diff
@@ -84,91 +84,166 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+	if err := extensionscontroller.DeleteFinalizer(r.ctx, r.client, FinalizerName, worker); err != nil {
+		r.logger.Error(err, "Error removing finalizer from Worker", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
+		return err
+	}
+	return nil
+}
+
+func (r *reconciler) migrate(worker *extensionsv1alpha1.Worker, cluster *extensionscontroller.Cluster) (reconcile.Result, error) {
+	// remove operation annotation 'migrate'
+	withOpAnnotationRestore := worker.DeepCopyObject()
```
plkokanov:
```suggestion
	withOpAnnotationMigrate := worker.DeepCopyObject() 
```

pkg/controller/worker/reconciler.go
```diff
@@ -84,91 +84,166 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+	delete(worker.Annotations, v1beta1constants.GardenerOperation)
```
plkokanov: By the way can't we do here similar to what we aggreed for the `restore` annotation - to remove it after migration/restoration is finished

pkg/controller/worker/genericactuator/actuator.go
```diff
@@ -211,3 +280,49 @@ func (a *genericActuator) cleanupMachineSets(ctx context.Context, namespace stri
...
+		if err := extensionscontroller.DeleteAllFinalizers(ctx, a.client, machineSetDeepCopy); err != nil {
+			return errors.Wrap(err, fmt.Sprintf("Error removing finalizer from MachineSet: %s/%s", machineSet.Namespace, machineSet.Name))
+		}
+	}
+	return nil
+}
+
+// shallowDeleteMachines deletes the existing machines in a given namespace
+// without waiting machine-controll-manager to do this
+func (a *genericActuator) shallowDeleteAllExistingMachines(ctx context.Context, namespace string) error {
```
plkokanov: Can we make the `shallowDelete<object>` functions a bit more generic? Atleast for `shallowDeleteAllExistingMachines` and `shallowDeleteAllExistingMachineSets`
You can pass in the object type that should be deleted. I

 haven't thought the others could also be generalised.

plkokanov (changes requested):
Had a quick look through the PR. 

I'll do a more detailed pass tomorrow

pkg/controller/worker/controller.go
```diff
@@ -93,3 +101,34 @@ func add(mgr manager.Manager, options controller.Options, predicates []predicate
...
+func addStateUpdatingController(mgr manager.Manager, options controller.Options) error {
+	stateActuator := NewStateActuator(log.Log.WithName("worker-state-actuator"))
+	stateReconciler := NewStateReconciler(mgr, stateActuator)
+	addStateUpdatingControllerOptions := controller.Options{
+		MaxConcurrentReconciles: options.MaxConcurrentReconciles,
+		Reconciler:              stateReconciler,
+	}
+	predicates := []predicate.Predicate{
+		extensionspredicate.Or(
+			extensionspredicate.IsDeleting(),
```
danielfoehrKn: This triggers the watch when the worker is being deleted? I wonder why this predicate is needed? The state reconciler does not reconcile anyways [when there is a deletion timestamp](https://github.com/gardener/gardener-extensions/blob/4adcb350a6b15157bd27846528dfbe9d67b6c987/pkg/controller/worker/state_reconciler.go#L98).

pkg/controller/worker/reconciler.go
danielfoehrKn: do I understand that correctly that the Migration annotation is removed before the migration operation has completed? Why is it not done after the successful migration?

pkg/controller/worker/reconciler.go
```diff
@@ -84,91 +84,166 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-		hasFinalizer, err := extensionscontroller.HasFinalizer(worker, FinalizerName)
-		if err != nil {
-			r.logger.Error(err, "Could not instantiate finalizer deletion")
-			return reconcile.Result{}, err
-		}
-		if !hasFinalizer {
-			r.logger.Info("Deleting worker causes a no-op as there is no finalizer.", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
-			return reconcile.Result{}, nil
-		}
+	operationType := gardencorev1beta1helper.ComputeOperationType(worker.ObjectMeta, worker.Status.LastOperation)
```
danielfoehrKn: You could hand over the operationType to the delete, restore and reconcile functions? (instead of re-calculating them)

danielfoehrKn (changes requested):
Nice PR! Some smaller things only.

danielfoehrKn:
Another question: what is the reason for using runtime.Raw extension in the [MachineDeploymentState](https://github.com/gardener/gardener-extensions/blob/4adcb350a6b15157bd27846528dfbe9d67b6c987/pkg/controller/worker/machines.go#L65)?   Is it do be agnostic to the actual Machine Controller implementation( only needs to also use MachineDeployments, MachineSets, Machines)? I mean there is already mcm specific knowledge bundled (e.g in the [restore operation](https://github.com/gardener/gardener-extensions/blob/4adcb350a6b15157bd27846528dfbe9d67b6c987/pkg/controller/worker/genericactuator/actuator_resore.go#L61): deleting the mcm first depends on the actual implementation?). How can the migration even be agnostic to the MCM implementation (needs to be able to restore from state out-of-the-box etc.)?

pkg/controller/worker/reconciler.go
plkokanov: We had some talks regarding that (wrt to the `restore` annotation) and we will try to make the `migrate` annotation behave similarly - so that it is removed at the end.

pkg/controller/worker/genericactuator/actuator_migrate.go
vlvasilev: Monkey see, monkey do?

pkg/controller/worker/genericactuator/actuator_resore.go
vlvasilev: I don't want to interfere the normal reconciliation process and deploy MachineDeployment which is new for the Shoot in the time been migrated. In normal condition this function is meaningless but if the Shoot operator decides to add another worker  in the pool there will be a new wanted MachineDeployment which will be deployed before the MachineClass and I am not sure if this can break the the normal reconciliation. We assume that after the  Restore we need a following up reconciliation process

pkg/controller/worker/reconciler.go
vlvasilev: Yes we can but we do not need because there is a GardenerOperationType Migrate and we know when we are in it.
If you wish I can put the annotation removal at the end but this will add only another Reconciliation call which will do nothing

pkg/controller/worker/genericactuator/actuator_resore.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	// Do the actual restoration
+	a.logger.Info("Deploying the machines and MacineSets", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
+	if err := a.deployMachineSetsAndMachines(ctx, worker, wantedMachineDeployments); err != nil {
+		return errors.Wrapf(err, "failed restoration of the machineSet and the machines")
+	}
+
+	// Generate machine deployment configuration based on previously computed list of deployments and deploy them.
+	a.logger.Info("Deploying the sored in worker status.state machine deployments", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
```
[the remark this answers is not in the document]
vlvasilev: Yes you are right. I will change it to:
```a.logger.Info("Deploying the MachineDeployments in Worker.Status.State", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))```

File: pkg/controller/worker/controller.go (hunk 18 above)
Comment (vlvasilev) [reply pkg/controller/worker/controller.go]: good point. Thanks

File: pkg/controller/worker/reconciler.go (hunk 16 above)
Comment (vlvasilev) [reply pkg/controller/worker/reconciler.go]: Ok, I will put it at the end

Comment (vlvasilev) [reply pkg/controller/worker/reconciler.go]: Ok, It will be moved at the end

**Review by rfranzke — CHANGES_REQUESTED** [review]
Very nice PR - very well readable and very well structured - thank you very much! 💯 
I've got a bunch of questions and recommendations, please take a look :)

File: pkg/controller/worker/genericactuator/actuator.go (hunk 17 above)
Comment (vlvasilev) [reply pkg/controller/worker/genericactuator/actuator.go]: done

File: pkg/controller/worker/genericactuator/actuator_restore.go [hunk 21]
```diff
@@ -0,0 +1,162 @@
...
+	return nil
+}
+
+// alreadyExists must be returned from from apierrors.IsAlreadyExists(err) where err commes from client.Create.
+// This function is needed because when you call client.Create and it successed if you immediately call TryUpdateStatus
+// it will return an error stating that the Object does not exists.
+// This bug comes form the fact that exponentialBackoff in TryUpdateStatus does not retry if client.Get returns error
+// so the function exits immediately with error
+func (a *genericActuator) updateStatus(ctx context.Context, obj runtime.Object, alreadyExists bool, transform func() error) error {
+	if alreadyExists {
```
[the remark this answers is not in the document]
vlvasilev: No we can't. In `TryUpdateStatus`  first we call client.Get(...) which due to some cashes return ErrorNotFound.  The `exponentialBackoff` function does not wait if error occurs and returns it immediately. Thus  if call `client.Create(...)` and then immediately call `TryUpdateStatus` you get an error. I have wasted one day debugging this.

pkg/predicate/predicate.go
```diff
@@ -238,3 +241,39 @@ func HasPurpose(purpose extensionsv1alpha1.Purpose) predicate.Predicate {
...
+			return false
+		}
+		newMachine, ok := newObj.(*machinev1alpha1.Machine)
+		if !ok {
+			return false
+		}
+		oldStatus := oldMachine.Status
+		newStatus := newMachine.Status
+
+		//Check the Node
```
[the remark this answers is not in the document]
vlvasilev: I think that at the beginning there is no .status.node and after successful creation of a node the .status.node is set. The old one is nil and the new on is != nil

pkg/controller/worker/state_actuator.go
```diff
@@ -0,0 +1,208 @@
...
+	for index, machine := range existingMachines.Items {
+		if len(machine.OwnerReferences) > 0 {
+			for _, referant := range machine.OwnerReferences {
+				if referant.Kind == machineSetKind {
+					machines[referant.Name] = append(machines[referant.Name], existingMachines.Items[index])
+				}
+			}
+		} else if len(machine.Labels) > 0 {
+			if machineDeploymentName, ok := machine.Labels[machineDeploymentLabel]; ok {
+				machines[machineDeploymentName] = append(machines[machineDeploymentName], existingMachines.Items[index])
```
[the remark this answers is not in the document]
vlvasilev: @rfranzke 
This is a ugly avoidance of not getting in account an orphaned machine. If a Machine has `OwnerReference` this will be the `MachineSet`. but If there is no owner then we can use the lable which represents the `MachineDeployment`

pkg/controller/worker/machines.go
```diff
@@ -51,11 +51,20 @@ type MachineDeployment struct {
...
 }
 
 // MachineDeployments is a list of machine deployments.
 type MachineDeployments []MachineDeployment
 
+// MachineDeploymentState stores the last versions of the machine sets and machine which
+// the machine deployment corresponds
+type MachineDeploymentState struct {
+	Replicas   int32
+	MachineSet *machinev1alpha1.MachineSet `json:"machineSet,omitempty"`
```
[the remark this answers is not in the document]
vlvasilev: @rfranzke Is it possible to have more than one `MachineSet` per `MachineDeployment` ?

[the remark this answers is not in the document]
prashanth26: Yes, it is possible to have multiple machineSets per machineDeployment. Especially true in rollingUpdate scenarios.

pkg/controller/worker/state_reconciler.go
```diff
@@ -0,0 +1,126 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+)
+
+const (
+	// StartToSyncState is used as part of the Event 'reason' when a Worker state starts to synchronize
+	StartToSyncState = "SynchronizingState"
```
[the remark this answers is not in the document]
vlvasilev: I haven't paid any attention. But there is one when all of the whatched resources ( MachineSet and Machines) are deployed. But because the worker controller waits untill all of the machines are ready then initialy we have only one State update. So during creation and restore there is only one. But on consequential scale down or up the shoot there will be one update for each deleted resorces then one for each new resources and one for each machine.Status.Node  change.
I can mitigate the count of updates if the `worker.lastOperation.Status  == "Success".`

pkg/controller/worker/state_reconciler.go
```diff
@@ -0,0 +1,126 @@
...
+	// Deletion flow
+	if worker.DeletionTimestamp != nil {
+		//Nothing to do
+		return reconcile.Result{}, nil
+	}
+
+	// Reconcile flow
+	operationType := gardencorev1beta1helper.ComputeOperationType(worker.ObjectMeta, worker.Status.LastOperation)
+	if operationType != gardencorev1beta1.LastOperationTypeReconcile {
+		return reconcile.Result{Requeue: true}, nil
```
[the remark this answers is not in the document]
vlvasilev: In case create there can be deployed a lot of resources which are going to trigger  reconciliation multiple times.
During this period the `Machines` have no `Status.Node` until the end when worker.status.lastOperation is set to "Success".
So if worker haven't been created yet (even if only one `Machine is pending`) , then is no big deal to loose its state.
If I do not re-queue the request after successful creation there will be no changes to trigger status update.
And by the way I am compering `operationType` which is the current operation not the lastOperation
In case of successful creation on the next row the  `operationType` will be "Reconcile" and the status updating controller will update the status at once.

pkg/controller/worker/genericactuator/actuator.go
```diff
@@ -211,3 +279,49 @@ func (a *genericActuator) cleanupMachineSets(ctx context.Context, namespace stri
 	}
 	return nil
 }
+
+// shallowDeleteMachineSets deletes the existing machineSets in a given namespace
+// by simply deleting the finalizers.
+func (a *genericActuator) shallowDeleteAllExistingMachineSets(ctx context.Context, namespace string) error {
+	machineSetList := &machinev1alpha1.MachineSetList{}
+	if err := a.client.List(ctx, machineSetList, client.InNamespace(namespace)); err != nil {
+		if apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
vlvasilev: This function does not exists anymore

pkg/controller/worker/genericactuator/actuator_migrate.go
```diff
@@ -0,0 +1,85 @@
...
+// Migrate removes all machine related resources (e.g. MachineDeployments, MachineClasses, MachineClassSecrets, MachineSets and Machines)
+// without waiting for machine-cotroll-manager to do that. Before removal it ensures that the MCM is deleted.
+func (a *genericActuator) Migrate(ctx context.Context, worker *extensionsv1alpha1.Worker, cluster *controller.Cluster) error {
+	workerDelegate, err := a.delegateFactory.WorkerDelegate(ctx, worker, cluster)
+	if err != nil {
+		return errors.Wrap(err, "could not instantiate actuator context")
+	}
+
+	// Make sure machine-controller-manager is deleted before deleting the machines.
+	if err := a.deleteMachineControllerManager(ctx, worker); err != nil {
```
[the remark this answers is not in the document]
vlvasilev: I will test it but it could trigger new machine creation if I don't manage to delete quickly the `MachineDeployment` and the `MachineSet`

pkg/controller/worker/controller.go
```diff
@@ -93,3 +101,33 @@ func add(mgr manager.Manager, options controller.Options, predicates []predicate
...
+		return err
+	}
+
+	if err := ctrl.Watch(&source.Kind{Type: &machinev1alpha1.MachineSet{}}, &extensionshandler.EnqueueRequestsFromMapFunc{
+		ToRequests: extensionshandler.SimpleMapper(MachineSetToWorkerMapper(nil), extensionshandler.UpdateWithNew),
+	}, predicates...); err != nil {
+		return err
+	}
+
+	return ctrl.Watch(&source.Kind{Type: &machinev1alpha1.Machine{}}, &extensionshandler.EnqueueRequestsFromMapFunc{
```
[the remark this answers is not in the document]
vlvasilev: Why? we are never saving any `MachineDeployment`. Only their name but only to find which state record corresponds to which wantedMachineDeployment.
`MachineDeployment` tell what the worker controller wants but does not means that it really have it and `MachineDeployment` is always the same.

plkokanov (changes requested):

pkg/controller/worker/state_actuator.go
```diff
@@ -0,0 +1,208 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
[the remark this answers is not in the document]
vlvasilev: No I can't. the State_Actuator import worker package and Contoller.go from worker package import State_Actuator. This ends in circular dependencies.

pkg/controller/utils_test.go
```diff
@@ -103,4 +105,105 @@ var _ = Describe("Utils", func() {
...
+			testFinalizer2,
+			testFinalizer3,
+		}
+
+		secretRef := corev1.SecretReference{
+			Name:      "test-secret",
+			Namespace: "test-namespace",
+		}
+
+		worker := &extensionsv1alpha1.Worker{
```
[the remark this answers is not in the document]
vlvasilev: In future It will be good to check if the function have side effects removing/alter some of not related worker definition

pkg/controller/worker/state_reconciler.go
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
danielfoehrKn:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/controller/worker/mapper_test.go
```diff
@@ -0,0 +1,311 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
danielfoehrKn:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/controller/worker/genericactuator/actuator_restore.go
```diff
@@ -0,0 +1,161 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
danielfoehrKn:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/controller/worker/genericactuator/actuator_migrate.go
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
danielfoehrKn:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

danielfoehrKn (changes requested):
Only nits from my side. However I did not test the code and find that hard to do especially now that there are not yet all pieces of the control-plane migration puzzle implemented.

pkg/controller/worker/genericactuator/actuator_restore.go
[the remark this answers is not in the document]
rfranzke: Hm, ok, then I would still try to retry instead of explicitly handling the `alreadyExists` error that may occur due to not-up-to-date caches.

Why not simply retrying the `a.client.Status().Update()` call for a few seconds using the [`retry.Until()`](https://github.com/gardener/gardener-extensions/blob/c9d757465d14ed7bbfcb7e823bce0e3786c63810/pkg/terraformer/waiter.go#L31-L46) function?

rfranzke (changes requested):
Great PR, a few minor suggestions, afterwards it's good to go in from my point of view! :)
PTAL

pkg/controller/worker/controller.go
```diff
@@ -17,10 +17,11 @@ package worker
 import (
 	extensionshandler "github.com/gardener/gardener-extensions/pkg/handler"
 	extensionspredicate "github.com/gardener/gardener-extensions/pkg/predicate"
-
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	machinev1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
```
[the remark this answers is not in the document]
vlvasilev: What Order should I use?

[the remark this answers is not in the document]
rfranzke: Usually: First block with core packages (`fmt`, etc.), second block with repository-specific packages (everything from `github.com/gardener/gardener-extensions/...`, and third block the rest).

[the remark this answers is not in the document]
danielfoehrKn: Thanks for pointing that out.

pkg/controller/worker/genericactuator/actuator_restore.go
```diff
@@ -0,0 +1,161 @@
...
+	a.logger.Info("Extracting the worker status.state", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
+	if err := a.addStateToMachineDeployment(ctx, worker, wantedMachineDeployments); err != nil {
+		return err
+	}
+
+	wantedMachineDeployments = removeWantedDeploymentWithoutState(wantedMachineDeployments)
+
+	// Delete the machine-controller-manager. During restoration MCM must not exists
+	a.logger.Info("Deleting machine-controller-manager for a Worker restore operation during control-plane migration", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
+	if err := a.deleteMachineControllerManager(ctx, worker); err != nil {
```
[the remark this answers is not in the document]
vlvasilev: Sure. Thanks

pkg/controller/worker/genericactuator/actuator_restore.go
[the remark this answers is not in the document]
vlvasilev: I will use retry.Until() with TryUpdateStatus

rfranzke (commented):
Looks great, will play with it beginning of next week and then we can hopefully merge it! :)

pkg/controller/worker/genericactuator/actuator_migrate.go
```diff
@@ -0,0 +1,95 @@
...
+	if err := a.shallowDeleteMachineClassSecrets(ctx, worker.Namespace, nil); err != nil {
+		return errors.Wrap(err, "cleaning up machine class secrets failed")
+	}
+
+	// Wait until all machine resources have been properly deleted.
+	timeoutCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+
+	if err := a.waitUntilMachineResourcesDeleted(timeoutCtx, worker, workerDelegate); err != nil {
+		return gardencorev1beta1helper.DetermineError(fmt.Sprintf("Failed while waiting for all machine resources to be deleted: '%s'", err.Error()))
```
rfranzke:
```suggestion
		return gardencorev1beta1helper.DetermineError(err, fmt.Sprintf("Failed while waiting for all machine resources to be deleted: '%s'", err.Error()))
```

pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -224,7 +224,13 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster
 		// If the machine deployment does not yet exist we set replicas to min so that the cluster
 		// autoscaler can scale them as required.
 		case existingMachineDeployment == nil:
-			replicas = deployment.Minimum
+			if deployment.State != nil {
+				// During restoration the actual replica count is in the State.Replicas
+				// If wanted deployment has no corresponding existing deployment, but has State, then we are in restoration process
+				replicas = int(deployment.State.Replicas)
```
rfranzke:
```suggestion
				replicas = deployment.State.Replicas
```

rfranzke (changes requested):
I played with it, and it worked very well, thanks.
Still, I have a few questions/suggestions/observations:

1. Please rebase to latest `master` and address the mentioned changes due to latest commits on `master`.

2. The format in the `.status.state` is

```yaml
status:
  state:
    <machineDeploymentName>:
      machineSets:
      - <machineSet>
      machines:
      - <machines>
```

This makes it a little hard in case we want to extend something in the future.
Can we change it to

```yaml
status:
  state:
    machineDeployments:
      <machineDeploymentName>:
        machineSets:
        - <machineSet>
        machines:
        - <machines>
```

3. The health checks are still performed even if the object is migrated or restored. Can you enhance the generic https://github.com/gardener/gardener-extensions/tree/master/pkg/controller/healthcheck controller to stop performing any health checks as long as migrate or restore operations are in progress? @danielfoehrKn might help and give some guidance how to do it in case you have questions.

4. I have run a shoot cluster with 2 machines. They were perfectly running and consequently also added to the `Worker`'s `.status.state` field. Then I scaled up to 3 machines. The third machines was IMMEDIATELY present int he `.status.state` as well. I didn't expect this, especially after our last conversation where you mentioned that only `Running` machines are added. Can you explain? Is this desired? If no, can you fix it?

vlvasilev:
@rfranzke About 4th question, after successful creation of the worker each consecutive operation will be reconcile, not create. Thus the state updater will see the reconcile operation and will start to update the worker's state.

rfranzke:
Does it mean that the state updater will NOT update anything if the `lastOperation` of the `Worker` is type=Create,state=Succeeded`?

vlvasilev:
@rfranzke If `lastOperation` is `type=Create,state=Succeeded` then the next operation will be reconcile. So the state updater will update everything.

rfranzke:
This might take up to `24h` in case the Gardener only reconciles in the maintenance time window. We can't afford this. Either we add only `Running` machines to the state (independent of the `lastOperation`) or we add all machines independent of their status (and again independent of the `lastOperation`). What do you suggest?

vlvasilev:
@rfranzke The state updater return reconcile.Result{Requeue: true} if the operation is different from reconciliation and migrate and it works on separated Queue. So after successful Creation of the Worker the state updater will be triggered. If you wish I can make storing the all machines regardless of their status.

rfranzke:
> So after successful Creation of the Worker the state updater will be triggered.

Now I'm confused. Didn't you just say that the state updater only acts when `lastOperation=Reconcile`? 

Anyways, I suggest to not take the `lastOperation` into account. It should either add all machines or only the `Running` machines (both is fine for me, whatever you think makes more sense, but probably adding all machines would be safer anyways). It should do nothing if the Worker is migrated or restored or deleted. WDYT?

vlvasilev:
@rfranzke [Here](https://github.com/vlvasilev/gardener-extensions/blob/b80a14cd66018f02e4ca28f5de3d4daf6401070d/pkg/controller/worker/state_reconciler.go#L101) is the logic. I am using `gardencorev1beta1helper.ComputeOperationType` which returns `reconcile` if the lastOperation is type=Create and status=Succed.

rfranzke:
OK. This means that your statements before are wrong and the state updater is adding the machines as soon as the cluster was created successfully, independent of the status of the machines, correct?

vlvasilev:
@rfranzke I was thinking that when the Worker is created successful then the machines status are successful. But you are right. After creation of the Worker all of the machine are taken into account despite their state. Should I leave it this way ?

rfranzke:
Yes, that's good. Thank you for clarifying.

rfranzke (approved):
/lgtm

PR #602, opened 2020-02-19
Use in-cluster shoot API-server urls for health-checks

zanetworker:
Currently the internal DNS entry for the Shoot API-servers is used for health-checks, these health-cheks translate to egress traffic costs. 

In an effort to optimize costs, this PR replaces the secret used for the kubeconfig of the seed cluster with another that makes use of an in-cluster url similar to https://github.com/gardener/gardener/pull/1928

Part of  #https://github.com/gardener/gardener/issues/1953 

**Release note**:

```improvement operator
Use internal Gardener secret for health checks.
```


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.13
 require (
 	github.com/Masterminds/semver v1.4.2
 	github.com/coreos/go-systemd v0.0.0-20190719114852-fd7a80b32e1f
-	github.com/gardener/gardener v1.0.1
+	github.com/gardener/gardener v1.0.1-0.20200213093126-7a6123b6ae21
```
ialidzhikov: Can you update to the sha of the current g/g master?

zanetworker: wasn't sure if I should do that, this change is dependent on a specific commit, which is beyond that release use by the branch. I could do that, but is there an advantage in doing so?

ialidzhikov: No problem, it can stay like this.

pkg/util/shoot_clients.go
```diff
@@ -69,7 +69,7 @@ func NewShootClients(c client.Client, clientset kubernetes.Interface, gardenerCl
 // NewClientForShoot returns the rest config and the client for the given shoot namespace.
 func NewClientForShoot(ctx context.Context, c client.Client, namespace string, opts client.Options) (*rest.Config, client.Client, error) {
 	gardenerSecret := &corev1.Secret{}
-	if err := c.Get(ctx, kutil.Key(namespace, v1beta1constants.SecretNameGardener), gardenerSecret); err != nil {
+	if err := c.Get(ctx, kutil.Key(namespace, v1beta1constants.SecretNameGardenerInternal), gardenerSecret); err != nil {
```
rfranzke: Can we have a fallback to the `SecretNameGardener` in case the `SecretNameGardenerInternal` is not found?

zanetworker: good point.

PR #622, opened 2020-03-16
Add old object to mutators

mvladev:
Mutators might be interested in the oldObject before the mutation. This old object is now passed to every single Ensurer. It is `nil` if the operation is `CREATE` and must always be checked.

Fixes # gardener/gardener#2048

This is still draft as it needs all other ensurers to get this implementation. 

**Release note**:

```action developer
All methods of `pkg/webhook/controlplane/genericmutator.Ensurer` now have an additional argument which can optionally contain the old value of the object. If the admission request was `CREATE` this value is always `nil`.
```

### DIFF
--- a/pkg/webhook/controlplane/genericmutator/mutator.go
+++ b/pkg/webhook/controlplane/genericmutator/mutator.go
@@ -48,29 +48,38 @@ type EnsurerContext interface {
 // If they don't initially, they are mutated accordingly.
 type Ensurer interface {
 	// EnsureKubeAPIServerService ensures that the kube-apiserver service conforms to the provider requirements.
-	EnsureKubeAPIServerService(context.Context, EnsurerContext, *corev1.Service) error
+	// "old" might be "nil" and must always be checked.
+	EnsureKubeAPIServerService(ctx context.Context, etcx EnsurerContext, new, old *corev1.Service) error
 	// EnsureKubeAPIServerDeployment ensures that the kube-apiserver deployment conforms to the provider requirements.
-	EnsureKubeAPIServerDeployment(context.Context, EnsurerContext, *appsv1.Deployment) error
+	// "old" might be "nil" and must always be checked.
+	EnsureKubeAPIServerDeployment(ctx context.Context, etcx EnsurerContext, new, old *appsv1.Deployment) error
 	// EnsureKubeControllerManagerDeployment ensures that the kube-controller-manager deployment conforms to the provider requirements.
-	EnsureKubeControllerManagerDeployment(context.Context, EnsurerContext, *appsv1.Deployment) error
+	// "old" might be "nil" and must always be checked.
+	EnsureKubeControllerManagerDeployment(ctx context.Context, etcx EnsurerContext, new, old *appsv1.Deployment) error
 	// EnsureKubeSchedulerDeployment ensures that the kube-scheduler deployment conforms to the provider requirements.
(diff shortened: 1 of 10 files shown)


mvladev:
Okay,  this is ready for review.

pkg/webhook/handler.go
```diff
@@ -99,9 +99,19 @@ func handle(ctx context.Context, req admission.Request, r *http.Request, f mutat
 		return admission.Errored(http.StatusBadRequest, errors.Wrapf(err, "could not get accessor for %v", obj))
 	}
 
+	var oldObj runtime.Object
+
+	// Only UPDATE and DELETE operations have oldObjects.
+	if len(req.OldObject.Raw) > 1 {
```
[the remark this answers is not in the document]
mvladev: There is no need for such check. nil slices always have len 0

pkg/webhook/network/mutator.go
```diff
@@ -61,9 +61,21 @@ func (m *mutator) Mutate(ctx context.Context, obj runtime.Object) error {
...
 		return fmt.Errorf("could not mutate, object is not of type \"Network\"")
 	}
-	return m.mutateFunc(network)
+
+	var oldNetwork *extensionsv1alpha1.Network
+
+	if oldObj == nil {
+		old, ok := oldObj.(*extensionsv1alpha1.Network)
+		if !ok {
+			return fmt.Errorf("could not mutate, old object is not of type \"Network\"")
```
ialidzhikov: It is up the the called to return "cannot mutate".
```suggestion
			return fmt.Errorf("could not cast old object to extensionsv1alpha1.Network")
```

pkg/webhook/handler.go
```diff
@@ -99,9 +99,19 @@ func handle(ctx context.Context, req admission.Request, r *http.Request, f mutat
...
+		oldObj = t.DeepCopyObject()
+		if err := decoder.DecodeRaw(ar.OldObject, oldObj); err != nil {
+			return admission.Errored(http.StatusBadRequest, errors.Wrapf(err, "could not get decode old object %v", oldObj))
```
ialidzhikov:
```suggestion
			return admission.Errored(http.StatusBadRequest, errors.Wrapf(err, "could not decode old object %v", oldObj))
```

pkg/webhook/handler.go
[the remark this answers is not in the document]
rfranzke: With it unnecessary mistakes can be avoided, https://github.com/gardener/gardener-extensions/pull/622#discussion_r394402266

mvladev:
@ialidzhikov PR updated - have another look

pkg/webhook/controlplane/genericmutator/mutator.go
```diff
@@ -48,29 +48,38 @@ type EnsurerContext interface {
 // If they don't initially, they are mutated accordingly.
 type Ensurer interface {
 	// EnsureKubeAPIServerService ensures that the kube-apiserver service conforms to the provider requirements.
-	EnsureKubeAPIServerService(context.Context, EnsurerContext, *corev1.Service) error
+	// "old" might be "nil" and must always be checked.
+	EnsureKubeAPIServerService(ctx context.Context, etcx EnsurerContext, new, old *corev1.Service) error
```
rfranzke: @ialidzhikov's comment https://github.com/gardener/gardener-extensions/pull/622#discussion_r394410496 is not addressed here yet

mvladev:
Renamed `newXXX` => `new`, `oldXXXX` => `old`

PR #624, opened 2020-03-18
Adapt terraformer Pod spec to the latest image

### LINKED ISSUE #597: SIGTERM is not sent to terraformer process
**What happened**:
terraformer package starts the terraformer Pod with
```yaml
  containers:
  - command:
    - sh
    - -c
    - sh /terraform.sh apply 2>&1; [[ -f /success ]] && exit 0 || exit 1
```

When the terraformer Pod needs to be terminated, kubelet sends SIGTERM signal to pid 1. But the process that actually gets the SIGTERM in this case is `sh`, but not `/terraform.sh`.
The shell shipped by default with Alpine Linux does not pass signals to children and `/terraform.sh` does not receive a SIGTERM to start a graceful termination.

See a really nice explanation at https://pracucci.com/graceful-shutdown-of-kubernetes-pods.html.

1. Deploy a terraformer pod with a relatively small `.spec.activeDeadlineSeconds` and `.spec.terminationGracePeriodSeconds` to be able to properly reproduce the issue. For example use `60` seconds instead of `630` seconds which is the value that the extension controllers pass to the terraformer obj.

2. Ensure that the Pod has a spec containing

```yaml
spec:
  activeDeadlineSeconds: 60
  terminationGracePeriodSeconds: 60
  containers:
  - command:
    - sh
    - -c
    - sh /terraform.sh apply 2>&1; [[ -f /success ]] && exit 0 || exit 1
```

Basically the terraformer package also sets `activeDeadlineSeconds` to `terminationGracePeriodSeconds` which is the amount of time to wait before SIGKILL is sent.

3. Apply the Pod

4. Delete the Pod after 10-20s of its initial start to simulate kubelet eviction or any other reason that can lead to Pod termination.

5. Get the logs of the terraformer Pod and ensure that it is running after the delete request is sent to the Pod. 

6. Ensure that after `terminationGracePeriodSeconds` (in our steps it is 60 seconds) the Pod will be killed and the terraformer process won't update anything in the state ConfigMap. Meanwhile the terraformer process already created several resources.

7. Ensure that if you deploy a second terraformer Pod (basically it is what the infrastructure controller does), it will fail with conflicts as the resources were already created with the previous run but the state for them was not preserved. 

**Environment**:

- gardener-extension-provider-aws: v1.3.0

### LINKED ISSUE #620: Cluster creation hangs due to subnet conflicts
**What happened**:
The cluster creation in a dedicated VPC is hanging due to subnet conflicts where the subnets are actually newly created by Gardener and should belong to the cluster. 

we had two scenarios :
1- the subnets have the correct names which contains the cluster name 
2- the subnets does not have any name 

the newly created subnets should be used by the cluster without attempting to recreate the same subnets.

ialidzhikov:
Adapt terraformer Pod spec to the changes in gardener/terraformer#36.

Fixes #597 
Fixes #620

This PR requires:
- [x] gardener/terraformer#36

**Release note**:

```improvement operator
`g/gardener-extensions` does now require the `g/terraformer@v1.0.0`.
```

### DIFF
--- a/pkg/terraformer/terraformer.go
+++ b/pkg/terraformer/terraformer.go
@@ -251,6 +251,11 @@ func (t *terraformer) createOrUpdateRole(ctx context.Context) error {
 				Resources: []string{"configmaps"},
 				Verbs:     []string{"*"},
 			},
+			{
+				APIGroups: []string{""},
+				Resources: []string{"secrets"},
+				Verbs:     []string{"*"},
+			},
 		}
 		return nil
 	})
@@ -320,7 +325,9 @@ func (t *terraformer) env() []corev1.EnvVar {
 	envVars := []corev1.EnvVar{
 		{Name: "MAX_BACKOFF_SEC", Value: "60"},
 		{Name: "MAX_TIME_SEC", Value: "1800"},
+		{Name: "TF_CONFIGURATION_CONFIG_MAP_NAME", Value: t.configName},
 		{Name: "TF_STATE_CONFIG_MAP_NAME", Value: t.stateName},
+		{Name: "TF_VARIABLES_SECRET_NAME", Value: t.variablesName},
 	}
 	for k, v := range t.variablesEnvironment {
 		envVars = append(envVars, corev1.EnvVar{Name: k, Value: v})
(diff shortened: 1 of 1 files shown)


timebertt (dismissed):
/lgtm

tim-ebert:
I assume, you will wait for the upgrade PR for `terraformer` and then merge this PR, right?

rfranzke (approved):
/lgtm

PR #626, opened 2020-03-20
Fix bug preventing worker reconciliation in some cases

### LINKED ISSUE #623: Generic worker actuator is not able to reconcile if spec was broken after it gets fixed again
**What happened**:
If the user configures broken worker pools and then fixes this later on, the generic worker actuator reconciliation does not clean up the previously created machine deployments that are undesired anymore…

https://github.com/gardener/gardener-extensions/blob/master/pkg/controller/worker/genericactuator/actuator_reconcile.go#L138-L155

It first deploys the “desired machine deployments” according to worker pool spec, afterwards it waits until they are ready, and only then it deletes all other still existing machine deployments.
However, the second step will block forever. Due to the invalid configuration before it will never get ready.

The reconciliation should succeed.

Configure worker pools with a spec that breaks, e.g., use machine types that are not available in certain zones. Afterwards, fix the spec and see the reconciliation never succeeding again.

**Anything else we need to know**:
The reason for this ordering is that, if a user adds a new worker pool and removes an older one in a single step, then we first want to create the new pool before deleting the old one to not cause interruption to the workload.

**Environment**:
Azure, 1.15.10, but it happens for all infras/versions.

rfranzke:
The generic worker actuator does now only consider the wanted machine deployments when it waits for them to be ready. Also, when it cleans up unwanted machine deployments it now waits until they have been deleted from the system.

Fixes #623 

**Release note**:

```improvement user
An issue in the machine reconciliation has been fixed that caused the reconciliation being stuck in some cases where a previously broken worker configuration (e.g., due to the specification of machine types that are not available in certain availability zones) has been corrected.
```

### DIFF
--- a/pkg/controller/worker/genericactuator/actuator_reconcile.go
+++ b/pkg/controller/worker/genericactuator/actuator_reconcile.go
@@ -145,7 +145,7 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 	timeoutCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
 	defer cancel()
 
-	if err := a.waitUntilMachineDeploymentsAvailable(timeoutCtx, cluster, worker, wantedMachineDeployments); err != nil {
+	if err := a.waitUntilWantedMachineDeploymentsAvailable(timeoutCtx, cluster, worker, wantedMachineDeployments); err != nil {
 		return gardencorev1beta1helper.DetermineError(fmt.Sprintf("Failed while waiting for all machine deployments to be ready: '%s'", err.Error()))
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -313,30 +321,32 @@ func (a *genericActuator) waitUntilMachineDeploymentsAvailable(ctx context.Conte
...
-			// replicas as desired (specified in the .spec.replicas).
-			for _, machineDeployment := range wantedMachineDeployments {
-				if machineDeployment.Name == existingMachineDeployment.Name {
-					if workerhealthcheck.CheckMachineDeployment(&existingMachineDeployment) == nil {
-						numHealthyDeployments++
-					}
-					numDesired += existingMachineDeployment.Spec.Replicas
-					numUpdated += existingMachineDeployment.Status.UpdatedReplicas
-				}
+			// If the Shoot is not hibernated we want to wait until all wanted machine deployments have been as many
```
danielfoehrKn:
```suggestion
			// If the Shoot is not hibernated we want to wait until all wanted machine deployments have as many
```

PR #627, opened 2020-03-22
Enhance health check actuator with precheck function

rfranzke:
With this PR the generic health check actuator is enhanced. It can now take a pre-check function that is performed before the health check is executed. If it returns false then the health check will be skipped. This accommodates extensions that only want to perform their health checks conditionally based on the object's or the cluster's settings.

**Release note**:

```action developer
The generic health check actuator can now take a pre-check function that is performed before the health check is executed. If it returns `false` then the health check will be skipped. This accommodates extensions that only want to perform their health checks conditionally based on the object's or the cluster's settings.
```

### DIFF
--- a/pkg/controller/healthcheck/healtcheck_actuator.go
+++ b/pkg/controller/healthcheck/healtcheck_actuator.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"sync"
 
+	extensionscontroller "github.com/gardener/gardener-extensions/pkg/controller"
 	"github.com/gardener/gardener-extensions/pkg/util"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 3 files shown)


pkg/controller/healthcheck/healtcheck_actuator.go
```diff
@@ -93,28 +96,73 @@ type checkResultForConditionType struct {
...
+					error:               err,
+					healthConditionType: healthConditionType,
+				}
+				return
+			}
+
+			if preCheckFunc != nil && !preCheckFunc(obj, cluster) {
+				channel <- channelResult{
+					healthCheckResult: &SingleCheckResult{
+						IsHealthy: true,
```
timuthy: I get the motivation setting this to `true` but in the future we should think about being more transparent, i.e. introducing a `ConditionStatus` that is acceptable/expected like `skipped`. For now we can at least log that the health check is skipped, so an operator gets this type of transparency. WDYT?

rfranzke: I can add the log, yes. The new condition status is a larger topic, indeed. In the g/g core care controller we also set it to `True` when the shoot is hibernated and no checks are performed.

zanetworker:
@rfranzke can you specify in what conditions is this needed, I can think of non-gardener setups that use some extensions (for testing), what else?

timuthy (dismissed):

pkg/controller/healthcheck/healtcheck_actuator.go
```diff
@@ -93,28 +96,74 @@ type checkResultForConditionType struct {
...
+						Detail:    err.Error(),
+						Reason:    "ReadExtensionObjectFailed",
+					},
+					error:               err,
+					healthConditionType: healthConditionType,
+				}
+				return
+			}
+
+			cluster, err := extensionscontroller.GetCluster(ctx, a.seedClient, request.Namespace)
```
ialidzhikov: I would suggest to move the get for cluster and extension object only if the `preCheckFunc != nil`. Currently the get calls  will be executed even if the `preCheckFunc  == nil`.

rfranzke: done, good idea :)

ialidzhikov: that was fast

ialidzhikov (approved):
/lgtm

danielfoehrKn (approved):

PR #629, opened 2020-03-23
Adapt to gardener/terraformer@v1.0.0

rfranzke:
Adapt `terraformer` package to `v1.0` of [gardener/terraformer](https://github.com/gardener/terraformer).

* ❌Requires merge of https://github.com/gardener/terraformer/pull/37
* ❌Release of [gardener/terraformer](https://github.com/gardener/terraformer)

**Release note**:

```action developer
When using this version of the Gardener Extensions library you have to make sure that you use at least `v1.0` of the `gardener/terraformer` image.
```

### DIFF
--- a/pkg/terraformer/terraformer.go
+++ b/pkg/terraformer/terraformer.go
@@ -320,7 +320,9 @@ func (t *terraformer) env() []corev1.EnvVar {
 	envVars := []corev1.EnvVar{
 		{Name: "MAX_BACKOFF_SEC", Value: "60"},
 		{Name: "MAX_TIME_SEC", Value: "1800"},
+		{Name: "TF_CONFIGURATION_CONFIG_MAP_NAME", Value: t.configName},
 		{Name: "TF_STATE_CONFIG_MAP_NAME", Value: t.stateName},
+		{Name: "TF_VARIABLES_SECRET_NAME", Value: t.variablesName},
 	}
 	for k, v := range t.variablesEnvironment {
 		envVars = append(envVars, corev1.EnvVar{Name: k, Value: v})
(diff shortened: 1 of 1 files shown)


pkg/terraformer/terraformer.go
```diff
@@ -320,7 +320,9 @@ func (t *terraformer) env() []corev1.EnvVar {
 	envVars := []corev1.EnvVar{
 		{Name: "MAX_BACKOFF_SEC", Value: "60"},
 		{Name: "MAX_TIME_SEC", Value: "1800"},
+		{Name: "TF_CONFIGURATION_CONFIG_MAP_NAME", Value: t.configName},
 		{Name: "TF_STATE_CONFIG_MAP_NAME", Value: t.stateName},
+		{Name: "TF_VARIABLES_SECRET_NAME", Value: t.variablesName},
```
ialidzhikov: I guess terraformer rbac needs to be updated
```
Error from server (Forbidden): secrets "bar.infra.tf-vars" is forbidden: User "system:serviceaccount:shoot--foo--bar:terraformer" cannot get resource "secrets" in API group "" in the namespace "shoot--foo--bar"
```

Btw what is the idea of this secret as its content is always `# New line is needed! Do not remove this comment.`?

ialidzhikov (changes requested):

pkg/terraformer/terraformer.go
rfranzke: Yes, will do, nice catch, thanks.

The secret contains the Terraform variables that themselves might include credentials. It's up to the caller to fill them. In our Gardener Extensions we decided to not use them, that's why the file is always empty (just has this comment). Generally, we don't know how others are calling or deploying our Terraformer. Though, as TF variables are a first-class citizen in the TF project I guess it makes sense to support it.

